PR #63, opened 2023-12-01
Update lifecycle_script.py install pyxis and enroot by default

sean-smith:
*Issue #, if available:*

*Description of changes:*


verdimrc (approved):
Suggest to bump to pyxis version 0.16.1 (which I just tested and it works).

Other than that, LGTM.

Issue #68, opened 2023-12-07
HP: easy-ssh.sh hardcodes region to us-west-2


verdimrc:
Easy-ssh.sh does not work outside us-west-2, because this script hardcodes the region to us-west-2.

KeitaW:
Resolved by the PR

PR #69, opened 2023-12-07
Add region spec in easy-ssh.sh

KeitaW:
*Issue #, if available:*
#68 


verdimrc:
If the `--region xxx` flag is made optional, then easy-ssh.sh will behave just like aws cli in how it treats env vars and cli arg.

verdimrc (approved):

PR #75, opened 2023-12-13
Update easy-ssh.sh

sean-smith:
Update command to echo directly to the `~/.ssh/config` file, saving a step

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/easy-ssh.sh
+++ b/1.architectures/5.sagemaker-hyperpod/easy-ssh.sh
@@ -61,9 +61,11 @@ echo -e "aws ssm start-session "${aws_cli_args[@]}" --target sagemaker-cluster:$
 
 echo -e "Add the following to your ~/.ssh/config to easily connect:"
 echo "
+cat <<EOF >> ~/.ssh/config
 Host ${cluster_name}
   User ubuntu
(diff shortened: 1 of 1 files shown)


1.architectures/5.sagemaker-hyperpod/easy-ssh.sh
```diff
@@ -61,9 +61,11 @@ echo -e "aws ssm start-session "${aws_cli_args[@]}" --target sagemaker-cluster:$
 
 echo -e "Add the following to your ~/.ssh/config to easily connect:"
 echo "
+cat <<EOF >> ~/.ssh/config
```
KeitaW, line 64: The script may require multiple executions, such as when a user forgets to add a key to the server. Currently, each run of the script appends the host configuration, which is not optimal. Introducing an optional argument to toggle this feature could help prevent unnecessary repetitions.

sean-smith: This is just echo'd out and not actually written to the file.

KeitaW: Sorry you are right.

KeitaW (approved):
LGTM

PR #79, opened 2023-12-14
Revert "Update lifecycle_script.py install pyxis and enroot by default"

johnbensnyder:
Reverts aws-samples/awsome-distributed-training#63


Roshrini (approved):

KeitaW:
Why do we want to revert the change? Appreciate if you could share the context for the change.

KeitaW (approved):
Synced up with Ben offline.

johnbensnyder:
Currently the docker_install is failing sporadically with `E: Could not get lock /var/lib/dpkg/lock-frontend. It is held by process 6836 (apt-get)`. This is causing cluster create to fail on larger clusters.

PR #111, opened 2024-01-23
Delete default Provisioning Parameters

sean-smith:
Having the wrong `provisioning_parameters.json` file causes cluster creation to fail so we're removing the default one.


aarashy:
Thanks!

KeitaW (approved):
LGTM

PR #133, opened 2024-02-08
Megatron-LM Llama2 example

verdimrc:
Make the example work on as low as 2 nodes (16x GPUs each with 80GB memory)

*Issue #, if available:* N/A

*Description of changes:* Pretrain Llama2 using Megatron-LM. Example tested to work on as few as 2x p4de (16x A100-80GB).


3.test_cases/1.megatron-lm/2.distributed-training.sbatch
```diff
@@ -101,9 +95,9 @@ srun -l "${ARGS[@]}" python -m torch.distributed.run "${TORCHRUN_ARGS[@]}" /work
         --log-interval 1 \
         --eval-iters 40 \
         --eval-interval 1000 \
-        --data-path "$(pwd)/gpt2/my-gpt2_text_document" \
-        --vocab-file "$(pwd)/gpt2/gpt2-vocab.json" \
-        --merge-file "$(pwd)/gpt2/gpt2-merges.txt" \
+        --data-path "${DATA_PATH}/gpt2/my-gpt2_text_document" \
```
sean-smith, line 98: what about multi-tenant environments? Reason I set it to `$(pwd)` is now the directory can be created in `/fsx/ubuntu` along with the submit files.

verdimrc: Yes, we can switch to data-path to `$(pwd)` too.

verdimrc: Hi @sean-smith , actually line 98 is correct, because `${DATA_PATH}` is the mounted path seen by the container.

verdimrc: In fact, the changes should be made earlier on line 34.

verdimrc: updated.

perifaws (approved):
LGTM

Issue #149, opened 2024-02-22
Reorganize repo and tag to V1


perifaws:
The repository needs reorganization:
1. Some directories are not numbered and need to be updated for consistency.
2. Change single digit to double digit for the example directory `1` -> `01`.
3. Create best practice directory for guidance on drivers, network optimizations.

Tag to V1 to provide a reference.

PR #150, opened 2024-02-22
Update organization and tag to V1

perifaws:
Addresses #149

Update naming for Jax, update naming for single digit examples, new directory for best practices (EFA cheat sheet was in architectures).


mhuguesaws:
Please stop numbering things. This is useless and will create plenty of challenges to add and remove things.

mhuguesaws:
Here is my proposal

```
- docs/
- core_infra/
- orchestrators
   - aws-parallelcluster/
   - sagemaker-hyperpod/
      - slurm/
          - lifecycle-scripts/
   - amazon-eks/
   - aws-batch/
- observability/
- ml-frameworks/
   - [FRAMEWORK_NAME]
       - slurm/
       - kubernetes/
       - Dockerfile
       - README
- ml-micro-benchmarks/
- infra-validation/
```

awsankur:
I like this structure. A couple of comments:
1. Observability solution will depend on the orchestrator. So we should have an observability section as part of each orchestrator. Ideally, we should be in a position where observability is automatically enabled when we build a cluster or can be enabled in a few steps when we have built a cluster.
2. I am assuming [FRAMEWORK_NAME] is a test case in our current structure. I think we can bring more clarity here. So we can have a given set of [FRAMEWORK_NAMES] which include:
a. Nvidia [Nemo, Nemo-Multimodal, BioNemo etc]...we can add DALI and MONAI in the future
b. MosaicML [MPT etc]
c. PyTorch [DDP, FSDP, etc]
d. SM [DataParallel, Model Parallel, FSDP, etc]
d. TensorFlow
e. JAX

Within each FRAMEWORK_NAME we can have Dockerfiles, sbatch scripts and kubernetes yaml and other necessary files for each model name

Thoughts?

mhuguesaws:
> I like this structure. A couple of comments:
> 
> 1. Observability solution will depend on the orchestrator. So we should have an observability section as part of each orchestrator. Ideally, we should be in a position where observability is automatically enabled when we build a cluster or can be enabled in a few steps when we have built a cluster.
> 2. I am assuming [FRAMEWORK_NAME] is a test case in our current structure. I think we can bring more clarity here. So we can have a given set of [FRAMEWORK_NAMES] which include:
>    a. Nvidia [Nemo, Nemo-Multimodal, BioNemo etc]...we can add DALI and MONAI in the future
>    b. MosaicML [MPT etc]
>    c. PyTorch [DDP, FSDP, etc]
>    d. SM [DataParallel, Model Parallel, FSDP, etc]
>    d. TensorFlow
>    e. JAX
> 
> Within each FRAMEWORK_NAME we can have Dockerfiles, sbatch scripts and kubernetes yaml and other necessary files for each model name
> 
> Thoughts?

Love 2. organize by "vendor"

For 1. don't think we'll go outside grafana+prometheus at this point. We can organize the observability section by orchestrator for since the deployment and setup will be different.

perifaws:
@mhuguesaws how about CloudWatch or profilers like Nsight?

mhuguesaws:
> @mhuguesaws how about CloudWatch or profilers like Nsight?

Profiler in profiler ;)

awsankur:
We should add Nsight

KeitaW:
I was wondering if `observability` should be under `orchestrators` or have subdirectories per orchestrators.

PR #158, opened 2024-02-26
Update add_users.sh

sean-smith:
extend to support FSx home directories.


sean-smith:
Needs to be moved

PR #169, opened 2024-02-29
SMHP: mask slurm daemons to prevent accidental broken cluster

verdimrc:
*Issue #, if available:* N/A

*Description of changes:* main use-case is to avoid accidentally starting slurmd on controller which will cause the controller's slurm.conf be overwritten.


sean-smith (approved):

nghtm:
Tested on g5 cluster, works as expected:

```
ubuntu@ip-10-1-43-197:~$ sudo systemctl restart slurmd
Failed to restart slurmd.service: Unit slurmd.service not found.
```

PR #231, opened 2024-04-01
Enable Auto-resume automatically

sean-smith:
This code enables auto-resume for hyperpod clusters. We should copy this snippet to all our examples:

```bash
AUTO_RESUME=""
    echo "Detected Hyperpod cluster.. enabling --auto-resume=1" 
    AUTO_RESUME="--auto-resume=1"
fi

srun ${AUTO_RESUME} ...
```

### DIFF
--- a/3.test_cases/16.pytorch-cpu-ddp/1.conda-train.sbatch
+++ b/3.test_cases/16.pytorch-cpu-ddp/1.conda-train.sbatch
@@ -3,7 +3,6 @@
 #SBATCH --exclusive
 #SBATCH --wait-all-nodes=1
 #SBATCH --nodes 2
-#SBATCH --cpus-per-task=4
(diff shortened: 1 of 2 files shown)


KeitaW:
LGTM but would you be so kind add the same to container example?

KeitaW:
Let's merge this once we figure out how to test H/W failure on CPU instance.

sean-smith:
> or enable only on gpu instances?

No let's enable on all instances since the [health checks](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-hyperpod-resiliency.html) work on more than just the

sean-smith:
@KeitaW can you review?

perifaws:
@KeitaW are we good on that one?

KeitaW (approved):
Sorry that I missed to review this. Yes LGTM hope we can test the functionality on CPU instance with generic node failure simulation.

PR #259, opened 2024-04-12
smhp: document requirement for multiple SGs

verdimrc:
*Issue #, if available:* cluster creation failed with "EFA health check error" on trn1/p4d/p4de/p5, when multiple SGs are used, and some of the SGs do not have EFA egress rules.

*Description of changes:* Add a reminder to user who create clusters with multiple security groups to ensure that every SG has the EFA egress rule.

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/2.SageMakerVPC.yaml
+++ b/1.architectures/5.sagemaker-hyperpod/2.SageMakerVPC.yaml
@@ -170,6 +170,14 @@ Resources:
     Properties:
       Domain: vpc
 
+  # NOTE: when you create additional security groups, you must ensure that every
+  # security group has ingress/egress from/to its own security group id. Failure
+  # to do so may cause trn1/p4d/p4de/p5 SMHP cluster creation to fail:
+  #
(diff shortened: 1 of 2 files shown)


mhuguesaws (approved):
LGTM

1.architectures/5.sagemaker-hyperpod/2.SageMakerVPC.yaml
```diff
@@ -170,6 +170,14 @@ Resources:
     Properties:
       Domain: vpc
 
+  # NOTE: when you create additional security groups, you must ensure that every
```
sean-smith, line 173: What's the scenario for multiple Security Groups? In the cluster-config there's only one right?

verdimrc: One customer used this setup. Their account has a specific way to configure networking, and decided to group different networking rules to different SGs.

verdimrc: And pass a list of SGs to the cluster-config.

PR #310, opened 2024-05-04
Instantiate model on CPU on rank=0 only to prevent CPU OOM

pbelevich:
e.g. 70B * 4 bytes * 8 processes > 2T RAM available on P5

### DIFF
--- a/3.test_cases/10.FSDP/train.py
+++ b/3.test_cases/10.FSDP/train.py
@@ -155,7 +155,13 @@ def main(args):
         logger.info(
             "Creating Model"
         )
-    model = AutoModelForCausalLM.from_config(model_config)
+    # Instantiate model on CPU on rank=0 only to prevent CPU OOM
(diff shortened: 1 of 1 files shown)


sean-smith (approved):
LGTM

KeitaW (approved):
LGTM. Thanks @pbelevich  for creating the PR.
Just a question for my future reference. 
in https://github.com/lchu-ibm/llama-recipes/blob/feaa344af3102d2dab840272e946abe93dd98afe/llama_finetuning.py#L152 they use `recurse=False` while in this PR you use `recurse=True`. Is there any difference in behavior because of this?

pbelevich:
@KeitaW I went through FSDP code and I think that `recurse` should be `False`. Thanks for careful code-review! Btw, I suggested a fix for llama-recipes: https://github.com/meta-llama/llama-recipes/pull/487

KeitaW:
For reference: https://github.com/pytorch/pytorch/issues/104187

PR #328, opened 2024-05-14
NCCL Tests for AWS ParallelCluster

sean-smith:
* This adds a nccl test script to accompany the workshop: https://catalog.us-east-1.prod.workshops.aws/workshops/6cbbf337-c498-4c6b-ad4f-99d22e03d8dc/en-US/05-nccl-tests

* It assumes NCCL is installed in `/opt/nccl` path and nccl tests are `/opt/nccl-tests`.

* By doing it this way instead of using a docker image it's significantly faster for users to get started

### DIFF
--- a/micro-benchmarks/nccl-tests/slurm/nccl-tests-pcluster.sbatch
+++ b/micro-benchmarks/nccl-tests/slurm/nccl-tests-pcluster.sbatch
@@ -1,0 +1,35 @@
+#!/bin/bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+#SBATCH --job-name=nccl-all_reduce_perf # name of your job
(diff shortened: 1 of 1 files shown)


mhuguesaws (changes requested):
No. We are not going to have a nccl tests per single service.

Either we changed to have 2 scripts for slurm: 1\ ami base and 2\ container based or no change at all.

It is way too much to maintain minimalistic variant of nccl script that will ultimately run on slurm.

sean-smith:
> No. We are not going to have a nccl tests per single service.
> 
> Either we changed to have 2 scripts for slurm: 1\ ami base and 2\ container based or no change at all.
> 
> It is way too much to maintain minimalistic variant of nccl script that will ultimately run on slurm.

Ok, I figured you'd say that so I included some `sed` commands in the NCCL test script so we can use the `nccl-tests-deep-learning-ami.sh` and just change the path to the `all_reduce` binary. Ideally I'll just parameterize this.

PR #329, opened 2024-05-14
NCCL Test Script for AMI

sean-smith:
* This parameterizes the NCCL Tests for Deep Learning AMI so it can be run on other systems

For example, to run on AWS ParallelCluster:

```bash
sbatch nccl-tests-ami.sbatch /opt/nccl-tests/build/all_reduce_perf /opt/nccl/build/lib
```

This passes in the path to the binary as argument 1 and any additional `LD_LIBRARY_PATH` paths to argument 2.


mhuguesaws (approved):
LGTM

PR #334, opened 2024-05-16
EKS CB fix for un-managed Node group

sean-smith:
This adds back the `MarketType` flag which is required for Capacity Blocks. See https://docs.aws.amazon.com/eks/latest/userguide/capacity-blocks.html


nghtm (approved):
approved, LGTM

PR #347, opened 2024-05-22
add metrics on the headnode

sean-smith:
This commit adds `node_exporter` to the headnode so we can see metrics like root volume size. 

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/lifecycle_script.py
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/lifecycle_script.py
@@ -198,6 +198,7 @@ def main(args):
                 wait_for_scontrol()
                 ExecuteBashScript("./utils/install_docker.sh").run()
                 ExecuteBashScript("./utils/install_slurm_exporter.sh").run()
+                ExecuteBashScript("./utils/install_head_node_exporter.sh").run()
                 ExecuteBashScript("./utils/install_prometheus.sh").run()
 
         # Install and configure SSSD for ActiveDirectory/LDAP integration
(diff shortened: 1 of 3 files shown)


1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_head_node_exporter.sh
```diff
@@ -0,0 +1,13 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+
+# Run the Docker container with appropriate configurations
+sudo docker run -d --restart always \
+  --net="host" \
+  --pid="host" \
+  -v "/:/host:ro,rslave" \
+  public.ecr.aws/bitnami/node-exporter:latest \
```
nghtm, line 12: This is a different container than EFA Node Exporter on the compute nodes, which I think makes sense given we dont need EFA metrics for head node, but I don't know how the metrics will change. I will need to test

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/lifecycle_script.py
```diff
@@ -198,6 +198,7 @@ def main(args):
                 wait_for_scontrol()
                 ExecuteBashScript("./utils/install_docker.sh").run()
                 ExecuteBashScript("./utils/install_slurm_exporter.sh").run()
+                ExecuteBashScript("./utils/install_head_node_exporter.sh").run()
```
nghtm, line 201: This makes sense, only install on head node

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_head_node_exporter.sh
nghtm: Tested successfully on a p5 cluster with m5.2xlarge head node.

Screenshot 2024-05-22 at 6 03 35 PM

nghtm (approved):

PR #376, opened 2024-07-15
update MosaicML composer image and MPT test case

KeitaW:
This PR updates MosaicML composer test case to be based on the common PyTorch image


KeitaW:
In the current version of codes, distributed training is failing with the following error:

```
0: terminate called after throwing an instance of 'c10::DistBackendError'
0:   what():  [PG 0 Rank 0] Process group watchdog thread terminated with exception: [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=1, OpType=ALLREDUCE, NumelIn=1, NumelOut=1, Timeout(ms)=600000) ran for 600022 milliseconds before timing out.
```

I will open this PR once I resolve the issue.

PR #377, opened 2024-07-16
Fixing the "breaking at validation_freq" error

amanshanbhag:
Added FI_EFA_USE_HUGE_PAGE=0 to sbatch training script (for p5)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/3.test_cases/10.FSDP/1.distributed-training.sbatch
+++ b/3.test_cases/10.FSDP/1.distributed-training.sbatch
@@ -27,6 +27,7 @@ export FI_EFA_USE_DEVICE_RDMA=1 # use for p4d
 export FI_EFA_FORK_SAFE=1
 export FI_LOG_LEVEL=1
 export FI_PROVIDER=efa
+export FI_EFA_USE_HUGE_PAGE=0    #Use for p4, p5. Comment out for others
 export NCCL_DEBUG=INFO
 ## Switching SYNC_MEMOPS to zero can boost throughput with FSDP
 ## Disables CU_POINTER_ATTRIBUTE_SYNC_MEMOPS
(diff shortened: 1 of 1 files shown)


3.test_cases/10.FSDP/1.distributed-training.sbatch
awsankur, line 1: This is only for P5s I think

KeitaW:
https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/efa-cheatsheet.md
> Disabling huge page causes minor performance hit, but it's needed to prevent fork fails due to the operating system running out of huge pages.

I'm not sure if we should make an option that could impact performance the default setting.

amanshanbhag:
> https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/efa-cheatsheet.md
> 
> > Disabling huge page causes minor performance hit, but it's needed to prevent fork fails due to the operating system running out of huge pages.
> 
> I'm not sure if we should make an option that could impact performance the default setting.

But without this, the FSDP code breaks with fork fail errors (same as the one quoted by you above).

Will test on p4 cluster soon too!

amanshanbhag:
Unable to replicate on p4de cluster. I'll remove the p4de from the comment

3.test_cases/10.FSDP/1.distributed-training.sbatch
```diff
@@ -27,6 +27,7 @@ export FI_EFA_USE_DEVICE_RDMA=1 # use for p4d
 export FI_EFA_FORK_SAFE=1
 export FI_LOG_LEVEL=1
 export FI_PROVIDER=efa
+export FI_EFA_USE_HUGE_PAGE=0    #Use for p5. Comment out for others
```
KeitaW, line 30:
```suggestion
export FI_EFA_USE_HUGE_PAGE=0    # Set to 0 when you see os.fork() causes OSError: Cannot allocate memory. 
```

KeitaW (approved):

PR #383, opened 2024-07-23
Update Llama3 training test case with Trainium on SMHP 

KeitaW:
This PR updates the test case to make it compatible with Neuron SDK 2.19


nghtm (approved):

nghtm:
Review approved. Only comment is why are 16 nodes required, is it possible with 4?

KeitaW:
Thanks @nghtm! It requires at least 8 instances I believe, 16 instances are prepared just to showcase 3D parallelism, with data parallel degree of 2.

PR #393, opened 2024-08-01
Fix device name for LDAP server.

mhuguesaws:
Currently, the LDAP server is created with a secondary volume instead of settings the root volume size.
This is due to Ubuntu uses `/dev/sda1` as root volume instead of `/dev/xvda`

*Issue #, if available:*

*Description of changes:*


mhuguesaws:
@nghtm can it review on this?

PR #394, opened 2024-08-02
Fix jax install

mhuguesaws:
Fix jax install due to jax not pinning package dependency.


KeitaW:
Allow me to merge the PR to create new JAX test case on top of the updated image.

KeitaW (approved):

PR #404, opened 2024-08-09
Add an option to activate CPU Offloading

KeitaW:
You can test 70B model training on single P5 instance with the flag.

```bash
#!/bin/bash

#SBATCH --nodes=1 # number of nodes to use
#SBATCH --job-name=FSDP # name of your job
#SBATCH --exclusive # job has exclusive use of the resource, no sharing

set -ex;

###########################
###### User Variables #####
###########################

GPUS_PER_NODE=8 # 4 for G5.12x, 8 for P4/P5

###########################
## Environment Variables ##
###########################

## Plenty of EFA level variables
## Comment out for non-efa instances (G4d, P3)
## For G5.12x, Comment out RDMA and Fork safe
## For G4dn and other G5, comment out all
export FI_EFA_USE_DEVICE_RDMA=1 # use for p4d
export FI_EFA_FORK_SAFE=1
export FI_LOG_LEVEL=1
export FI_PROVIDER=efa
# export FI_EFA_USE_HUGE_PAGE=0    # Set to 0 when you see os.fork() causes OSError: Cannot allocate memory.  Disabling huge page causes minor performance hit.
export NCCL_DEBUG=INFO
## Switching SYNC_MEMOPS to zero can boost throughput with FSDP
## Disables CU_POINTER_ATTRIBUTE_SYNC_MEMOPS
## Reduces memory synchronizations
## https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__UNIFIED.html
export FI_EFA_SET_CUDA_SYNC_MEMOPS=0

###########################
####### Torch Dist  #######
###########################

declare -a TORCHRUN_ARGS=(
    --nproc_per_node=$GPUS_PER_NODE
    --nnodes=$SLURM_JOB_NUM_NODES
    --rdzv_id=$SLURM_JOB_ID
    --rdzv_backend=c10d
    --rdzv_endpoint=$(hostname)
)

export TORCHRUN=./pt_fsdp/bin/torchrun
export TRAIN_SCRIPT=./train.py

############################
# Llama 2 Training Params ##
############################

declare -a TRAINING_ARGS=(
    --max_context_width=4096
    --num_key_value_heads=8 # 7b: 32 13b: 40 70b: 8
    --intermediate_size=28672 # 7b: 11008 13b: 13824 70b: 28672
    --hidden_width=8192 # 7b: 4096 13b: 5120 70b: 8192
    --num_layers=80 # 7b: 32 13b: 40 70b: 80
    --num_heads=64 # 7b: 32 13b: 40 70b: 64
    --model_type=llama_v2
    --tokenizer="hf-internal-testing/llama-tokenizer"
    --checkpoint_freq=5000
    --validation_freq=500
    --max_steps=5000
    --checkpoint_dir=./checkpoints
    --dataset='c4'
    --dataset_config_name='en'
    --resume_from_checkpoint=./checkpoints
    --train_batch_size=1
    --val_batch_size=1
    --sharding_strategy="full" # https://pytorch.org/docs/stable/fsdp.html
    --offload_activations=1
    --cpu_offload=1
)

AUTO_RESUME=""
    echo "Detected Hyperpod cluster.. enabling --auto-resume=1"
    AUTO_RESUME="--auto-resume=1"
fi

srun ${AUTO_RESUME} -l ${TORCHRUN} "${TORCHRUN_ARGS[@]}" $TRAIN_SCRIPT "${TRAINING_ARGS[@]}"

```


3.test_cases/10.FSDP/model_utils/arguments.py
```diff
@@ -86,6 +86,11 @@ def parse_args():  # pylint: disable=too-many-statements
         choices=["full", "hybrid"],
         help="FSDP sharding strategy https://pytorch.org/docs/stable/fsdp.html",
     )
+    fsdp_grp.add_argument(
+        "--cpu_offload",
+        type=int, defalut=0
+        help="CPU offloading https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.StateDictConfig"
```
KeitaW, line 92:
```suggestion
        help="CPU offloading https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.CPUOffload"
```

amanshanbhag (commented):
“ If True, then this offloads gradients to CPU as well, meaning that the optimizer step runs on CPU.” - does this lead to a performance hit? 

Regardless of the answer, I think it’s a good option to have as a flag, I’m just curious^

KeitaW:
> “ If True, then this offloads gradients to CPU as well, meaning that the optimizer step runs on CPU.” - does this lead to a performance hit?

Using this feature may slow down the training considerably, due to frequent copying of tensors from host to device, but it could help improve memory efficiency and train larger scale models.

> Regardless of the answer, I think it’s a good option to have as a flag, I’m just curious^

I agree with you that specifying it as a flag is more intuitive, however the existing `offload_activations` flag takes [integer value](https://github.com/aws-samples/awsome-distributed-training/blob/7b6ae1cefa4c806432c0b667fdd9c1fbe61e41f0/3.test_cases/10.FSDP/train.py#L221). We may modify both as a proper flag to make the code consistent.

amanshanbhag (approved):
Thanks for answering! LGTM.

PR #408, opened 2024-08-16
Change slurm NCCL test to avoid mangled output.

mhuguesaws:
This PR avoid rank mangled output for NCCL test by removing `-u` options


nghtm (approved):
Approved, minor changes only

PR #411, opened 2024-08-24
Allow FSxL metadata IOPS configuration update

shimomut:
*Description of changes:*

Adding FSxL MetadataConfiguration to allow customers to update FSx metadata IOPS.

* Specifying AUTOMATIC mode in the template, so the performance doesn't change until customer updates the setting to User-provisioned mode.
* Without MetadataConfiguration, FSx doesn't allow configuration change, so we need to include MetadataConfiguration in the LustreConfiguration even with AUTOMATIC mode.


KeitaW:
Confirmed that the FSxL filesystem created with the updated template allow us to update Metadata IOPS a posteriori.
Filesystem created with the updated template:
image
Filesystem created with the current template:
image

KeitaW (approved):

KeitaW:
`s3://awsome-distributed-training/templates/sagemaker-hyperpod.yaml`  have also been updated so that we can create VPC stack with the new CfN template through the workshop.

PR #451, opened 2024-10-07
add nginx

KeitaW:
This PR introduces NGINX proxy server to the LDAP server so that we can use HTTPs access to the LDAP server.


mhuguesaws:
This is in #509 to have certificate. We can close this?

PR #455, opened 2024-10-09
fix nccl test eks

roywei:
*Issue #, if available:*

*Description of changes:*

1. Fixed image pull policy config, to resolve this error during deployment
```
Error from server (BadRequest): error when creating "nccl-tests.yaml": MPIJob in version "v2beta1" cannot be handled as a MPIJob: strict decoding error: unknown field "spec.mpiReplicaSpecs.Launcher.template.spec.imagePullPolicy", unknown field "spec.mpiReplicaSpecs.Worker.template.spec.imagePullPolicy"  yaml file apiVersion: 
```
2. removed unnecessary flags for EFA, these are auto-set in OFI plugin now.

### DIFF
--- a/micro-benchmarks/nccl-tests/kubernetes/nccl-tests.yaml
+++ b/micro-benchmarks/nccl-tests/kubernetes/nccl-tests.yaml
@@ -12,11 +12,11 @@ spec:
       replicas: 1
       template:
          spec:
-          imagePullPolicy: IfNotPresent
           restartPolicy: OnFailure
-          containers:
+          containers: 
           - image: <account>.dkr.ecr.<region>.amazonaws.com/<image>:<tag>
             name: test-nccl-launcher
+            imagePullPolicy: IfNotPresent
             env:
              - name: LD_LIBRARY_PATH
                value: /opt/amazon/openmpi/lib:/opt/nccl/build/lib:/opt/amazon/efa/lib:/opt/aws-ofi-nccl/install/lib:/usr/local/nvidia/lib:$LD_LIBRARY_PATH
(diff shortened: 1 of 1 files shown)


micro-benchmarks/nccl-tests/kubernetes/nccl-tests.yaml
```diff
@@ -39,10 +39,6 @@ spec:
             - -x
             - FI_PROVIDER=efa
             - -x
-            - FI_EFA_USE_DEVICE_RDMA=1
```
mhuguesaws, line 42: Keep those variable to be consistent with slurm.

mhuguesaws (commented):
Left comments

micro-benchmarks/nccl-tests/kubernetes/nccl-tests.yaml
mhuguesaws: @roywei do you plan to update this PR?

roywei: Hi @mhuguesaws thanks for the review and sorry for late reply. I think these flags are not needed anymore per OFI docs, no matter if this is on slurm or k8s. Also if you set `FI_EFA_USE_DEVICE_RDMA` on some old p3 instances with new OFI and libfabric version if may cause crash. See here https://github.com/aws/aws-ofi-nccl/blob/master/doc/efa-env-var.md

roywei (commented):
let me know if i missed anything about the flags. I can also update the slurm nccl-tests flags if needed

micro-benchmarks/nccl-tests/kubernetes/nccl-tests.yaml
mhuguesaws: This test is specific to p5. Please keep as it is or make it consistent in Slurm as well. Thank you

PR #464, opened 2024-10-23
Update pcluster architecture guidance

KeitaW:
*Issue #, if available:*

*Description of changes:*


mhuguesaws:
You have also some docs on IAM identity center setup here -> https://github.com/aws/amazon-vpc-cni-k8s/pull/3047

KeitaW:
Thanks for the info! however, it would be a bit hard for user to jump around the repository and identify the right guidance, so I would prefer to have it in the same doc.

PR #470, opened 2024-10-29
Head node and login node ip in motd

sean-smith:
image


nghtm:
"You are on the controller" but we then say "Head node IP". Lets be consistent in naming convention, ie controller or head

nghtm:
otherwise, approving

nghtm (approved):
approved, left comments

sean-smith:
> "You are on the controller" but we then say "Head node IP". Lets be consistent in naming convention, ie controller or head

Updated to "Controller"

PR #508, opened 2024-12-08
Send SSH public key to cluster

sean-smith:
This commit allows the user to add their SSH public key to to the cluster so they can easily ssh into the cluster.

It works like so:

image

After the user runs this script for the first time they can simply run:

```
$ ssh ml-cluster
```


mhuguesaws (approved):
Approve

PR #514, opened 2024-12-10
easy smhp slurm and eks

gmgtgh:
Automate SageMaker HyperPod Cluster Creation for EKS and Slurm orchestrator

A bash script that automates the manual cluster creation process for SageMaker HyperPod SLURM and EKS


mhuguesaws (changes requested):
fix the PR.

PR #537, opened 2025-01-30
Update megatron lm testcase

KeitaW:
This PR introduces following changes:
* Update container image to the latest
* Separate `slurm` and `kubernetes` instructions


mhuguesaws (changes requested):
Thank you for reorganizing this.
I think it will be better to have Model > Orchestrator. We tend to look for model and then orchestrator rather than the opposite.
What do you think?

KeitaW:
That works for me. Will update the PR shortly...

KeitaW:
Closing in favor of https://github.com/aws-samples/awsome-distributed-training/pull/637/files#diff-c9ce0becb3cdf8cca026132dacd8f30bf6da2eb8c311ed7d7596545bc5ea8418

PR #543, opened 2025-02-04
Adding examples for SMHP EKS task governance

jwoschitz:
*Issue #, if available:*

*Description of changes:* Added a new section on Task Governance for SageMaker HyperPod on EKS, including setup instructions for cluster policies and compute allocations, along with examples demonstrating resource borrowing, guaranteed compute, and priority-based preemption.


nadknish (changes requested):
Reviewed cluster policy, compute quotas, and corresponding job manifest files. Can we add priority class labels (even low priority) to each example? If there are no priority classes then it won't show up in the 'task' section of the Dashboard tab since currently UI filters on priority class.

nadknish (approved):

PR #545, opened 2025-02-05
Fix a crash with empty InstanceGroup in resource_config.json

olegtalalov:
*Issue #, if available:*

*Description of changes:*

With TrainingPlan launch SMHP allows to have InstanceGroups with 0 instances. "Instances" field is set to None in this case and caused a crash in LCS. 


shimomut:
Looks good to me.
Alternative is to use `get(key,default_value)`, instead of or get(key) or default_value.

shimomut (approved):

olegtalalov:
> Looks good to me. Alternative is to use `get(key,default_value)`, instead of `get(key) or default_value`.

It will not work. Because key "Instances" exists with None value.

PR #591, opened 2025-03-18
avoid using ephemeral port in nccom-tests.sbatch

KeitaW:
*Issue #, if available:*

> can you try moving the `NEURON_RT_ROOT_COMM_ID` outside of the ephemeral port range (>62000 instead of PORT=$((RANDOM + 10000)))? The ephemeral port might still be open from the previous run.


jianyinglangaws (approved):
Looks good to me!

PR #611, opened 2025-03-27
Reduce efa exporter container images

mhuguesaws:
This PR changes the container file for efa node exporter to be distroless.
It allows to reduce container image size from ~1.3GB to ~25.5MB.


nghtm (approved):
tested successful build on a g5.8xlarge instance on hyperpod. 

did not test anything around the buildspec.yaml file, what is this files purpose? 

Approving request.

PR #613, opened 2025-03-28
update default mhn db properties

andrew-j-cai:
This PR updates some of the default properties of the RDS DB that customers need to create for HyperPod multi-headnode clusters.
- default instance type changed from t3.medium to **m5.4xlarge**.
- default storage type changed from gp2 to **gp3**.


amanshanbhag:
What’s the reason for the upgrades?

andrew-j-cai:
We conducted load testing recently and found the old settings had some issues in intensive use cases. So to be safe we are upgrading the recommendations a bit. The customer can change these properties freely based on their workload.

amanshanbhag (approved):
LGTM

Issue #632, opened 2025-04-04
Transient fsx_ubuntu.sh failure Due to race condition with mount_fsx.sh


bluecrayon52:
The [fsx_ubuntu.sh](https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/fsx_ubuntu.sh) will sometimes fail because the filesystem isn't yet fully ready for write operations, even though the [mount_fsx.sh](https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx.sh) completed successfully and mounting operations are complete.

nghtm:
Interesting, have not encountered this before. Did you encounter this when deploying a SMHP slurm cluster? What was the size of the cluster?

bluecrayon52:
I was testing the [HP Slurm workshop](https://catalog.workshops.aws/sagemaker-hyperpod/en-US) in a test event account using an instance group of  2 ml.g5.8xlarge instances. Cluster creation failed with the following error:
```
Instance i-0b66516c4a8bfcea4 failed to provision with the following error: "Lifecycle scripts did not run successfully. Ensure the scripts exist in provided S3 path, are accessible, and run without errors. Please see CloudWatch logs for lifecycle script execution details." Note that multiple instances may be impacted.
```
After checking the CloudWatch logs for the instance group, I see that the [fsx_ubuntu.sh](https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/fsx_ubuntu.sh) script, which is called as a subprocess from [lifecycle_script.py](https://github.com/aws-samples/awsome-distributed-training/blob/000a60825f7561e98d31ff0ff56b0484073f3638/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/lifecycle_script.py#L195), is failing with this error:
```
usermod: cannot rename directory /home/ubuntu to /fsx/ubuntu
```
I added a loop to the script, and that seemed to fix things. It was merged in PR #633

nghtm:
I think we need to revisit this, wait condition should not be used here. We should better understand why a node where /mount_fsx.sh is run would still not have write permissions.

PR #622, opened 2025-03-31
Openzfs smhp

amanshanbhag:
*Issue #, if available: #601 *

*Description of changes:*
Adding in FSx OpenZFS to be an **optional** home directory for SMHP SLURM users. Changes included in CF stack template, automate cluster script, and lifecycle scripts. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #633, opened 2025-04-04
updated fsx_ubuntu.sh script with wait loop

bluecrayon52:
*Issue #, if available:*
#632 

*Description of changes:*
I've added a wait loop to [fsx_ubuntu.sh](https://github.com/aws-samples/awsome-distributed-training/blob/fix/fsx_ubuntu/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/fsx_ubuntu.sh) that checks the mounted filesystem for write capability every 10 seconds for 60 seconds. 
I've tested this change, and it seems to resolve the transient race condition with [mount_fsx.sh](https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx.sh)


amanshanbhag (approved):
Thank you. LGTM. Will incorporate these changes into #622 too.

PR #656, opened 2025-04-29
Feat/picotron resume from checkpoint

KeitaW:
*Issue #, if available:*

*Description of changes:*


allela-roy:
@KeitaW the training currently fails when trying to save the checkpoint. Shared the full error log for debugging.

KeitaW:
using docker command to create config is suboptimal due to permission issue (Thanks @allela-roy for point out). PR changed to use enroot command instead. However,  `squashfuse` and `use-overlayfs` are required. Created PR suggesting to install them by default.  https://github.com/aws-samples/awsome-distributed-training/pull/661

KeitaW:
Due to https://github.com/NVIDIA/enroot/issues/130 it would be better to create enroot container manually. Addressed in https://github.com/aws-samples/awsome-distributed-training/pull/656/commits/20f9a4281543c1b296ad51315823dfcce5ecf595

PR #689, opened 2025-05-21
Fix nvidia container toolkit to 1.17.6

mhuguesaws:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_docker.sh
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_docker.sh
@@ -32,7 +32,7 @@ chmod g+s $(which docker)
 systemctl enable docker.service
 systemctl start docker.service
 
-# install nvidia docker toolkit
+# install nvidia docker toolkit, pinning to version 1.17.6-1 due to known issue https://github.com/NVIDIA/nvidia-container-toolkit/issues/1093
 export NVIDIA_CONTAINER_TLK_VERSION="1.17.6-1"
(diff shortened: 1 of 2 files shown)


nghtm:
Can you add a package hold to avoid automatic updates:

```
# Hold the package version to prevent automatic updates
sudo apt-mark hold nvidia-container-toolkit nvidia-container-runtime libnvidia-container-tools libnvidia-container1

```

nghtm:
install_enroot_pyxis.sh also needs to be updated. It runs after install_docker.sh and calls the following:

```
apt_install_with_retry "squashfs-tools parallel libnvidia-container-tools"
```

which runs the risk of updating package versions. 

We should add a check in install_enroot_pyxis as well. working on a feature branch, will push in 10 minutes after testing

mhuguesaws:
> install_enroot_pyxis.sh also needs to be updated. It runs after install_docker.sh and calls the following:
> 
> ```
> apt_install_with_retry "squashfs-tools parallel libnvidia-container-tools"
> ```
> 
> which runs the risk of updating package versions.
> 
> We should add a check in install_enroot_pyxis as well. working on a feature branch, will cut a PR in 10 minutes after testing

so many cross dependencies... Thanks for checking.

amanshanbhag:
Can we add a `requirements.txt` at the start and pin versions as per AMI?

nghtm:
Aman - go enjoy the graduation!

nghtm:
Testing now

nghtm:
LGTM

nghtm (approved):
comments inline

PR #706, opened 2025-05-30
new fsdp dataset: nvidia-deeplearningexamples

mvinci12:
*Issue #, if available:*

*Description of changes:*
due to allenai/c4 dataset error `**[rank3]: FileNotFoundError: gzip://c4-validation.00003-of-00008.json::hf://datasets/allenai/c4@1588ec454efa1a09f29cd18ddd04fe05fc8653a2/en/c4-validation.00003-of-00008.json.gz**`, we are changing dataset to https://huggingface.co/datasets/BEE-spoke-data/Nvidia-DeepLearningExamples/viewer/default/train?views%5B%5D=train


KeitaW:
Internally discussed with @mvinci12 for the motivation, however [Nvidia-DeepLearningExamples](https://github.com/NVIDIA/DeepLearningExamples) seems to be a coding dataset, while [C4 dataset](https://huggingface.co/datasets/allenai/c4) is more general corpus. Can we see if other general corpus, such as [FineWeb](https://huggingface.co/datasets/HuggingFaceFW/fineweb) be used as an alternative?

KeitaW:
Tips: FineWeb dataset does not have train/validation split but you can always split data manually:
ex:

```bash
streaming_train = load_dataset("HuggingFaceFW/fineweb", split="train[:90%]", streaming=True)
streaming_val = load_dataset("HuggingFaceFW/fineweb", split="train[90%:]", streaming=True)
```

cf. https://huggingface.co/docs/datasets/v1.11.0/splits.html

mvinci12:
new commit uses wikimedia/wikipedia dataset (https://huggingface.co/datasets/wikimedia/wikipedia). also added new arguments for users:

--train_split : the specific HF split you'd like to use for training
--val_split : split for validation
--n_samples_train : number of samples you'd like to use for training 
--n_samples_val : number of samples you'd like to use for validation

training and validation split were hardcoded to use "train" and "validation" previously, but new popular datasets only have "train" as a split.

3.test_cases/pytorch/FSDP/src/model_utils/train_utils.py
```diff
@@ -474,10 +474,12 @@ def create_streaming_dataloader(dataset,
                       batch_size=1,
                       max_context_width=4096,
                       workers=4,
-                      split=None):
+                      split=None,
+                      n_samples=6150):
     print(f"dataset={dataset}, name={name}")
     tokenizer = AutoTokenizer.from_pretrained(tokenizer)
     data = load_dataset(dataset, name=name, streaming=True, split=split, trust_remote_code=True).shuffle(42+global_rank)
```
KeitaW: Aren't we getting different shuffle per GPU hence loading different splits per rank?

3.test_cases/pytorch/FSDP/src/model_utils/train_utils.py
```diff
@@ -474,10 +474,12 @@ def create_streaming_dataloader(dataset,
...
+    data = data.take(n_samples)
```
KeitaW: Why do we need to sample here? The resultant dataset seems to be very small.

KeitaW (commented):
added comments

3.test_cases/pytorch/FSDP/src/model_utils/train_utils.py
mvinci12: Yeah, not needed for this dataset. But I'm thinking it should be configurable for larger/custom datasets

mhuguesaws:
Run failure due to

> requests.exceptions.HTTPError: 429 Client Error: Too Many Requests for url: https://huggingface.co/datasets/wikimedia/wikipedia/resolve/b04c8d1ceb2f5cd4588862100d08de323dccfbaa/20231101.en/train-00032-of-00041.parquet

Check results in CI https://github.com/aws-samples/awsome-distributed-training/actions/runs/15500434999/job/43646863450

Let's keep C4 dataset and ask to have a HF_TOKEN created.

nghtm:
@mvinci12 You can find readme instructions for creating HF Token here: https://github.com/aws-samples/awsome-distributed-training/tree/main/3.test_cases/pytorch/picotron#build-environment

mvinci12:
newest commit is back to c4 dataset, and i added an argument to input the huggingface access token. the regression test will probably fail due to no huggingface token in argument but working with @amanshanbhag on this tomorrow

mhuguesaws:
HF token was already in the doc for mistral models since mandatory. HF_TOKEN as environment variable is sufficient.

mhuguesaws:
Let's keep C4 dataset, since the cause was unauthenticated requested to HF. New PR will probably better and we can close this one. Thoughts ?

PR #750, opened 2025-06-17
Observability change target scrapping rate to 1 minute

mhuguesaws:
*Issue #, if available:*

*Description of changes:*


nghtm (approved):

nghtm:
Minor edits, is this to reduce ammount of metrics we scrape?

I think 1 min is reasonable

mhuguesaws:
> Minor edits, is this to reduce ammount of metrics we scrape?
> 
> I think 1 min is reasonable

This is to reduce the frequency to 1 minute. That will avoid overloading the prometheus collector.
This change is only for the 1click solution.

nghtm:
LGTM

PR #770, opened 2025-07-03
Update README.md

hariby:
The term "CR" is used for both ODCR and CBML [[docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservation-overview.html)]. To avoid ambiguity, it is referred as **OD**CR explicitly as "user-procured **On-demand** Capacity Reservations (**OD**CR)" when the `PlacementGroup` should be set to `True`. 

*Issue #, if available:*

*Description of changes:*


KeitaW (approved):

KeitaW:
Thank you!

PR #780, opened 2025-07-09
Add retry logic w/ exponential backoff to get_ip_address() function.

giuseppeporcelli:
*Issue #, if available:*
The issue is that w/  Ubuntu 22.04 AMI the network stack might not be ready when the get_ip_address() function code in the LCS gets executed, thus causing the IP to be resolved to 127.0.0.1 and subsequent failures.

*Description of changes:*
Implemented retry logic w/ exponential backoff in the get_ip_address() function, to wait for the network stack to be initialized.


nghtm:
Thanks for the quick review and merge team!

PR #788, opened 2025-07-17
Resolving "Unsupported block type" error in Terraform

allela-roy:
*Issue #, if available:*

*Description of changes:*

-Fixed Helm provider configuration by changing the `kubernetes` block to use attribute syntax with = sign instead of block syntax, resolving the "Unsupported block type" error. 

- Updated all instances of the deprecated `data.aws_region.current.name` attribute to use `data.aws_region.current.id` to address the deprecation warnings.


bluecrayon52 (approved):
Reviewed and approved.

PR #819, opened 2025-08-11
regionalized hma helm deployment and simplified helm_chart module

bluecrayon52:
*Issue #, if available:*

*Description of changes:*
**TLDR;** you can set the region for the health monitoring agent now to support restricted instance groups for Nova model customization. On a separate note, users now have to clone the Helm chart repo as a prerequisite, but this simplifies the terraform logic and makes behavior more consistent between plan and apply phases. 

- `helm_release.hyperpod` now sets `health-monitoring-agent.region` equal to the target region, as set by `var.region` in the main module, which configures the `aws` provider. 
- `helm_chart` module simplified to avoid `terraform plan` errors when the `/tmp/helm-repo/helm_chart/HyperPodHelmChart/Chart.yaml` file doesn't yet exist.
    **Explanation:** Even though the `null_resource.git_clone` resource did successfully fetch the helm charts before `helm_release.hyperpod` is evaluated during the `terraform apply` phase according to the `depends_on` reference, `null_resource` types are not executed during the `terraform plan` phase, causing this issue. This is primarily a side effect of the helm charts being stored and pulled from a GitHub repo rather than using an OCI registry like GHCR or ECR. This requires running git client commands to stage the helm charts locally. If they were stored in an OCI registry, the `terraform plan` phase would succeed natively since the `helm_release.hyperpod` resource would be able to resolve the `chart` URL. Other work arounds were explored, but they all required adding at least one additional step for the user to execute, so the README instructions have been updated to simply require `git clone https://github.com/aws/sagemaker-hyperpod-cli.git /tmp/helm-repo` as an initial step rather than housing and executing that logic within the `helm_chart` module. With this change, `terraform plan` now runs consistently with `terraform apply`, and the `helm_chart` module code is primed for migrating the helm charts to an OCI registry. 

### DIFF
--- a/1.architectures/7.sagemaker-hyperpod-eks/terraform-modules/README.md
+++ b/1.architectures/7.sagemaker-hyperpod-eks/terraform-modules/README.md
@@ -33,6 +33,7 @@ kubernetes_version = "1.32"
 eks_cluster_name = "my-eks-cluster"
 hyperpod_cluster_name = "my-hp-cluster"
 resource_name_prefix = "hp-eks-test"
+aws_region = "us-west-2"
 availability_zone_id  = "usw2-az2"
 instance_groups = {
     accelerated-instance-group-1 = {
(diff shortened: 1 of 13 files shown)


allela-roy:
LGTM, tested SMHP cluster creation and works fine.

allela-roy (approved):
Tested end-end and works fine.

PR #824, opened 2025-08-15
add default log rotation for Slurm daemon logs

taewonk-amzn:
*Issue #, if available:* None

*Description of changes:*

- adds log rotation for Slurm daemon logs (for HyperPod) via lifecycle scripts
  - broadly follows this [recommendation](https://slurm.schedmd.com/slurm.conf.html#SECTION_LOGGING) from SchedMD
- enables the above by default in the Slurm lifecycle scripts


amanshanbhag (approved):
LGTM

PR #827, opened 2025-08-21
Update containerd config for local pause image and data migration

prajjwal-24:
- Change sandbox_image from registry.k8s.io/pause:3.8 to localhost/kubernetes/pause
- Add containerd data migration from /var/lib/containerd to EBS volume for AL2023


shimomut (approved):
LGTM

PR #838, opened 2025-09-02
Updating CF stack to allow for local zone deployments for GB200

amanshanbhag:
*Issue #, if available:*

*Description of changes:*
This is currently a draft, pending tests. Please wait until testing is complete (ETA: EoD 09/02) until merging. 

GB200 is only supported today in the DFW56 local zone. This CF stack adds a T/F toggle to allow for deployment in the local zone. It:
- Creates and configured a NAT instance (no NAT gateway) for internet access
- Sets up all resources in the local zone (including file system)
- Has the VPC still "span" the region -- with a subnet in the local zone 
- If `IsLocalZone` is set to `false`, then it will automatically revert to original setup (deploying in chosen AZ). Otherwise, it expects a valid local zone in the AZ field. 

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
+++ b/1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
@@ -13,6 +13,7 @@ Description: >
   1.2 TB storage which can be overridden by parameter. A role is also created which
   helps to execute HyperPod cluster operations.
 
+#TODO: DO THIS FOR EKS TOO. 
 
 ####################
 ## Stack Metadata ##
@@ -31,6 +32,7 @@ Metadata:
         Parameters:
           - PrimarySubnetAZ
           - BackupSubnetAZ
+          - IsLocalZone
       - Label:
           default: FSx Lustre configuration
         Parameters:
@@ -59,7 +61,7 @@ Metadata:
       SSMDocumentName:
         default: True/False; Create SSM Session Manager Document. Only set to False if SSM-SessionManagerRunShellAsUbuntu document exists in your account.
       PrimarySubnetAZ:
-        default: Availability zone id to deploy the primary subnets
+        default: Availability zone id to deploy the primary subnets (OR set this to your Local Zone ID if you set IsLocalZone to True. Example use1-dfw2-az1)
       BackupSubnetAZ:
         default: (Optional) Availability zone id to deploy the backup private subnet
       CreateS3Endpoint:
@@ -175,6 +177,14 @@ Parameters:
     Default: 0
     MinValue: 0
     MaxValue: 400000
+  
+  IsLocalZone:
+    Type: String
+    Default: 'false'
+    AllowedValues:
+      - 'true'
+      - 'false'
+    Description: Set to true if you are using a local zone for GB200 (DFW only currently).
 
 ###############################
 ## Conditions for Parameters ##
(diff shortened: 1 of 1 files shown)


1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
```diff
@@ -13,6 +13,7 @@ Description: >
   1.2 TB storage which can be overridden by parameter. A role is also created which
   helps to execute HyperPod cluster operations.
 
+#TODO: DO THIS FOR EKS TOO.
```
nghtm: Remove, this template is just for Slurm, so context will not make sense for users

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
```diff
@@ -279,15 +291,97 @@ Resources:
...
+    Properties:
+      Domain: vpc
+  
+  # For GB200: Create a NAT Instance for the DFW Local Zone and an EIP
+  LocalZoneEIP:
+    Condition: UseNATInstance
     Type: AWS::EC2::EIP
     Properties:
       Domain: vpc
+      NetworkBorderGroup: !Sub "${AWS::Region}-dfw-2"   # TODO: CURRENTLY HARDCODED TO DFW.
```
nghtm: Can we fetch Local Zone ID from somewhere in the stack, or as an input param, to make this dynamic?

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
```diff
@@ -279,15 +291,97 @@ Resources:
...
+    
+  NATInstance:
```
nghtm: Is NAT Instance required for local zone? 

NatGW recommended for HA, resiliency. NAT instance introduced single point of failure?

nghtm (changes requested):

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
amanshanbhag: Yes. NATGW not supported in LZ

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
amanshanbhag: Yes. Like I said, this is a draft atm, will test with hardcoded values and make it dynamic later.

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
amanshanbhag: Yes, in draft mode still.

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
nghtm: Nat GW can be used in IAD, and traffic routed from LZ to the NATGW in IAD. @bluecrayon52 implemented this for EKS VPC stack. Plz consult with him. An AWS managed NAT removes single point of failure and also finger pointing to the CFN stack NAT Instance if customer experiences network outages. It comes with cost of potential latency overhead and traffic. Hrushi Gangur would be good SA to consult on this as well

1.architectures/5.sagemaker-hyperpod/sagemaker-hyperpod.yaml
```diff
@@ -279,16 +306,130 @@ Resources:
...
+    Type: AWS::EC2::EIP
+    Properties:
+      Domain: vpc
+      NetworkBorderGroup: !Sub "${AWS::Region}-dfw-2"   # TODO: CURRENTLY HARDCODED TO DFW.
+    
+  NATInstance:
+    Condition: UseLocalZoneNATInstance
+    Type: AWS::EC2::Instance
+    Properties:
+      ImageId: ami-00ca32bbc84273381    # TODO: CURRENTLY HARDCODED TO IAD -- create map.
```
KeitaW, line 365: Should be a variable.

amanshanbhag:
These have both been tested and validated. I will streamline next week, and we'll be good to merge.

PR #843, opened 2025-09-04
Fixing parameters for template

Madhubalasri-B:
*Issue #, if available:*

*Description of changes:*
Fixing the parameters to use default set value


4.validation_and_observability/4.prometheus-grafana/cluster-observability.yaml
```diff
@@ -11,14 +11,6 @@ Parameters:
     Type: String
     Default: awsome-distributed-training
     Description: The name of the S3 bucket containing the custom resource.
-  CustomResourceFunctionS3Key:
```
shimomut, line 14: Can we keep CustomResourceFunctionS3Key and GrafanaServiceAccountfunctionS3Key?
These are helpful for the flexibility for S3 bucket structure.

shimomut (approved):

PR #884, opened 2025-10-30
Nodes upgraded from Amazon Linux 2 to AL23 fail to create pod sandboxes

anshuman8800:
**Root Cause:** 
AL2 and AL23 have different containerd versions. Persisted AL2 containerd data on `/opt/sagemaker/containerd/data-root` is incompatible with AL23 containerd.

Solution
Modified the LCS script to clean stale containerd data before AL23 setup


shimomut (approved):
LGTM

Issue #897, opened 2025-11-11
[SMHP Slurm] ssh login permissions for new users through automate-cluster-creation script not working


aravneelaws:
`automate-cluster-creation.sh` script has the option to create posix users when setting up the slurm SMHP clusters. But these users are missing the .ssh directory in the $HOME directory with the gen-keypair script not run for these users. The LCS script is currently running only for the ubuntu user currently. 
LCS will need to be updated to run on all users that were created in addition to the ubuntu account. 
This will need to be tested with the OpenZFS home directory to check that the symlink is also created correctly for all users with the correct permissions.

aravneelaws:
PR #899 addressed this.

Issue #896, opened 2025-11-11
[SMHP Slurm LCS] OpenZFS mount issue


aravneelaws:
OpenZFS home directory was not mounted on the compute nodes of the Slurm cluster. It is mounted on the login and controller node but not the compute node. 
The entry is present in `/etc/fstab` but not mounted on the compute node. A simple `ls /home` access did not mount it either. `sudo mount -a` did the trick. 

This needs to be fixed in the LCS.

aravneelaws:
PR #899 addresses this.

PR #899, opened 2025-11-13
aravneel update lcs - openzfs mount and user add permissions

aravneelaws:
*Issue:* #896 #897 

*Description of changes:*
This PR addresses the following issues:
1. OpenZFS mounting issue in compute nodes
2. .ssh permissions issue
3. Adding users during cluster creation with correct permissions
4. easy-ssh script updated with -u flag for logging as a specific user (non-ubuntu user)
5. automate-cluster-creation script print statement updated for SSM preference if openZFS /home directory exists.

The changes has been tested on cluster with openZFS for /home and without openZFS (with only fsxL). 

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
@@ -79,11 +79,13 @@ mount_fs()
     local max_attempts=5
     local attempt=1
     local delay=5
+    local test_file="$OPENZFS_MOUNT_POINT/test_file_$(hostname)"
 
     echo "[INFO] Ensuring $OPENZFS_MOUNT_POINT directory exists..."
     ansible localhost -b -m ansible.builtin.file -a "path=$OPENZFS_MOUNT_POINT state=directory" || true
 
     echo "[INFO] Mounting FSx OpenZFS on $OPENZFS_MOUNT_POINT..."
+    echo "[INFO] Using test file: $test_file"
 
     while (( attempt <= max_attempts )); do
         echo "============================"
@@ -107,13 +109,13 @@ mount_fs()
         ls -la "$OPENZFS_MOUNT_POINT" >/dev/null 2>&1 || true
 
         echo "[STEP] Testing file access (touch)..."
-        if ! ansible localhost -b -m ansible.builtin.file -a "path=$OPENZFS_MOUNT_POINT/test_file state=touch"; then
+        if ! ansible localhost -b -m ansible.builtin.file -a "path=$test_file state=touch"; then
             echo "[WARN] Touch failed — retrying in $delay seconds"
             sleep "$delay"; ((attempt++)); continue
         fi
 
         echo "[STEP] Testing file access (delete)..."
-        if ! ansible localhost -b -m ansible.builtin.file -a "path=$OPENZFS_MOUNT_POINT/test_file state=absent"; then
+        if ! ansible localhost -b -m ansible.builtin.file -a "path=$test_file state=absent"; then
             echo "[WARN] Delete failed — retrying in $delay seconds"
             sleep "$delay"; ((attempt++)); continue
         fi
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx.sh
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx.sh
@@ -166,11 +166,13 @@ mount_fs() {
     local max_attempts=5
     local attempt=1
     local delay=5
+    local test_file="$MOUNT_POINT/test_file_$(hostname)"
 
     echo "[INFO] Ensuring $MOUNT_POINT directory exists..."
     ansible localhost -b -m ansible.builtin.file -a "path=$MOUNT_POINT state=directory" || true
 
     echo "[INFO] Mounting FSx Lustre on $MOUNT_POINT..."
+    echo "[INFO] Using test file: $test_file"
 
     while (( attempt <= max_attempts )); do
         echo "============================"
@@ -193,13 +195,13 @@ mount_fs() {
         ls -la "$MOUNT_POINT" >/dev/null 2>&1 || true
 
         echo "[STEP] Testing file access (touch)..."
-        if ! ansible localhost -b -m ansible.builtin.file -a "path=$MOUNT_POINT/test_file state=touch"; then
+        if ! ansible localhost -b -m ansible.builtin.file -a "path=$test_file state=touch"; then
             echo "[WARN] Touch failed — retrying in $delay seconds"
             sleep "$delay"; ((attempt++)); continue
         fi
 
         echo "[STEP] Testing file access (delete)..."
-        if ! ansible localhost -b -m ansible.builtin.file -a "path=$MOUNT_POINT/test_file state=absent"; then
+        if ! ansible localhost -b -m ansible.builtin.file -a "path=$test_file state=absent"; then
             echo "[WARN] Delete failed — retrying in $delay seconds"
             sleep "$delay"; ((attempt++)); continue
         fi


1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
```diff
@@ -76,21 +76,63 @@ install_nfs_client()
...
+
+        echo "[STEP] Verifying mountpoint..."
+        if ! ansible localhost -b -m ansible.builtin.command -a "mountpoint $OPENZFS_MOUNT_POINT"; then
+            echo "[WARN] Mountpoint verification failed — retrying in $delay seconds"
+            sleep "$delay"; ((attempt++)); continue
+        fi
 
-    retry_with_backoff $MAX_ATTEMPTS $INITIAL_BACKOFF "ansible localhost -b -m ansible.posix.mount -a \"path=$OPENZFS_MOUNT_POINT src=$FSX_OPENZFS_DNS_NAME:/fsx fstype=nfs opts=nfsvers=$NFS_VERSION,_netdev,nconnect=16,x-systemd.automount,x-systemd.requires=network-online.target dump=0 passno=0 state=mounted\""
+        echo "[STEP] Triggering automount..."
+        ls -la "$OPENZFS_MOUNT_POINT" >/dev/null 2>&1 || true
```
PremiumSpider, line 109: Is this a potential silent error due to (...|| true)
Do we need to retry ? can we at least log if there is issue ? 

### Example
```
echo "[STEP] Triggering automount..."
if ! ls -la "$OPENZFS_MOUNT_POINT" >/dev/null 2>&1; then
    echo "[WARN] Could not list mount directory contents"
    # log the issue
fi
```

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
```diff
@@ -76,21 +76,63 @@ install_nfs_client()
...
-    # Create mount point directory if it doesn't exist
-    if [ ! -d "$OPENZFS_MOUNT_POINT" ]; then
-        mkdir -p "$OPENZFS_MOUNT_POINT"
-    fi
+    local max_attempts=5
+    local attempt=1
+    local delay=5
+
+    echo "[INFO] Ensuring $OPENZFS_MOUNT_POINT directory exists..."
+    ansible localhost -b -m ansible.builtin.file -a "path=$OPENZFS_MOUNT_POINT state=directory" || true
```
PremiumSpider, line 85: Suppose we have 2 nodes, if they are all creating a global static *test_file*, do we need to worry about race conditions ?

```
T1: Instance A creates test_file
T2: Instance B creates test_file (same name)
T3: Instance A deletes test_file  
T4: Instance B tries to delete test_file → FAILS (already gone)
T5: Instance B thinks mount is broken, retries entire process
```

Would something like below help ? 

```
TEST_FILE="$OPENZFS_MOUNT_POINT/test_file_${HOSTNAME}_$$"
```

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/gen-keypair-ubuntu.sh
```diff
@@ -2,52 +2,112 @@
```
PremiumSpider, line 2: ### Would it help if started adding Header comments such as 

```
# SSH Key Setup Script for SageMaker HyperPod
# 
# Purpose: 
#   - Configures SSH keys for passwordless communication between cluster nodes
#   - Supports multiple users via shared_users.txt CSV file (username,uid,home)
#   - Creates symlinks between FSx Lustre and OpenZFS storage for unified SSH access
#
# Storage Strategy:
#   - Primary SSH keys stored on FSx Lustre (/fsx/username/.ssh)
#   - OpenZFS home directories (/home/username/.ssh) symlinked to FSx Lustre
#   - Ensures SSH keys persist and are accessible from both storage systems
#
# Multi-User Support:
#   - Always configures default 'ubuntu' user first
#   - Processes additional users from ../shared_users.txt if present
#   - Validates user existence before SSH setup
#   - Generates RSA 4096-bit keypairs with proper permissions (600/644/700)
#
# Cluster Behavior:
#   - Safe to run on multiple nodes simultaneously (|| true for key generation)
#   - Reuses existing keys if present, generates new ones if missing
#   - Adds public keys to authorized_keys for passwordless SSH
#

set -exuo pipefail
```

etc

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
aravneelaws: That is a good catch. I can implement this here. Thanks for catching this.

1.architectures/5.sagemaker-hyperpod/automate-smhp-slurm/automate-cluster-creation.sh
```diff
@@ -995,7 +995,11 @@ configure_cluster_users() {
                         read -p "Hit ENTER once step is complete: "
                         
                         echo -e "\n3. Change the Linux shell profile."
-                        echo -e "   It should have '/bin/bash -c 'export HOME=/fsx/\$(whoami) && cd \${HOME} && exec /bin/bash' in its first and only line"
+                        if [[ "$ENABLE_FSX_OPENZFS" == "true" ]]; then
```
PremiumSpider, line 998: Hi I might be missing context, but is this change mainly for when users log into **HOME** in any of the nodes ?

aravneelaws: Correct. When the cluster is deployed only with fsx lustre, there won't be a /home dir and the HOME_DIR is set to /fsx/USER. When fsx for OpenZFS is used as /home dir, we have to explicitly call out this so the users can set the correct preference in SSM to successfully login as a USER.

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/gen-keypair-ubuntu.sh
aravneelaws: Agreed. But I am going to keep this for a separate PR to improve code documentation

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
aravneelaws: @PremiumSpider This does qualify as a silent error due to (..|| true). But we don't need to log anything here because this is just triggering the automount. The ls command is trying to access the space which will mount the FS. We have explicit verification in the next step with touch and delete. I would leave as it is in my opinion.

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
aravneelaws: @PremiumSpider  This has been implemented now.

aravneelaws:
Common tests that can be added to pipeline include:
1. Checking if filesystem mounting was successful - /fsx/USER and /home/USER (for openZFS) - on all nodes (login/controller/compute).
2. Checking the if .ssh directory exists with correct user permissions.
3. Checking if key-pair and authorized_keys exists within .ssh directory.
4. Checking all these functionality work for all USERs (in addition to the default ubuntu user).

These are the main tests that needs to be performed for the above PR. I have tested these on my end. These can be added to your LCS test pipeline as well.

1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/mount_fsx_openzfs.sh
aravneelaws: Tested this as well and looks good.

PremiumSpider: Thank you for testing on your end as well.

PR #904, opened 2025-11-21
dynamically set Global Batch Size

allela-roy:
removing the static GBS to dynamically set GBS to correspond to num of nodes.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
+++ b/3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
@@ -14,9 +14,18 @@ set -ex;
 ###### User Variables #####
 ###########################
 
-# Parallelism decomposition variables
-: "${TENSOR_PARALLEL:=4}"
-: "${PIPELINE_PARALLEL:=2}"
+# configure TP/PP/GBS based on node count
+if [ $NODES -le 4 ]; then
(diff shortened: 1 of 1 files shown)


3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
```diff
@@ -28,7 +28,10 @@ set -ex;
 : "${SEQ_LENGTH:=2048}"
 : "${MAX_POSITION_EMBEDDINGS:=2048}"
 : "${MICRO_BATCH_SIZE:=1}"
-: "${GLOBAL_BATCH_SIZE:=288}"
+
+# set GBS dynamically
+NODES=${SLURM_JOB_NUM_NODES}
+: "${GLOBAL_BATCH_SIZE:=$((288 * NODES / 2))}"
```
KeitaW: Shouldn't it be based on micro batch size? ex: `global_batch_size = nnodes * ngpus * micro_batch_size` .

allela-roy:
@KeitaW , updated this and also adjusted TP/PP to correspond to node count for optimal performance.Verified with tests across 2,4,8,16 nodes and should scale with higher node counts.

KeitaW:
Thanks @allela-roy . I think that makes more sense, assuming that the model size is constant.

PR #905, opened 2025-11-22
ray dashboard integration improvement

mvinci12:
*Issue #, if available:*

*Description of changes:*

push for more reliable integration of ray dashboards into hyperpod observability stack - by updating `customServiceScrapeTargets` rather than otel collector deployment.

### DIFF
--- a/3.test_cases/pytorch/verl/rlvr/observability/README.md
+++ b/3.test_cases/pytorch/verl/rlvr/observability/README.md
@@ -18,7 +18,7 @@ To integrate Ray metrics into the HyperPod Observability stack, we use the `cust
 
 > **Note:** The following script overwrites spec.customServiceScrapeTargets on the ObservabilityConfig. If you've manually added other scrape targets, either:
 > - add Ray targets manually instead of using the following script
-> - merge your additional customServiceScrapeTargets back into the ObservabilityConfig and re-run the script
+> - run the script then merge your additional customServiceScrapeTargets manually back into the ObservabilityConfig 
 
 Run the provided script to add Ray metrics scraping to the HyperPod ObservabilityConfig:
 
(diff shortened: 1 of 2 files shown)


3.test_cases/pytorch/verl/rlvr/observability/README.md
```diff
@@ -1,31 +1,41 @@
-# Ray Observability for HyperPod EKS
+# Ray Observability for HyperPod
```
KeitaW:
```suggestion
# Ray Observability for HyperPod EKS
```

3.test_cases/pytorch/verl/rlvr/observability/add-ray-metrics.sh
```diff
@@ -0,0 +1,52 @@
...
+    "customServiceScrapeTargets": [
+      {
+        "target": "${RAY_CLUSTER_NAME}-head-svc.${RAY_NAMESPACE}.svc.cluster.local:8080",
+        "metricsPath": "/metrics",
+        "scrapeInterval": 30
+      }
+    ]
+  }
+}
+EOF
```
KeitaW, line 29: If `spec.customServiceScrapeTargets` already has entries (for other workloads), this script will wipe them out and replace them with just the Ray head target.  
https://datatracker.ietf.org/doc/html/rfc7386
> “If the patch is anything other than an object, the result will always be to replace the entire target with the entire patch.

Call this out explicitly in the README.
Something like:
> Note: This script overwrites spec.customServiceScrapeTargets on the ObservabilityConfig. If you’ve manually
 * added other scrape targets, either:
* add Ray targets manually instead of using this script; or
* merge your additional targets back into the array and re‑run the script.

mvinci12: good callout!

PR #906, opened 2025-11-23
Minor Updates for RIG Support with Better UX

bluecrayon52:
*Issue #, if available:*

*Description of changes:*
- When in RIG mode, the `s3_bucket` and `lifecycle_script` modules are disabled since they are not used. 
- When in RIG mode, `instance_group` definitions are ignored, so the customer does not need to set this field to an empty map in their `rig_custom.tfvars` 
- Helm chart values set to the default values in the hyperpod-cli repo for the specific targeted revisions
- Tested in standard and RIG mode for validation


nghtm (approved):
approved, lgtm

Issue #944, opened 2026-02-03
Broken training llama with PyTorch FSDP example on SMHP


aravneelaws:
I was trying to run the llama training with Pytorch FSDP on 8x ml.g5.8xlarge on SMHP EKS cluster. The instructions are available in https://awslabs.github.io/ai-on-sagemaker-hyperpod/docs/eks-blueprints/training/fsdp/fully-sharded-data-parallel. But the container being built end up with torch and torchvision package compatibility issue. The requirements.txt file used in Dockerfile during build has torch version set to 2.7.1. But after the container is built, it ends up with:
```
torch                    2.6.0
torchaudio               2.7.1+cu128
torchvision              0.22.1+cu128
```

When I try a simple import for torch and torchvision, the latter fails with the same error I see when running the FSDP PT job.

```
>>> import torch
/usr/local/lib/python3.10/dist-packages/torch/cuda/__init__.py:61: FutureWarning: The pynvml package is deprecated. Please install nvidia-ml-py instead. If you did not install pynvml directly, please report this to the maintainers of the package that installed pynvml for you.
  import pynvml  # type: ignore[import]
>>> import torchvision
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/dist-packages/torchvision/__init__.py", line 10, in <module>
    from torchvision import _meta_registrations, datasets, io, models, ops, transforms, utils  # usort:skip
  File "/usr/local/lib/python3.10/dist-packages/torchvision/_meta_registrations.py", line 164, in <module>
    def meta_nms(dets, scores, iou_threshold):
  File "/usr/local/lib/python3.10/dist-packages/torch/library.py", line 828, in register
    use_lib._register_fake(op_name, func, _stacklevel=stacklevel + 1)
  File "/usr/local/lib/python3.10/dist-packages/torch/library.py", line 198, in _register_fake
    handle = entry.fake_impl.register(func_to_register, source)
  File "/usr/local/lib/python3.10/dist-packages/torch/_library/fake_impl.py", line 31, in register
    if torch._C._dispatch_has_kernel_for_dispatch_key(self.qualname, "Meta"):
RuntimeError: operator torchvision::nms does not exist
```

When I manually override the torch packages in the container and try to import them, the issue disappears. We will need to restrict the packages to requested versions to start resolving this issue.

aravneelaws:
[...]

PR #949, opened 2026-02-09
Rolling back FSDP requirements to fix llama example

aravneelaws:
Llama FSDP example is broken due to `hyperpod-elastic-agent` requirement of `torch` < 2.7.0 as described in issue #944 . Hence, rolling back to 2.5.1 (previous tested version) to satisfy this criteria. This change has been tested successfully. We will remain in this version until `hyperpod-elastic-agent` upgrade is available with support for `tprch` > 2.7.0.

### DIFF
--- a/3.test_cases/pytorch/FSDP/src/requirements.txt
+++ b/3.test_cases/pytorch/FSDP/src/requirements.txt
@@ -1,6 +1,6 @@
---extra-index-url https://download.pytorch.org/whl/cu128
+--extra-index-url https://download.pytorch.org/whl/cu124
 datasets
-torch==2.7.1
-torchaudio==2.7.1
-torchvision==0.22.1
-transformers==4.53.0
+torch==2.5.1
+torchaudio==2.5.1
+torchvision==0.20.1
+transformers==4.50.3


KeitaW (changes requested):
Don't do this. Blackwell supports CUDA 12.8 or later. Could you please suggest another version which works well on B200? I have few customers relying on it.

aravneelaws:
In that case, the elastic agent needs to be upgraded to support torch >2.7. 
For blackwell (cuda 12.8+), i think torch version that is compatible is 2.7 or more. I can try to see if 2.7.0 works for this and update accordingly.

KeitaW:
Or could we work with SageMaker ST to update the `hyperpod-elastic-agent` dependency? Torch 2.7 is almost one year old.

aravneelaws:
Yes, I have cut a ticket for this. Will share the link shortly

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

paragao:
> Don't do this. Blackwell supports CUDA 12.8 or later. Could you please suggest another version which works well on B200? I have few customers relying on it.



> Don't do this. Blackwell supports CUDA 12.8 or later. Could you please suggest another version which works well on B200? I have few customers relying on it.

+1 to that. Also, we should work on using an updated version of everything that is into this container. EFA, OFI-NCCL, NCCL, CUDA, etc. This is the foundation for workloads to run smoothly and with better performance.

paragao:
## ✅ Version Verification (Critical Gap)

**File:** Dockerfile

**Issue:** The PR changes torch versions in requirements.txt but doesn't verify the correct versions are actually installed in the container. The base image (nccl-tests:latest) comes with pre-installed packages that could conflict.

**Recommendation:** Add a version verification step after pip install:



This catches version mismatches at build time rather than runtime.

paragao:
## ✅ Import Test at Build Time

**File:** Dockerfile

**Issue:** No verification that all dependencies can be imported successfully before runtime.

**Recommendation:** Add an import test as a final build step:



This prevents runtime failures due to binary incompatibilities or missing dependencies.

paragao:
## 📚 Documentation of Version Constraints

**File:** README.md

**Issue:** The PR mentions hyperpod-elastic-agent constraint in the PR description, but this isn't documented in the repository. Users won't understand why specific versions are pinned.

**Recommendation:** Add a section to README.md explaining the version constraints:

paragao:
## 📝 Comment in requirements.txt

**File:** src/requirements.txt

**Issue:** The requirements.txt doesn't explain why specific versions are pinned, making it unclear if/when versions can be upgraded.

**Recommendation:** Add comments explaining the version constraints:

aravneelaws:
Closing this PR since the changes will be addressed in PR #951 .

Issue #959, opened 2026-02-14
Repository clone size is unnecessarily large due to deleted binary in history and untracked large files


KeitaW:
## Problem

The `.git` directory weighs 177 MB despite roughly 72 MB of working tree content, making fresh clones significantly slower than necessary. A single 164 MB pack file containing 12,525 objects accounts for nearly all of this overhead, and `git count-objects -vH` reports 5,599 non-delta objects, indicating a substantial proportion of binary or incompressible content.

## Investigation methodology

The analysis combined several approaches to identify where the bloat originates:

- Measured the `.git` directory size with `du -sh .git` and inspected `ls -lh .git/objects/pack/` to quantify the overhead at 177 MB total, dominated by a single 164 MB pack file.
- Ran `git count-objects -vH` for git's own size accounting, confirming 5,599 non-delta objects — a high count that points to many binary files.
- Enumerated the largest blobs across all history using `git rev-list --objects --all | git cat-file --batch-check` to surface objects that persist in the pack regardless of whether they still exist in the current working tree.
- Scanned the working tree for large files by extension (GIF, ZIP, PNG, tar.bz2, CSV, JSON) using `find` with size sorting to identify current binary bloat.
- Checked `git log --diff-filter=D --summary` to identify large files that were committed and later deleted but still consume space in the object database.
- Verified `.gitattributes` contents and checked `git lfs ls-files` — Git LFS is not configured.
- Measured per-directory sizes with `du -sh */` to understand which areas of the repo contribute most.

## Root cause analysis

The following table lists the sources of bloat in descending order of impact:

| Source | Size | Notes |
|--------|------|-------|
| Deleted `miniconda.sh` at `3.test_cases/9.nemo-multimodal/nemo_configs/miniconda.sh` | 95.8 MB | Added in commit `d0227b2`, deleted in commit `32b3313` (Oct 2023). Still stored in the git object database and downloaded by every clone. This single file accounts for ~54% of the `.git` directory size. |
| Demo GIF files (`automate-smhp-demo.gif`, `automate-smhp-eks-demo.gif`) | ~30 MB | Two animated GIFs used for documentation. |
| Lambda function ZIP (`grafana-service-token-lambda-function.zip`) | 15 MB | Binary archive checked into the tree. |
| PNG screenshots across `0.docs/` and other directories | ~10 MB | Dozens of image files spread across multiple directories. |
[...]

PR #995, opened 2026-02-24
Add QLoRA fine-tuning test case for Qwen3-8B on EKS

yhou-uk:
## Summary

- Add a self-contained test case for QLoRA fine-tuning of Qwen3-8B using DDP + DeepSpeed ZeRO-2 on Amazon EKS with 4x NVIDIA A10G GPUs
- Includes training code, Dockerfile, Kubernetes Job manifest, eksctl cluster config, DeepSpeed configs, notebooks, docs, and numbered deployment scripts
- End-to-end tested: cluster creation, Docker image build, training job deployment verified with loss converging from 2.24 to 1.36

## Test plan

- [x] EKS cluster creation via `0.setup-cluster.sh` (g5.12xlarge with 4x A10G)
- [x] Docker image build and push to ECR via `1.build-image.sh`
- [x] Training job deployment via `2.deploy-training.sh`
- [x] Training runs successfully on 4 GPUs with DDP + DeepSpeed ZeRO-2
- [x] Loss decreasing (2.24 to 1.36 over initial steps)
- [x] GPU memory usage healthy (~5.8GB / 22.3GB per GPU)

### DIFF
--- a/3.test_cases/pytorch/deepspeed/qlora/Dockerfile
+++ b/3.test_cases/pytorch/deepspeed/qlora/Dockerfile
@@ -69,7 +69,6 @@ RUN mkdir -p /workspace/outputs /workspace/hf_cache
 # Set environment variables
 ENV PYTHONPATH=/app
 ENV HF_HOME=/workspace/hf_cache
-ENV TRANSFORMERS_CACHE=/workspace/hf_cache
 ENV PYTHONUNBUFFERED=1
 # Do NOT set CUDA_VISIBLE_DEVICES here — let torchrun / K8s manage GPU visibility
 # DeepSpeed / NCCL settings for multi-GPU communication


3.test_cases/pytorch/qlora/qwen3/README.md
```diff
@@ -0,0 +1,158 @@
+# QLoRA Fine-tuning: Qwen3-8B on Amazon EKS
```
KeitaW: ### Test case should follow the repository's established directory convention

The repository's [test case structure](https://github.com/awslabs/awsome-distributed-training?tab=readme-ov-file#3-test-cases) organizes cases by `<framework>/<library>/<model>/`, and this test case's primary distributed training library is DeepSpeed (with ZeRO-2). There's already an existing [`3.test_cases/pytorch/deepspeed/`](https://github.com/awslabs/awsome-distributed-training/tree/main/3.test_cases/pytorch/deepspeed) directory with its own environment setup instructions (Dockerfile, Enroot/Pyxis for Slurm, etc.) that new DeepSpeed-based test cases should build on.

**Suggestion:** Please review the expected structure in the repo README and relocate the test case to sit under the existing DeepSpeed directory:

```
3.test_cases/pytorch/deepspeed/
├── 0.deepspeed.dockerfile          # existing — shared environment setup
├── 1.build-image.sbatch            # existing
├── README.md                       # existing — shared prep instructions
├── examples_megatron_deepspeed/    # existing test case
└── qlora/                          # ← this PR's content goes here
    ├── Dockerfile                  # QLoRA-specific additions (can layer on the shared base)
    ├── README.md                   # training code overview, prerequisites, usage
    ├── configs/
    ├── src/
    └── kubernetes/                 # EKS-specific manifests and instructions
        ├── README.md               # EKS deployment guide
        ├── eks-cluster.yaml
        ├── qwen3_8b-qlora.yaml
        └── storage.yaml            # StorageClass + PVC for EBS volumes
```

Key benefits:
- Follows the established `<framework>/<library>/<test_case>/` convention used by other test cases in the repo
- Leverages the existing DeepSpeed environment setup in `3.test_cases/pytorch/deepspeed/README.md` rather than duplicating it
- Keeps Kubernetes/EKS-specific instructions in a `kubernetes/` subdirectory (following the pattern described in the repo README: top-level for general setup, subdirectories for service-specific launch instructions)
- Makes the test case discoverable alongside other DeepSpeed examples

3.test_cases/pytorch/qlora/qwen3/notebooks/inference_demo.ipynb
```diff
@@ -0,0 +1,401 @@
+{
```
KeitaW: ### Please do not include `.ipynb` files — use scripts + markdown instead

Notebook files embed cell outputs (including base64-encoded images, training logs, and large text blobs) in JSON, which inflates repository storage, produces noisy diffs that are hard to review, and increases `git clone` size permanently — even if outputs are later stripped, the old objects remain in git history. This repository recently underwent a cleanup to purge large files and migrate to Git LFS (see [#959](https://github.com/awslabs/awsome-distributed-training/issues/959)), so we'd like to avoid reintroducing storage bloat.

**Please convert each notebook to a plain Python script (`.py`) + a companion Markdown document (`.md`):**
- `notebooks/inference_demo.ipynb` → `src/inference_demo.py` + `docs/INFERENCE_GUIDE.md`
- `notebooks/qwen3_qlora_finetuning.ipynb` → `docs/TRAINING_WALKTHROUGH.md` (primarily educational text, so a Markdown file alone would work well)

**Important:** If the `.ipynb` files have already been committed on this branch, please remove them from git history entirely rather than just deleting them in a follow-up commit. A simple `git rm` would leave the notebook blobs in the branch history, and they would be merged into `main` permanently.

To rewrite your branch history and remove the notebooks completely:

```bash
# Install git-filter-repo if you don't have it
pip install git-filter-repo

# From your local clone, on the feat/qlora-qwen3-eks branch:
git filter-repo \
    --path 3.test_cases/pytorch/qlora/qwen3/notebooks/inference_demo.ipynb \
    --path 3.test_cases/pytorch/qlora/qwen3/notebooks/qwen3_qlora_finetuning.ipynb \
    --invert-paths \
    --refs feat/qlora-qwen3-eks \
    --force

# Force-push the rewritten branch
git push origin feat/qlora-qwen3-eks --force-with-lease
```

This ensures no notebook artifacts end up in the repository at all. After rewriting, add the replacement `.py` scripts and `.md` docs as new commits on the clean branch.

3.test_cases/pytorch/qlora/qwen3/0.setup-cluster.sh
```diff
@@ -0,0 +1,237 @@
+#!/bin/bash
```
KeitaW: ### Cluster setup should be a prerequisite, not part of the test case

This script (237 lines), `docs/EKS_SETUP.md` (317 lines), and `kubernetes/eks-cluster.yaml` (134 lines) together account for ~690 lines, roughly 14% of the PR. Cluster provisioning is infrastructure setup that sits outside the scope of a training test case, and the repo already has dedicated architecture templates in [`1.architectures/4.amazon-eks/`](https://github.com/awslabs/awsome-distributed-training/tree/main/1.architectures/4.amazon-eks) for this purpose.

**Suggestion:** Rather than bundling cluster setup into the test case, list the following as prerequisites in the test case README:
- A running EKS cluster with a GPU node group (link to `1.architectures/4.amazon-eks/` or the [EKS getting started guide](https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html))
- NVIDIA device plugin installed
- `kubectl` configured to access the cluster
- An ECR repository for the training image

Then move the EBS storage setup (StorageClass + PVC) into the `kubernetes/` subdirectory as a standalone manifest (e.g., `kubernetes/storage.yaml`) since that's specific to this training job's storage needs. This keeps the test case focused on the QLoRA training itself.

The `0.setup-cluster.sh`, `docs/EKS_SETUP.md`, and `kubernetes/eks-cluster.yaml` files can be removed from the PR.

3.test_cases/pytorch/qlora/qwen3/2.deploy-training.sh
```diff
@@ -0,0 +1,57 @@
...
+
+###########################
+###### User Variables #####
+###########################
+
+export NAMESPACE=${NAMESPACE:-ml-training}
+export NUM_GPUS=${NUM_GPUS:-4}
+
+# IMAGE must be set
+if [ -z "${IMAGE:-}" ]; then
```
KeitaW: ### Deploy script and manifest have a variable mismatch

It looks like the deploy script and the K8s manifest got out of sync. This script asks the user to set `$IMAGE` here, but the manifest template (`kubernetes/qwen3_8b-qlora.yaml:64`) uses `${AWS_ACCOUNT_ID}` and `${AWS_REGION}` instead. Since the script doesn't export those two variables, `envsubst` on line 52 would replace them with empty strings, resulting in an invalid image reference like `.dkr.ecr..amazonaws.com/qwen3-qlora-training:latest`.

Also worth noting: `envsubst` is called without a variable whitelist (line 52), so it would substitute all environment variables (`$HOME`, `$PATH`, etc.) in the YAML too.

**Suggestion:** Either:
- (A) Change the YAML line 64 to `image: ${IMAGE}` and call `envsubst '$IMAGE $NAMESPACE $NUM_GPUS'`, or
- (B) Have the deploy script compute and export `AWS_ACCOUNT_ID` and `AWS_REGION` (matching the YAML's own usage comment on lines 14-15).

3.test_cases/pytorch/qlora/qwen3/kubernetes/qwen3_8b-qlora.yaml
```diff
@@ -0,0 +1,152 @@
...
+# GPU requirements: 4x NVIDIA A10G (g5.12xlarge) for multi-GPU mode,
+# or 1x A10G (g5.4xlarge) for single-GPU mode.
+#
+# Usage:
+#   export AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+#   export AWS_REGION=us-east-1
+#   envsubst < kubernetes/qwen3_8b-qlora.yaml | kubectl apply -f -
+#
+apiVersion: batch/v1
+kind: Job
```
KeitaW: ### Consider converting to a Kubeflow `PyTorchJob`

The manifest currently defines a standard `kind: Job`, but several other files already reference `PyTorchJob` conventions:

| File | Line | Current reference |
|------|------|-------------------|
| `kubernetes/README.md` | 23-25 | Kubeflow Training Operator install step |
| `kubernetes/README.md` | 111 | Pod name `qwen3-qlora-training-master-0` (PyTorchJob naming) |
| `kubernetes/README.md` | 120 | `kubectl delete pytorchjob` |
| `3.cleanup.sh` | 24 | `kubectl delete pytorchjob qwen3-qlora-training` |

Since this repo focuses on distributed training and users may naturally want to scale to multi-node in the future, it would be worth converting the manifest to a Kubeflow `PyTorchJob`. This would:
- Make the existing README and cleanup script references correct as-is
- Give the Training Operator control over `torchrun` orchestration, `MASTER_ADDR`/`MASTER_PORT`/`RANK`/`WORLD_SIZE` injection, and coordinated pod lifecycle
- Provide a natural upgrade path to multi-node training (e.g., 2x g5.12xlarge = 8 GPUs) without needing to rework the manifest
- Better fit the spirit of the `awsome-distributed-training` repository

For the current single-node case, a `PyTorchJob` with `replicas: 1` and `nprocPerNode: "4"` works just as well as a plain Job, and the Training Operator is already referenced in the prerequisites. The conversion is straightforward — the main changes would be updating `apiVersion`, `kind`, and wrapping the pod spec under `pytorchReplicaSpecs.Master`.

3.test_cases/pytorch/qlora/qwen3/3.cleanup.sh
```diff
@@ -0,0 +1,36 @@
...
+###########################
+
+NAMESPACE=${NAMESPACE:-ml-training}
+
+###########################
+###### Delete Job #########
+###########################
+
+echo "Deleting PyTorchJob..."
+kubectl delete pytorchjob qwen3-qlora-training -n ${NAMESPACE} 2>/dev/null || true
```
KeitaW: This line tries to delete a `pytorchjob` resource, but the manifest defines `kind: Job` (batch/v1). Because stderr is suppressed and `|| true` is set, this fails silently and the actual Job keeps running.

If the manifest stays as a `Job`: change to `kubectl delete job qwen3-qlora-training -n "${NAMESPACE}"`.
If converted to `PyTorchJob` per the suggestion above: this line is already correct.

3.test_cases/pytorch/deepspeed/qlora/src/model_setup.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Example:
+        >>> config = get_quantization_config()
+        >>> model = AutoModelForCausalLM.from_pretrained(
+        ...     "Qwen/Qwen3-8B",
+        ...     quantization_config=config
+        ... )
+    """
+    # Convert string dtype to torch dtype
+    compute_dtype = getattr(torch, bnb_4bit_compute_dtype)
```
KeitaW, line 87: ### Consider validating the dtype string before `getattr`

This uses a user-supplied config string as an attribute lookup on `torch`. It would be safer to validate against a known set of dtypes (`float16`, `bfloat16`, `float32`) to give a clear error message for typos or invalid values.

3.test_cases/pytorch/qlora/qwen3/src/model_setup.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Returns:
+        Device map specification for from_pretrained()
+    """
+    local_rank = int(os.environ.get("LOCAL_RANK", -1))
+    world_size = int(os.environ.get("WORLD_SIZE", 1))
+
+    if world_size > 1:
+        # DDP / DeepSpeed: pin model to this rank's GPU
+        device_map = {"": local_rank}
```
KeitaW: ### `device_map` may not work with DeepSpeed ZeRO-3

This works well for ZeRO-2, but would conflict with DeepSpeed ZeRO-3 if a user switches to the provided `deepspeed_zero3.json` config. HuggingFace's Transformers library will actually raise a hard error in this case:

> `ValueError: DeepSpeed Zero-3 is not compatible with low_cpu_mem_usage=True or with passing a device_map.`

Here's why they're incompatible:

- **ZeRO-2** shards only optimizer states and gradients — each GPU still holds a full copy of the model parameters. So `device_map={""': local_rank}` correctly pins the full model to each rank's GPU, which is exactly what ZeRO-2 expects.

- **ZeRO-3** shards the model parameters themselves across all ranks — no single GPU holds the full model. It does this by hooking into parameter access via [`deepspeed.zero.Init()`](https://deepspeed.readthedocs.io/en/latest/zero3.html#deepspeed.zero.Init). Passing `device_map` triggers HF Accelerate's naive pipeline parallelism placement logic, which fundamentally conflicts with DeepSpeed's sharded data parallelism — these are two different parallelism strategies that cannot coexist (see [huggingface/peft#306](https://github.com/huggingface/peft/issues/306)).

- The correct approach for ZeRO-3 is `device_map=None`. The [HuggingFace DeepSpeed docs](https://huggingface.co/docs/transformers/en/deepspeed#initialize-large-models) explain that when using ZeRO-3, `TrainingArguments` must be created **before** calling `from_pretrained()`, and the Trainer's integration with `HfDeepSpeedConfig` handles parameter partitioning automatically — no manual device mapping needed.

Since the PR provides both `deepspeed_zero2.json` and `deepspeed_zero3.json` configs, could you confirm whether you've tested the ZeRO-3 config end-to-end? Given that `_get_device_map()` always passes `device_map={""': local_rank}` in multi-GPU mode, we'd expect this to raise the `ValueError` above when combined with ZeRO-3.

3.test_cases/pytorch/qlora/qwen3/src/data_preparation.py
```diff
@@ -0,0 +1,245 @@
...
+    tokenized = tokenizer(
+        examples["text"],
+        truncation=True,
+        max_length=max_length,
+        padding=False,  # We'll pad in the data collator
+    )
+
+    # For causal LM, labels are the same as input_ids
+    # The model will shift them internally
+    tokenized["labels"] = tokenized["input_ids"].copy()
```
KeitaW: ### Labels include the prompt tokens — consider masking for SFT

For supervised fine-tuning, it's generally recommended to mask the prompt/system tokens in labels (set to `-100`) so the model only trains on generating the response portion. The current approach computes loss over the full sequence, which can reduce training efficiency. The model will still converge, but masking the prompt tokens would likely improve results.

3.test_cases/pytorch/qlora/qwen3/src/train.py
```diff
@@ -0,0 +1,559 @@
...
+    inspect_sample
+)
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+    handlers=[
+        logging.StreamHandler(sys.stdout),
+        logging.FileHandler('training.log')
```
KeitaW: ### Log file may get corrupted in multi-GPU mode

With `torchrun` launching 4 processes, all ranks would write to the same `training.log` file simultaneously, which can produce interleaved output. Also, the relative path means the file location depends on the working directory at import time.

**Suggestion:** Move logging setup into `main()`, use `config.training.output_dir` for the path, and either log to file only from rank 0 or use rank-specific filenames.

3.test_cases/pytorch/qlora/qwen3/src/train.py
```diff
@@ -0,0 +1,559 @@
...
+import argparse
+import logging
+from pathlib import Path
+from typing import Optional
+
+import torch
+from transformers import TrainingArguments, Trainer, EarlyStoppingCallback
+
+# Add parent directory to path for imports
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
KeitaW: ### Missing `src/__init__.py`

The `src/` directory doesn't have an `__init__.py`, so `from src.config import ...` relies on this path manipulation. This can be fragile if the working directory differs from expected (e.g., inside Docker or when `torchrun` changes CWD). Adding an empty `src/__init__.py` would make this more robust.

3.test_cases/pytorch/qlora/qwen3/src/train.py
```diff
@@ -0,0 +1,559 @@
...
+        config = FullConfig.from_yaml(args.config_path)
+        logger.info("Configuration loaded from YAML file")
+    else:
+        config = FullConfig()
+        logger.info("Using default configuration")
+
+    # Override settings from command line
+    if args.output_dir:
+        config.training.output_dir = args.output_dir
+    if args.max_samples:
```
KeitaW: ### `if args.max_samples:` doesn't catch `max_samples=0`

Since `0` is falsy in Python, passing `--max_samples=0` wouldn't apply the override. Using `if args.max_samples is not None:` would handle this edge case.

3.test_cases/pytorch/qlora/qwen3/src/train.py
```diff
@@ -0,0 +1,559 @@
...
+
+    # Save the full model (with merged weights if desired)
+    final_path = f"{config.training.output_dir}/final_model"
+    trainer.save_model(final_path)
+    tokenizer.save_pretrained(final_path)
+    logger.info(f"Full model saved to: {final_path}")
+
+    # Save just the LoRA adapter (much smaller)
+    lora_path = f"{config.training.output_dir}/lora_adapter"
+    model.save_pretrained(lora_path)
```
KeitaW: ### Direct `model.save_pretrained()` may not work with ZeRO-3

Under ZeRO-3, model parameters are sharded across ranks, so calling `model.save_pretrained()` directly could save incomplete weights. The `trainer.save_model()` call just above (line 535) handles this correctly through the Trainer's DeepSpeed integration. It might be worth routing this through the Trainer as well, or wrapping with `deepspeed.zero.GatheredParameters()`.

3.test_cases/pytorch/qlora/qwen3/kubernetes/qwen3_8b-qlora.yaml
```diff
@@ -0,0 +1,152 @@
...
+              value: "/workspace/hf_cache"
+            - name: TRANSFORMERS_CACHE
+              value: "/workspace/hf_cache"
+            # Do NOT set CUDA_VISIBLE_DEVICES — torchrun manages GPU assignment
+            - name: PYTORCH_CUDA_ALLOC_CONF
+              value: "max_split_size_mb:128,expandable_segments:True"
+            # NCCL settings for multi-GPU communication
+            - name: NCCL_DEBUG
+              value: "INFO"
+            - name: NCCL_SOCKET_IFNAME
```
KeitaW: ### `NCCL_SOCKET_IFNAME=eth0` doesn't follow the repo's established pattern

This (and `Dockerfile:79`) hardcode `NCCL_SOCKET_IFNAME=eth0`, which positively selects only the `eth0` interface for NCCL communication. This differs from the pattern used across the rest of the repository, which uses exclusion-based patterns instead. For reference, see the [nccl-tests Dockerfile](https://github.com/awslabs/awsome-distributed-training/blob/17e53d28cdc21789af4396d5589306db21d14727/micro-benchmarks/nccl-tests/nccl-tests.Dockerfile#L111):

```dockerfile
# nccl-tests Dockerfile (line 111):
NCCL_SOCKET_IFNAME=^docker,lo,veth
```

The repo's Kubernetes manifests consistently use `NCCL_SOCKET_IFNAME=^lo` (all FSDP, Megatron, and Neuronx K8s manifests), which excludes only the loopback interface and lets NCCL discover the correct network interface automatically. The positive `eth0` approach is fragile because:
- The interface name can vary depending on the AMI, CNI plugin, or instance type
- On instances with EFA, the EFA interface has a different name (e.g., `ens`, `enp` prefixes) and `eth0` would miss it entirely
- The [EFA cheatsheet](https://github.com/awslabs/awsome-distributed-training/blob/main/1.architectures/efa-cheatsheet.md) in this repo recommends using `en` (prefix match) for p5/p4 instances, not `eth0`

**Suggestion:** Use the exclusion-based pattern consistent with other Kubernetes manifests in the repo:

```yaml
- name: NCCL_SOCKET_IFNAME
  value: "^lo"
```

3.test_cases/pytorch/qlora/qwen3/docs/EKS_SETUP.md
```diff
@@ -0,0 +1,317 @@
...
+│  │               Model Checkpoints & Cache                    ││
+│  └──────────────────────────────────────────────────────────┘│
+└───────────────────────────────────────────────────────────────┘
+```
+
+## Step-by-Step Setup
+
+### Step 1: Create the Cluster
+
+The cluster configuration is in `eks/cluster-config.yaml`. Key settings:
```
KeitaW: ### File paths don't match actual locations

This references `eks/cluster-config.yaml`, but the actual file is `kubernetes/eks-cluster.yaml`. Similarly, lines 314-315 reference `./scripts/build_and_push.sh` and `./scripts/deploy_training.sh` — the actual names are `./1.build-image.sh` and `./2.deploy-training.sh`. Looks like these paths may be from an earlier directory layout.

3.test_cases/pytorch/qlora/qwen3/kubernetes/README.md
```diff
@@ -0,0 +1,143 @@
...
+    ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com
+
+# Build and push (from the qwen3/ directory)
+docker build -t ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/qwen3-qlora-training:latest .
+docker push ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com/qwen3-qlora-training:latest
+```
+
+Or use the provided script:
+```bash
+./0.build-image.sh
```
KeitaW: ### Script names don't match actual filenames

| README says | Actual file |
|-------------|-------------|
| `./0.build-image.sh` | `./1.build-image.sh` |
| `./1.deploy-training.sh` (line 101) | `./2.deploy-training.sh` |
| `./2.cleanup.sh` (line 131) | `./3.cleanup.sh` |

3.test_cases/pytorch/qlora/qwen3/requirements.txt
```diff
@@ -0,0 +1,36 @@
+# Core ML Libraries
+torch>=2.1.0
```
KeitaW: ### Dependencies could benefit from upper-bound version pins

Most dependencies use open-ended `>=` pins (e.g., `torch>=2.1.0`, `peft>=0.10.0`). Adding upper bounds (e.g., `torch>=2.1.0,<2.6.0`) would help ensure reproducible builds.

Also, `torch` appears both here and in the Dockerfile (line 56 with the CUDA index URL) — the `requirements.txt` entry could potentially pull a CPU-only version from PyPI. Consider removing `torch` from `requirements.txt` since the Dockerfile already handles it.

3.test_cases/pytorch/deepspeed/qlora/configs/training_config_zero2.yaml
```diff
@@ -0,0 +1,111 @@
+# Training Configuration for QLoRA Fine-tuning of Qwen3-8B
```
KeitaW, line 1: ### Notebook training config doesn't match this file

The notebook `qwen3_qlora_finetuning.ipynb` (cell 19) shows `per_device_train_batch_size: 2` and `gradient_accumulation_steps: 8`, but this YAML uses `1` and `4` respectively. The comment in this file even notes the batch size was reduced from 2 to prevent OOM. It would be good to sync these so users following the notebook don't run into unexpected memory issues.

KeitaW (commented):
## Review 4/4: Documentation & Dependencies

Several file path references and script names in the docs don't match the actual files, and a few other items. See inline comments below.

**Additional nits** (not worth individual inline comments):

- `0.setup-cluster.sh:14` — Usage header says `setup-cluster.sh`; actual filename is `0.setup-cluster.sh`. Same pattern in other scripts' usage headers.
- `3.cleanup.sh:9` — Usage says `./2.cleanup.sh`; actual is `./3.cleanup.sh`.
- `0.setup-cluster.sh:34-44` — Log helper functions only print `$1`, so extra arguments get dropped. Using `"$*"` would capture everything.
- `kubernetes/README.md:11-14` — References `PLACEHOLDER_*` values that don't exist in `eks-cluster.yaml` (the file uses hardcoded values instead).
- `data_preparation.py:244` — Comment says "first 500 chars" but `[:500]` slices 500 tokens, which would be considerably more text.
- `Dockerfile:36-37` — `apt-get clean` after `rm -rf /var/lib/apt/lists/*` is redundant (the rm already handles it).
- `docs/TROUBLESHOOTING.md:462` — `print(bnb.cuda_available)` should likely be `print(bnb.cuda_available())` (function call).
- `docs/TROUBLESHOOTING.md:481` — References "Section 11 in the inference notebook" which doesn't exist.
- `docs/QLORA_EXPLAINED.md:321` — The bitsandbytes repo link points to the old `TimDettmers/bitsandbytes` URL; the project has moved to `bitsandbytes-foundation/bitsandbytes`.

KeitaW (changes requested):
Left some comments.

3.test_cases/pytorch/qlora/qwen3/requirements.txt
KeitaW: Make sure to that user can run the usecase on the latest hardware (ex, RTX Pro 6000). CUDA 12.8 or later is a current good starting point.

3.test_cases/pytorch/qlora/qwen3/0.setup-cluster.sh
paragao: +1 to Keita. Plus, let's make sure one of the setup options is SageMaker HyperPod not only Amazon EKS. 

Please, add instructions on how to setup MIG for this use case too.

3.test_cases/pytorch/qlora/qwen3/README.md
yhou-uk: Done/ Entire project moved from pytorch/qlora/qwen3/ to pytorch/deepspeed/qlora/.

3.test_cases/pytorch/qlora/qwen3/2.deploy-training.sh
yhou-uk: Done. 1.deploy-training.sh now passes an explicit variable list to envsubst (e.g., '$IMAGE $NAMESPACE $NUM_GPUS ...').

3.test_cases/pytorch/qlora/qwen3/kubernetes/qwen3_8b-qlora.yaml
yhou-uk: Done. Both qwen3_8b-qlora-zero2.yaml and qwen3_8b-qlora-zero3.yaml are Kubeflow PyTorchJob with Master + Worker replicas for multi-node support.

3.test_cases/pytorch/qlora/qwen3/3.cleanup.sh
yhou-uk: 2.cleanup.sh deletes pytorchjob resources, matching the new manifests.

3.test_cases/pytorch/deepspeed/qlora/src/model_setup.py
yhou-uk: Done. src/model_setup.py validates the dtype string before getattr(torch, ...).

3.test_cases/pytorch/qlora/qwen3/src/model_setup.py
yhou-uk: Done. src/model_setup.py uses device_map=None when DeepSpeed is active, allowing ZeRO-3 to manage parameter placement.

3.test_cases/pytorch/qlora/qwen3/src/data_preparation.py
yhou-uk: Done. src/data_preparation.py masks prompt/system tokens with -100 so loss is only computed on the assistant response.

3.test_cases/pytorch/qlora/qwen3/src/train.py
yhou-uk: Done. src/train.py logging is rank-0 only and writes to config.training.output_dir.

3.test_cases/pytorch/qlora/qwen3/src/train.py
yhou-uk: Added, removing the need for sys.path.insert() hacks.

3.test_cases/pytorch/qlora/qwen3/src/train.py
yhou-uk: Changed to is not None check.

3.test_cases/pytorch/qlora/qwen3/src/train.py
yhou-uk: Final model save routes through the Trainer / uses proper DeepSpeed parameter gathering.

3.test_cases/pytorch/qlora/qwen3/kubernetes/qwen3_8b-qlora.yaml
yhou-uk: Changed to NCCL_SOCKET_IFNAME=^lo (exclusion-based) in all manifests and sbatch scripts.

3.test_cases/pytorch/qlora/qwen3/docs/EKS_SETUP.md
yhou-uk: Done. File removed entirely, replaced by kubernetes/README.md with correct references.

3.test_cases/pytorch/qlora/qwen3/kubernetes/README.md
yhou-uk: Renumbered scripts to 0.build-image.sh, 1.deploy-training.sh, 2.cleanup.sh and all docs reference the correct names.

3.test_cases/pytorch/qlora/qwen3/requirements.txt
yhou-uk: All dependencies have upper bounds.

yhou-uk: Torch removed from requirements.txt,  installed separately in Dockerfile (cu128)

3.test_cases/pytorch/deepspeed/qlora/configs/training_config_zero2.yaml
yhou-uk: Notebooks removed, config YAML has inline comments explaining the values (per_device_train_batch_size: 1, gradient_accumulation_steps: 4)

3.test_cases/pytorch/qlora/qwen3/0.setup-cluster.sh
yhou-uk: Done. Full HyperPod support for both EKS and Slurm orchestrators, with auto-resume annotations, health-aware scheduling, and FSx for Lustre storage.
MIG configuration instructions are added. slurm/README.md has a full MIG section (step-by-step setup, supported profiles table, auto-detection in sbatch scripts, parallel experiments). kubernetes/README.md covers MIG for EKS (single + mixed strategy with MIG_PROFILE env var).

yhou-uk:
Hi @KeitaW and @paragao ,  thank you so much for your thorough and valuable reviews and comments. Sorry I resolved some comments by mistake, please check the resolved conversations as well. 

Here is a summary of all the changes I have made:                                                                                                                                                                                                         
                                                                                                                                                                                                                
  - Relocated project from pytorch/qlora/qwen3/ to pytorch/deepspeed/qlora/ per repo convention
  - Purged .ipynb files from git history via git-filter-repo,  replaced with src/inference_demo.py and docs/QLORA_EXPLAINED.md
  - Converted Kubernetes manifests from kind: Job to Kubeflow PyTorchJob with multi-node Master/Worker support
  - Fixed envsubst to use explicit variable whitelist, standardized on ${IMAGE} in manifests
  - Training code fixes: device_map=None for ZeRO-3, dtype validation, prompt token masking (-100), rank-0-only logging, is not None for max_samples, DeepSpeed-aware model saving, NCCL_SOCKET_IFNAME=^lo
  - Added src/__init__.py to eliminate sys.path hacks
  - Added upper bounds to all dependency pins in requirements.txt, removed torch (installed separately per deployment path)
  - Fixed all stale doc references, script name mismatches, bnb.cuda_available typo, and outdated bitsandbytes repo link
  - Added full SageMaker HyperPod support for both EKS and Slurm with auto-resume, health-aware scheduling, FSx for Lustre, and MIG instructions
  - Upgraded Dockerfile to CUDA 12.8.1 with torch+cu128 
  - Added ZeRO-3 variants (configs, manifests, sbatch scripts) for all deployment paths

Thank you!

3.test_cases/pytorch/deepspeed/qlora/entrypoint.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# Entrypoint for the QLoRA training container.
```
KeitaW, line 6: ### Missing copyright header

All shell scripts in the repo must include the standard copyright header. This file starts with `#!/bin/bash` followed by a descriptive comment, but the copyright/license block is missing.

```suggestion
#!/bin/bash

# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT-0

# Entrypoint for the QLoRA training container.
```

3.test_cases/pytorch/deepspeed/qlora/0.build-image.sh
```diff
@@ -0,0 +1,62 @@
...
+    aws ecr create-repository \
+        --repository-name ${REPO_NAME} \
+        --region ${AWS_REGION} \
+        --image-scanning-configuration scanOnPush=true
+
+###########################
+###### Authenticate #######
+###########################
+
+echo "Authenticating to ECR..."
```
KeitaW: ### Unquoted variable expansions

Several variable expansions are unquoted (`${REPO_NAME}`, `${AWS_REGION}`). While these are unlikely to contain spaces in practice, the repo convention is to always quote variable expansions in shell scripts (`"${VAR}"`).

```suggestion
echo "Creating ECR repository (if not exists)..."
aws ecr describe-repositories --repository-names "${REPO_NAME}" --region "${AWS_REGION}" 2>/dev/null || \
    aws ecr create-repository \
        --repository-name "${REPO_NAME}" \
        --region "${AWS_REGION}" \
        --image-scanning-configuration scanOnPush=true
```

3.test_cases/pytorch/deepspeed/qlora/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+COPY src/ /app/src/
+COPY configs/ /app/configs/
+
+# Create directories for outputs and cache
+RUN mkdir -p /workspace/outputs /workspace/hf_cache
+
+# Set environment variables
+ENV PYTHONPATH=/app
+ENV HF_HOME=/workspace/hf_cache
+ENV TRANSFORMERS_CACHE=/workspace/hf_cache
```
KeitaW: ### Deprecated `TRANSFORMERS_CACHE` env var

`TRANSFORMERS_CACHE` has been deprecated in favor of `HF_HOME` (already set on the line above). Since `HF_HOME` is sufficient, this line can be removed to avoid confusing users into thinking both are needed.

The same applies to the `TRANSFORMERS_CACHE` entries in `kubernetes/qwen3_8b-qlora-zero2.yaml` (Master and Worker env sections) and `kubernetes/qwen3_8b-qlora-zero3.yaml` (Master and Worker env sections).

Reference: [HuggingFace environment variables](https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables)

```suggestion
```

PR #997, opened 2026-02-25
Add GPT-OSS 20B LoRA + GRPO finetuning recipe

mijanur132:
## Purpose

Fine-tune GPT-OSS 20B with LoRA and GRPO for multilingual reasoning on Amazon SageMaker HyperPod EKS. Includes SFT training, GRPO reinforcement learning, inference, and evaluation scripts. 

## Changes

New test case. 

## Test Plan

**Environment:**
- AWS Service: AWS Sagemaker Hyperpod, EKS 
- Instance type: ml.g6e.12xlarge
- Number of nodes: 4

**Test commands:**
```bash

```

## Directory Structure

```
3.test_cases/
└── pytorch/                # e.g. pytorch, megatron, jax
    └── trl/              # e.g. picotron, FSDP, megatron-lm
        └── gpt-oss            # e.g. SmolLM-1.7B (may be omitted for single-model cases)
            ├── artifacts     # src code, config
            ├── README.md       # Overview, prerequisites, usage
            └── train/   # training manifests
            └── inference/   # inference manifests
            └── Evaluaiton/   # Evaluation manifests

```

- Top-level files (`Dockerfile`, `README.md`, training scripts, configs) cover general setup.
- Subdirectories (`slurm/`, `kubernetes/`, `hyperpod-eks/`) contain service-specific launch instructions.
- Not all service subdirectories are required — include only the ones relevant to your test case.

## Checklist

- [* ] I have read the [contributing guidelines](https://github.com/awslabs/awsome-distributed-training/blob/main/CONTRIBUTING.md).
- [ *] I am working against the latest `main` branch.
- [ *] I have searched existing open and recently merged PRs to confirm this is not a duplicate.
- [ *] The contribution is self-contained with documentation and scripts.
- [ *] External dependencies are pinned to a specific version or tag (no `latest`).
- [ *] A README is included or updated with prerequisites, instructions, and known issues.
- [ *] New test cases follow the [expected directory structure](#directory-structure).

PR #1000, opened 2026-02-26
Restructure trl/ test cases: shared base image and descriptive naming

KeitaW:
## Summary

- Rename `grpo/` to `grpo-math-reasoning/` for descriptive test case naming
- Create a unified `trl/Dockerfile` with all TRL example dependencies (shared base image)
- Add a library-level `trl/README.md` with test case table and build instructions
- Delete per-test-case `grpo.Dockerfile` (replaced by shared base)
- Update all path references in `train.sbatch` and test case `README.md`

This establishes the pattern for future TRL test cases to share a single base image rather than duplicating Dockerfiles.

> **Merge order:** This PR should be merged **before** #997 is updated. Once landed, #997 can rebase onto `main` and add its new test case (`gpt-oss-lora-grpo/`) alongside `grpo-math-reasoning/`, reusing the shared `trl/Dockerfile` instead of duplicating dependencies.

## Test plan

- [ ] Verify `git log --follow` shows rename history for moved files
- [ ] Verify no references to old `grpo/` or `grpo.Dockerfile` remain
- [ ] Verify `trl/Dockerfile` builds successfully
- [ ] Verify all srun commands in README and sbatch use `trl-base.sqsh` and `/grpo-math-reasoning/` mount paths

### DIFF
--- a/3.test_cases/pytorch/trl/README.md
+++ b/3.test_cases/pytorch/trl/README.md
@@ -13,15 +13,6 @@ cd 3.test_cases/pytorch/trl
 docker build -t trl-base:latest .
 ```
 
-### HyperPod EKS
-
-For running on Amazon SageMaker HyperPod EKS clusters, use the extended image in [`hyperpod-eks/Dockerfile`](hyperpod-eks/Dockerfile) which adds the `hyperpod-elastic-agent`:
-
-```bash
-cd hyperpod-eks
-docker build -t trl-hyperpod:latest .
-```
-
 ## Test Cases
 
 | Test Case | Description | Platform | Model |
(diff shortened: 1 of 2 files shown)


3.test_cases/pytorch/trl/grpo-math-reasoning/train.sbatch
```diff
@@ -1,8 +1,10 @@
...
-#SBATCH --job-name=grpo
+#SBATCH --job-name=grpo-math-reasoning
 #SBATCH --nodes=9
 #SBATCH --ntasks-per-node 1
 
+set -ex
+
 ## Set libfabric flags to use EFA
 export FI_PROVIDER=efa
 export FI_EFA_FORK_SAFE=1
```
paragao: we don't need this anymore, since EFA 1.43 (if I'm not mistaken). It used to resolve a os.fork() issue that got fixed. Should be remove to avoid cluttering exported variables and avoid adding complexity when troubleshooting.

PR #1013, opened 2026-03-10
Add NCCL send/recv ring benchmark for multi-GPU testing

paulogallotti:
## Purpose

Minimal point-to-point benchmark using ncclSend/ncclRecv in a ring topology. Measures latency and bandwidth across GPUs over 100 timed iterations with a 64 MB buffer.

Includes a Slurm batch script configured for SageMaker HyperPod clusters with EFA networking (p4d.24xlarge, 2 nodes, 8 GPUs/node).

### DIFF
--- a/micro-benchmarks/nccl-sendrecv/README.md
+++ b/micro-benchmarks/nccl-sendrecv/README.md
@@ -2,6 +2,10 @@
 
 A minimal NCCL point-to-point benchmark that measures send/recv latency and bandwidth across GPUs in a ring topology. Designed to run on AWS SageMaker HyperPod clusters with Slurm and EFA.
 
+## Why this exists
+
+The repo already includes [`micro-benchmarks/nccl-tests/`](../nccl-tests/), which wraps NVIDIA's `nccl-tests` suite (including `sendrecv_perf`). This standalone benchmark is a minimal, self-contained alternative — no `nccl-tests` build dependency, a single C file, and easy to modify. It is intended as a starting point for users who want to understand or extend NCCL send/recv patterns without pulling in the full suite.
+
 ## What it does
 
 Each MPI rank sends a 64 MB buffer to its right neighbour and receives from its left neighbour in a ring pattern using `ncclSend`/`ncclRecv` grouped calls. After a warm-up phase, it runs 100 timed iterations and reports average latency and bandwidth.
@@ -81,6 +85,6 @@ Message size    : 67108864 bytes (64 MB)
 Iterations      : 100
 Avg latency     : 2.450 ms
 Algo BW (1-dir) : 27.39 GB/s
-Bus BW (bidir)  : 54.78 GB/s
+Bidir throughput : 54.78 GB/s
 -----------------------------------------------
 ```
--- a/micro-benchmarks/nccl-sendrecv/nccl_sendrecv_test.c
+++ b/micro-benchmarks/nccl-sendrecv/nccl_sendrecv_test.c
@@ -118,11 +118,18 @@ int main(int argc, char *argv[])
         total_ms += (double)ms;
     }
 
+    /* Use the slowest rank's total time for an accurate end-to-end picture */
+    double max_total_ms;
+    MPI_Reduce(&total_ms, &max_total_ms, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
+
     /* Report from rank 0 */
-    double avg_ms = total_ms / iterations;
-    double bw_gbps = (2.0 * msg_size) / (avg_ms * 1e-3) / 1e9;
-    /* Algorithm BW (single direction) */
+    double avg_ms = max_total_ms / iterations;
+    /* Algorithmic BW: single-direction throughput (msg_size / time) */
     double algo_bw_gbps = (double)msg_size / (avg_ms * 1e-3) / 1e9;
+    /* Bidirectional throughput: 2x algo BW since each rank sends and receives
+       simultaneously.  Note: this is NOT the same as nccl-tests "busBw", which
+       normalises by the collective's communication pattern. */
+    double bidir_gbps = (2.0 * msg_size) / (avg_ms * 1e-3) / 1e9;
 
     if (rank == 0) {
         printf("---------- NCCL Send/Recv Ring Test ----------\n");
@@ -132,7 +139,7 @@ int main(int argc, char *argv[])
         printf("Iterations      : %d\n", iterations);
         printf("Avg latency     : %.3f ms\n", avg_ms);
         printf("Algo BW (1-dir) : %.2f GB/s\n", algo_bw_gbps);
-        printf("Bus BW (bidir)  : %.2f GB/s\n", bw_gbps);
+        printf("Bidir throughput : %.2f GB/s\n", bidir_gbps);
         printf("-----------------------------------------------\n");
     }
 


micro-benchmarks/nccl-sendrecv/Makefile
```diff
@@ -0,0 +1,21 @@
+CUDA_HOME  ?= /usr/local/cuda
```
KeitaW, line 4: ### Missing license header

The repo convention requires every file to include the Amazon copyright header. The existing nccl-tests scripts (e.g., `nccl-tests-ami.sbatch`) all follow this pattern. Could you add headers to all four files?

For the Makefile and sbatch, use `#` comments; for the `.c` file use `//` or `/* */`.

```suggestion
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT-0

CUDA_HOME  ?= /usr/local/cuda
```

micro-benchmarks/nccl-tests/slurm/nccl-tests-c/run_nccl_test.sbatch
```diff
@@ -0,0 +1,56 @@
...
+#SBATCH --job-name=nccl-sendrecv-test
+#SBATCH --nodes=2
+#SBATCH --ntasks-per-node=8
+#SBATCH --gpus-per-node=8
+#SBATCH --exclusive
+#SBATCH --output=nccl_test_%j.out
+#SBATCH --error=nccl_test_%j.err
+#SBATCH --wait-all-nodes=1
+
+###############################################################################
```
KeitaW, line 16: ### Missing `set -ex`

The existing nccl-tests sbatch scripts (e.g., `nccl-tests-ami.sbatch`) include `set -ex` immediately after the `#SBATCH` directives and copyright header. This ensures the script fails fast on errors and logs each command for debugging. I'd suggest adding it here to follow the established pattern.

```suggestion
#SBATCH --wait-all-nodes=1

set -ex
```

micro-benchmarks/nccl-tests/slurm/nccl-tests-c/run_nccl_test.sbatch
```diff
@@ -0,0 +1,56 @@
...
+make -C "${WORKDIR}"
+
+# ---------- Run ----------
+echo "=== Running NCCL Send/Recv test ==="
+echo "Nodes: ${SLURM_JOB_NUM_NODES}, Tasks: ${SLURM_NTASKS}, GPUs/node: ${SLURM_GPUS_PER_NODE}"
+
+srun \
+    --mpi=pmix \
+    --gpus-per-node=${SLURM_GPUS_PER_NODE} \
+    --ntasks-per-node=${SLURM_NTASKS_PER_NODE} \
```
KeitaW: ### Unquoted variable expansions

Per repo conventions, variable expansions should be quoted. While Slurm sets these to numeric values (so word-splitting isn't a practical risk here), quoting is the repo-wide standard.

```suggestion
srun \
    --mpi=pmix \
    --gpus-per-node="${SLURM_GPUS_PER_NODE}" \
    --ntasks-per-node="${SLURM_NTASKS_PER_NODE}" \
```

micro-benchmarks/nccl-sendrecv/run_nccl_test.sbatch
```diff
@@ -0,0 +1,56 @@
...
+#SBATCH --output=nccl_test_%j.out
+#SBATCH --error=nccl_test_%j.err
+#SBATCH --wait-all-nodes=1
+
+###############################################################################
+# NCCL Send/Recv Ring Test — SageMaker HyperPod (Slurm + EFA)
+#
+# Current config: p4d.24xlarge (8 A100 GPUs/node, 2 nodes)
+###############################################################################
+
```
KeitaW, line 21: ### HyperPod auto-resume detection

Since the README explicitly targets SageMaker HyperPod, it might be worth including the standard HyperPod auto-resume detection block that other HyperPod-targeted scripts in the repo use:

```bash
    # HyperPod auto-resume logic
fi
```

Not strictly required, but would be consistent with the repo convention.

micro-benchmarks/nccl-tests/slurm/nccl-tests-c/run_nccl_test.sbatch
```diff
@@ -0,0 +1,56 @@
...
+export FI_PROVIDER=efa
+export FI_EFA_FORK_SAFE=1
+
+# ---------- NCCL tuning for AWS EFA ----------
+export NCCL_DEBUG=WARN
+export NCCL_SOCKET_IFNAME=^lo,docker0
+
+# Let NCCL auto-select the best algorithm and protocol
+# (previously forced Ring/Simple which limited performance)
+
```
KeitaW: ### Changelog-style comment in a fresh contribution

This reads like a changelog entry ("previously forced Ring/Simple which limited performance") which could be confusing for readers since there's no prior version to compare against. I'd suggest simplifying:

```suggestion
# Let NCCL auto-select the best algorithm and protocol
# for optimal performance across different message sizes.
```

micro-benchmarks/nccl-sendrecv/run_nccl_test.sbatch
```diff
@@ -0,0 +1,56 @@
...
+# Increase cross-node channels to better utilise 4x EFA NICs
+export NCCL_MIN_NCHANNELS=4
+export NCCL_MAX_NCHANNELS=16
+
+# Larger buffers for 64MB messages
+export NCCL_BUFFSIZE=8388608
+
+# Use all 4 EFA interfaces on p4d.24xlarge
+export NCCL_NET_GDR_LEVEL=SYS
+
```
KeitaW, line 44: ### Brief comment on GDR level

It might be worth adding a note that this setting enables GPU Direct RDMA and assumes EFA is available on the instance, e.g.:

```suggestion
export NCCL_NET_GDR_LEVEL=SYS   # GPU Direct RDMA at system level (requires EFA)
```

micro-benchmarks/nccl-sendrecv/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+## Configuration
+
+Compile-time constants in `nccl_sendrecv_test.c`:
+
+| Define | Default | Description |
+|---|---|---|
+| `MSG_SIZE` | 64 MB | Buffer size per send/recv |
+| `ITERATIONS` | 100 | Timed iterations |
```
KeitaW, line 74: ### Example output could note values are illustrative

The specific bandwidth numbers shown (27.39 GB/s algo BW, 54.78 GB/s bus BW) are plausible for p4d but users might interpret them as expected baselines. It might be worth adding a brief note like:

> *Values will vary based on instance type, message size, and network conditions.*

KeitaW:
any update?

paragao:
@paulogallotti a few notes:
- make sure you reuse the same [container as the all_reduce_perf](https://github.com/awslabs/awsome-distributed-training/blob/main/micro-benchmarks/nccl-tests/nccl-tests.Dockerfile) use case
-use the recommended NCCL flags, as you can see on [this sbatch script]().

paulogallotti:
I'm working on the changes

paulogallotti:
> @paulogallotti a few notes:
> 
>     * make sure you reuse the same [container as the all_reduce_perf](https://github.com/awslabs/awsome-distributed-training/blob/main/micro-benchmarks/nccl-tests/nccl-tests.Dockerfile) use case

@paragao just to confirm if I understood your first comment. The script I created is not using containers. Do you want me to add a container based version similar to what exists today for nccl-tests for slurm? I had to move nccl-sendrecv to a separate folder as requested by @KeitaW.

>       -use the recommended NCCL flags, as you can see on this sbatch script.

Can you confirm the link to the recommended flags? Current link seems to be pointing to the same PR URL

paulogallotti:
@KeitaW addressed your feedback in my latest commit.

aravneelaws:
@KeitaW can you review the changes Paulo has incorporated and let me know if this can be merged?

KeitaW:
Let me check today. Thanks both!

micro-benchmarks/nccl-sendrecv/README.md
```diff
@@ -0,0 +1,86 @@
+# NCCL Send/Recv Ring Test
+
+A minimal NCCL point-to-point benchmark that measures send/recv latency and bandwidth across GPUs in a ring topology. Designed to run on AWS SageMaker HyperPod clusters with Slurm and EFA.
```
KeitaW, line 3: ### Consider noting how this relates to `nccl-tests sendrecv_perf`

Since the repo already has `micro-benchmarks/nccl-tests/` which wraps NVIDIA's nccl-tests (including `sendrecv_perf`), it might be worth adding a sentence here — or in a "Motivation" section — explaining why a standalone custom implementation is valuable (e.g., simpler build, easier to modify, no nccl-tests dependency, serves as a code example for NCCL API usage).

KeitaW (commented):
## Review Batch 1/2 — Structure & Repository Hygiene

*Re-review of PR #1013. Previous review items (directory placement, license headers, `set -ex`, HyperPod auto-resume, NCCL env var comments) have all been addressed — nice work.*

### Overlap with existing `nccl-tests` benchmark

The existing [`micro-benchmarks/nccl-tests/`](https://github.com/awslabs/awsome-distributed-training/tree/main/micro-benchmarks/nccl-tests) directory wraps the standard [NCCL Tests](https://github.com/NVIDIA/nccl-tests) suite, which already includes `sendrecv_perf` — a send/recv benchmark that measures the same ring-topology point-to-point pattern with configurable message sizes and iteration counts. The new `nccl-sendrecv` test provides a custom C implementation of similar functionality.

I'd suggest adding a brief note to the README explaining what this custom benchmark provides that `nccl-tests sendrecv_perf` does not — for example, if the goal is a self-contained, minimal example without the full nccl-tests build dependency, or a simpler starting point for users writing their own NCCL code, that's a perfectly valid reason. Making that rationale explicit helps future contributors know which tool to reach for.

micro-benchmarks/nccl-sendrecv/nccl_sendrecv_test.c
```diff
@@ -0,0 +1,148 @@
...
+        NCCLCHECK(ncclGroupEnd());
+        CUDACHECK(cudaEventRecord(ev_end, stream));
+        CUDACHECK(cudaEventSynchronize(ev_end));
+
+        float ms;
+        CUDACHECK(cudaEventElapsedTime(&ms, ev_start, ev_end));
+        total_ms += (double)ms;
+    }
+
+    /* Report from rank 0 */
```
KeitaW, line 125: ### Timing reflects only rank 0's perspective

Each rank measures its own `cudaEventElapsedTime` independently, but only rank 0's accumulated `total_ms` is used for the final report. In a ring topology, different ranks may experience different latencies — especially ranks whose send/recv crosses a node boundary (inter-node EFA) versus ranks communicating intra-node (NVLink). The reported numbers could under- or over-represent the true end-to-end iteration time depending on where rank 0 sits in the topology.

I'd suggest using `MPI_Reduce` with `MPI_MAX` to collect the maximum elapsed time across all ranks before computing the average:

```suggestion
        total_ms += (double)ms;
    }

    /* Report from rank 0 */
    double max_total_ms;
    MPI_Reduce(&total_ms, &max_total_ms, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
    double avg_ms = max_total_ms / iterations;
```

This gives a more accurate picture of the actual completion time per iteration.

micro-benchmarks/nccl-sendrecv/nccl_sendrecv_test.c
```diff
@@ -0,0 +1,148 @@
...
+    double avg_ms = total_ms / iterations;
+    double bw_gbps = (2.0 * msg_size) / (avg_ms * 1e-3) / 1e9;
+    /* Algorithm BW (single direction) */
```
KeitaW: ### "Bus BW (bidir)" label may be misleading

In NCCL's standard test suite, "bus bandwidth" has a [specific definition](https://github.com/NVIDIA/nccl-tests/blob/master/doc/PERFORMANCE.md) that normalizes by the algorithm's communication pattern to approximate the peak per-link bandwidth. Here, the value is simply 2x the algorithmic bandwidth (since each rank both sends and receives simultaneously), which is the aggregate bidirectional throughput — a different (though still useful) metric.

I'd suggest either renaming to something like "Bidir throughput" to avoid confusion with nccl-tests terminology, or adding a brief comment clarifying how it differs from the nccl-tests "busBw" definition. This is especially relevant since users are likely to compare these numbers against `sendrecv_perf` output.

KeitaW (commented):
## Review Batch 2/2 — Code Quality

---

## Things That Look Great

- **Clean, readable C code**: The error-checking macros (`CUDACHECK`, `NCCLCHECK`) are well-written and the overall structure — initialization, warmup, timed loop, cleanup — reads like a textbook example. Easy to understand and modify.
- **Proper NCCL grouped operations**: Using `ncclGroupStart()`/`ncclGroupEnd()` to batch the send and recv is correct and avoids deadlocks.
- **Local rank detection via `MPI_Comm_split_type`**: This is the right way to derive local rank for GPU assignment rather than relying on environment variables that may not exist on all schedulers.
- **Thorough README**: The documentation covers building, running, customization, and expected output with clear tables. The "Configuration" section listing compile-time constants is especially helpful. The note that example values are illustrative is a nice touch.
- **Slurm script follows repo conventions**: Copyright header, `set -ex`, EFA environment variables with exclusion-based `NCCL_SOCKET_IFNAME`, HyperPod auto-resume detection — all align with the patterns in existing `micro-benchmarks/` directories.
- **Well-commented NCCL tuning**: The EFA/NCCL environment variables have clear comments explaining the rationale for each setting.
- **Warm-up phase**: Including 10 warm-up iterations before timing is good benchmarking practice.
- **Makefile supports path overrides**: The `?=` syntax for `CUDA_HOME`, `NCCL_HOME`, and `MPI_HOME` makes the build portable.
- **All previous review items addressed**: Directory placement, license headers, `set -ex`, HyperPod auto-resume, and NCCL env var comments have all been incorporated cleanly.

KeitaW:
@paulogallotti @aravneelaws  Left few additional things I'm wondering (I could be wrong). I appreciate if you could take a look before we merge the PR. Thank you so much!

paulogallotti:
@KeitaW I just updated the PR with the latest updates based on your feedback.

KeitaW (approved):
LGTM. Thank you so much @paulogallotti  and @aravneelaws  for your patience!

PR #1016, opened 2026-03-11
fix: replace undefined $NODES and hardcoded GPU/node counts in Megatron-LM and BioNeMo scripts

nkumaraws:
## Summary

Fixes two bugs in Megatron-LM and BioNeMo training scripts where variables were either undefined or hardcoded, preventing the scripts from working correctly with different cluster configurations.

### Megatron-LM GPT3 (`slurm/gpt3/2.distributed-training.sbatch`)
- **Bug**: `$NODES` was referenced in the TP/PP/GBS scaling logic (lines 18, 22, 25) but never defined. This caused the entire conditional block to silently fail.
- **Fix**: Add `NODES=${SLURM_JOB_NUM_NODES:-2}` before the conditional.
- **Bug**: `TOTAL_GPUS=$((NODES * 8))` and `--nproc_per_node=8` hardcoded 8 GPUs per node.
- **Fix**: Add `GPUS_PER_NODE=8` variable and reference it in both places, making it easy to adjust for instances with fewer GPUs.

### BioNeMo ESM2 (`bionemo_2.5/train-esm.sbatch`)
- **Bug**: `--num-gpus 8` and `--num-nodes 2` were hardcoded, so the script only worked on exactly 2 nodes with 8 GPUs each.
- **Fix**: Replace with `${SLURM_GPUS_PER_NODE:-8}` and `${SLURM_JOB_NUM_NODES:-2}` so the script adapts to the actual Slurm allocation.

## Testing

- Megatron-LM GPT3 live-tested on HyperPod EKS g5.12xlarge (4 GPUs/node, 2 nodes) — 5 training iterations completed successfully
- Both fixes use standard Slurm environment variables with safe defaults matching the original hardcoded values

### DIFF
--- a/3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
+++ b/3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
@@ -18,7 +18,7 @@ GPUS_PER_NODE=${GPUS_PER_NODE:-8} # 4 for g5.12xlarge, 8 for p4de/p5/p5en
 NODES=${SLURM_JOB_NUM_NODES:-2}
 
 # configure TP/PP/GBS based on node count
-if [ $NODES -le 4 ]; then
+if [ "$NODES" -le 4 ]; then
     : "${TENSOR_PARALLEL:=4}"
     : "${PIPELINE_PARALLEL:=2}"
     : "${GLOBAL_BATCH_SIZE:=288}"


3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
```diff
@@ -14,6 +14,9 @@ set -ex;
 ###### User Variables #####
 ###########################
 
+GPUS_PER_NODE=8 # 4 for g5.12xlarge, 8 for p4de/p5/p5en
```
KeitaW: ### Inconsistent parameterization approach between the two scripts

The two scripts in this PR take different approaches to the same problem:

- **Megatron-LM** uses a hardcoded assignment: `GPUS_PER_NODE=8` — users must manually edit the value.
- **BioNeMo** uses Slurm variable substitution: `${SLURM_GPUS_PER_NODE:-8}` — the script auto-adapts to the actual Slurm allocation.

The BioNeMo approach is more robust because it respects the `#SBATCH --gpus-per-node` directive. If a user adds `#SBATCH --gpus-per-node=4` but forgets to also edit this line, `torchrun` will try to launch 8 processes on a node with 4 GPUs allocated.

I'd suggest making this consistent with BioNeMo and with the `${VAR:-default}` pattern used for `IMAGE`, `DATA_PATH`, and `FSX_MOUNT` later in this same file:

```suggestion
GPUS_PER_NODE=${GPUS_PER_NODE:-8} # 4 for g5.12xlarge, 8 for p4de/p5/p5en
```

This preserves the manual-edit workflow (users can `export GPUS_PER_NODE=4` or edit the line) while also allowing environment variable override.

3.test_cases/megatron/megatron-lm/slurm/gpt3/2.distributed-training.sbatch
```diff
@@ -14,6 +14,9 @@ set -ex;
...
+NODES=${SLURM_JOB_NUM_NODES:-2}
+
 # configure TP/PP/GBS based on node count
 if [ $NODES -le 4 ]; then
```
KeitaW: ### Unquoted variable in conditional test (pre-existing)

Now that `NODES` is properly defined, this works correctly, but `[ $NODES -le 4 ]` with an unquoted variable is fragile — if `SLURM_JOB_NUM_NODES` were ever unset *and* the default were removed, this would produce a syntax error. Since you're already touching the surrounding code, this is a one-character defensive fix:

```suggestion
if [ "$NODES" -le 4 ]; then
```

This is a pre-existing issue, not introduced by this PR — mentioning it only as a drive-by improvement.

PR #1024, opened 2026-03-17
fix: overhaul CI workflows for FSDP regression tests

paragao:
## Summary

Restructure and harden the GitHub Actions CI workflows that run FSDP distributed training regression tests on a remote Slurm cluster via SSH.

### Workflow structural fixes
- Fix YAML syntax errors in all 4 Slurm workflows (heredoc `EOF` terminators at column 0 inside YAML block scalars)
- Remove broken EKS regression workflow (`fsdp-eks-regression.yml`)
- Replace heavyweight 814-line PR review workflow with lightweight path-aware linter (`pr-lint.yml`)
- Update `actions/stale` from v5 to v9
- Resolve all actionlint and shellcheck findings

### Slurm job monitoring improvements
- Replace broken `squeue`-based job status detection with `sacct` exit code checking (`squeue` returning empty was incorrectly treated as COMPLETED)
- Add SSH retry wrapper (`ssh_cmd`) with exponential backoff
- Add SSH keepalive settings (`ServerAliveInterval`/`ServerAliveCountMax`)
- Add dedicated enroot cleanup job to avoid race conditions
- Add inline error log dump (last 200 lines) before job failure exit

### Runtime fixes
- Skip `sudo apt install` in `create_venv.sh` when sudo is unavailable
- Fix shell variable escaping for sbatch filename in heredoc
- Insert venv activation after last `#SBATCH` directive (not line 2)
- Add `/opt/slurm/bin` to `PATH` for non-login SSH sessions
- Fix `LD_PRELOAD` path to system NCCL (`/lib/x86_64-linux-gnu/libnccl.so`)
- Pass `HF_TOKEN` to Slurm jobs via sed injection

### HuggingFace data loading (fixes HTTP 429 rate limiting)
- Pre-download C4 dataset shards to local FSx storage on the cluster
- Generate a JSON manifest of local file paths per split
- Modify `train_utils.py` to use `load_dataset("json", data_files=...)` when `HF_DATA_FILES_MANIFEST` env var points to a valid manifest
- Set `HF_HOME` and `HF_DATASETS_CACHE` for shared FSx caching
- Set `HF_HUB_OFFLINE=1` in sbatch files to block runtime API calls
- Pre-cache tokenizers in the workflow pre-download step

### Test matrix
- Align both venv and container workflows: `cluster: [p5]`, `model_config: [llama3_1_8b, llama3_1_70b]`
- Set `NCCL_DEBUG=WARN` (reduced from `INFO`) in all sbatch files

## Files changed (19)

| Category | Files |
|---|---|
| **Workflows** | `fsdp-regression-test-venv.yml`, `fsdp-regression-test-container.yml`, `megatron-ci-slurm.yaml`, `closing-soon.yml`, `pr-lint.yml` (new) |
| **Workflows removed** | `fsdp-eks-regression.yml`, `pr-review-and-slurm-test.yml` |
| **Sbatch files** (9) | All `3.test_cases/pytorch/FSDP/slurm/*-training.sbatch` — LD_PRELOAD, HF caching, NCCL_DEBUG |
| **Scripts** | `create_venv.sh` — sudo fallback |
| **Training code** | `train_utils.py` — manifest-based local data loading |
| **Other** | `.gitignore` — added `log.failed` |

## Testing

Validated through 15+ CI workflow runs on the p5 cluster, iteratively fixing issues from OIDC configuration through data loading.


.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,101 @@
...
+              ERRORS=$((ERRORS + 1))
+            fi
+          done
+
+          if [ "$ERRORS" -gt 0 ]; then
+            echo "Found syntax errors in $ERRORS shell script(s)"
+            exit 1
+          fi
+
+          echo "All shell scripts passed syntax check"
```
KeitaW, line 101: ### Shell syntax check never fails due to subshell variable scope

The `ERRORS` counter is incremented inside a `while` loop that reads from a pipe (`echo | while`). In bash, the right side of a pipe runs in a subshell, so `ERRORS` is always 0 in the parent shell after the loop exits. The `if [ "$ERRORS" -gt 0 ]` check will never trigger, even when `bash -n` reports syntax errors.

I'd suggest using a here-string to keep the loop in the parent shell:

```suggestion
          ERRORS=0

          while IFS= read -r f; do
            [ -z "$f" ] && continue
            [ -f "$f" ] || continue
            echo "  Checking: $f"
            if ! bash -n "$f" 2>&1; then
              ERRORS=$((ERRORS + 1))
            fi
          done <<< "${{ steps.changed.outputs.sh_files }}"

          if [ "$ERRORS" -gt 0 ]; then
            echo "Found syntax errors in $ERRORS shell script(s)"
            exit 1
          fi

          echo "All shell scripts passed syntax check"
```

.github/workflows/pr-lint.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: Setup Python
+        if: steps.changed.outputs.py_count != '0'
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Lint Python Files
+        if: steps.changed.outputs.py_count != '0'
+        run: |
+          pip install flake8
```
KeitaW, line 63: ### Unpinned `flake8` version

`pip install flake8` installs whatever version is current. A future major release could change rules or output format and break the workflow. I'd suggest pinning to a specific version.

```suggestion
          pip install flake8==7.1.1
```

3.test_cases/pytorch/FSDP/src/model_utils/train_utils.py
```diff
@@ -39,53 +42,54 @@ def get_date_of_run():
...
-        print(
-                f"Train Epoch: \t{epoch}, Loss: \t{train_accuracy:.4f}"
-            )
+        print(f"Train Epoch: \t{epoch}, Loss: \t{train_accuracy:.4f}")
     return train_accuracy
 
 
 def validation(model, rank, world_size, val_loader):
     model.eval()
     correct = 0
```
KeitaW, line 91: ### Flake8 F841: unused variable `correct`

`correct = 0` is assigned but never used in the `validation()` function. This is one of the 8 flake8 errors failing the lint check.

```suggestion
```

3.test_cases/pytorch/FSDP/src/model_utils/train_utils.py
```diff
@@ -307,16 +326,15 @@ def apply_activation_checkpoint(args, model=None):
     )
 
     transformer_layer = get_transformer_layer(args.model_type)
-    check_fn_gpt = lambda submodule: isinstance(
-        submodule, transformer_layer
-    )
+    check_fn_gpt = lambda submodule: isinstance(submodule, transformer_layer)
```
KeitaW, line 329: ### Flake8 E731: lambda assigned to variable

Flake8 flags `check_fn_gpt = lambda ...` — the convention is to use `def` instead of assigning a lambda. This is one of the 8 flake8 errors failing the lint check.

```suggestion
    def check_fn_gpt(submodule): return isinstance(submodule, transformer_layer)
```

PR #1025, opened 2026-03-17
Add NeMo RL GRPO training on P5en with EFA RDMA

dmvevents:
## Summary

- Adds a new test case for multi-node GRPO (Group Relative Policy Optimization) training using **NVIDIA NeMo RL** on Amazon EKS with **P5en.48xlarge** instances (H200 GPUs + EFA networking)
- Trains **Nemotron-Mini-4B-Instruct** on synthetic math problems using attention-only LoRA, achieving **+7 percentage point accuracy improvement** (28% to 35%) in 50 training steps
- Includes production-ready Dockerfile, Kubernetes RayJob manifest, training entrypoint, evaluation script, and full SBOM

## Architecture

- **Hardware**: 2x P5en.48xlarge (16x H200 141GB, 32x EFA 200Gbps)
- **Orchestration**: Ray 2.49.2 RayJob with `backoffLimit` for fault tolerance
- **Training**: NeMo RL GRPO with DTensor policy optimization + vLLM generation rollouts
- **Networking**: NCCL 2.27.5 + aws-ofi-nccl 1.18.0 + Libfabric 2.4 over EFA RDMA
- **Storage**: FSx for Lustre (model weights, checkpoints, dataset)

## Artifacts

| File | Purpose |
|------|---------|
| `Dockerfile` | Multi-stage build: NeMo RL + EFA stack (libfabric 2.4, aws-ofi-nccl 1.18, GDRCopy) |
| `rayjob-grpo.yaml` | K8s RayJob manifest for 2-node P5en training with fault recovery |
| `rayjob_entrypoint.sh` | Training entrypoint with NCCL H200/NVSwitch workarounds |
| `eval_nemotron_goldilocks.py` | Before/after accuracy evaluation on 200 held-out problems |
| `sbom/` | Software Bill of Materials (pip packages, dpkg, platform summary) |

## Related upstream work

- [NVIDIA/nccl#2024](https://github.com/NVIDIA/nccl/issues/2024) — Ring Broadcast hang on H200/NVSwitch (filed during this work)
- [ray-project/ray#61073](https://github.com/ray-project/ray/issues/61073) — NCCL performance regression via Ray on H200
- [NVIDIA/nvidia-resiliency-ext#278](https://github.com/NVIDIA/nvidia-resiliency-ext/pull/278) — NVRx fault tolerance integration
- [ai-dynamo/dynamo#7390](https://github.com/ai-dynamo/dynamo/pull/7390) — Local model path fix for Dynamo inference

## Test plan

- [x] Verified training completes 50 GRPO steps on 2x P5en.48xlarge (CGK ap-southeast-3)
- [x] Verified evaluation script reports +7pp accuracy improvement (28% to 35%)
- [x] Verified fault injection and checkpoint recovery via `backoffLimit: 2`
- [x] Verified Dockerfile builds and produces working image
- [x] Verified SBOM matches running container inventory
- [ ] Reviewer: confirm README follows repo conventions
- [ ] Reviewer: confirm namespace/PVC placeholders are clear for adaptation


3.test_cases/pytorch/nemo-rl/grpo/Dockerfile
```diff
@@ -0,0 +1,326 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        autoconf automake libtool libhwloc-dev \
+        environment-modules tcl \
+        udev dmidecode \
+        libevent-core-2.1-7t64 libevent-pthreads-2.1-7t64 libevent-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# GDRCopy — GPU Direct RDMA copy library (needed for EFA GDRDMA)
+RUN cd /tmp && \
+    git clone --depth 1 https://github.com/NVIDIA/gdrcopy.git && \
```
KeitaW: ### Unpinned dependency: GDRCopy

This clones `HEAD` of GDRCopy with no tag or commit pin. Upstream changes could silently break the build. I'd suggest pinning to the version that was tested:

```suggestion
    git clone --depth 1 --branch v2.4.4 https://github.com/NVIDIA/gdrcopy.git && \
```

Reference: [repo convention on fixed version tags](https://github.com/awslabs/awsome-distributed-training/blob/main/CONTRIBUTING.md)

3.test_cases/pytorch/nemo-rl/grpo/Dockerfile
```diff
@@ -0,0 +1,326 @@
...
+# EFA installer — mode selection
+# "full":    Non-minimal install gives libfabric 2.4 + ofi-nccl 1.18 from installer
+# "minimal": Minimal install (libfabric only) + manual ofi-nccl build from source
+RUN <<"EOF" bash -exu
+if [ "${EFA_INSTALL_MODE}" = "full" ]; then
+    # Full mode: purge old libfabric, install EFA with bundled ofi-nccl
+    dpkg --purge --force-depends libfabric1-aws libfabric-aws-bin libfabric-aws-dev 2>/dev/null || true
+    rm -rf /opt/amazon/efa
+    cd /tmp
+    curl -sL https://efa-installer.amazonaws.com/aws-efa-installer-latest.tar.gz | tar xz
```
KeitaW: ### Unpinned dependency: EFA installer

`aws-efa-installer-latest.tar.gz` always pulls the latest version. The repo convention requires pinned versions for reproducible builds. I'd suggest pinning to the version that was tested, e.g.:

```suggestion
    curl -sL https://efa-installer.amazonaws.com/aws-efa-installer-1.38.0.tar.gz | tar xz
```

(Adjust version number to match what you tested with.) The same issue appears on line 272 for the minimal mode path.

KeitaW (commented):

Thanks for this contribution, Anton! The Dockerfile and README are excellent quality. I have several structural and documentation findings below.

### Test case should live under `3.test_cases/megatron/`, not `pytorch/`

NeMo RL is built on top of NVIDIA's Megatron-based stack — the training loop uses Megatron-Core for model parallelism and DTensor for distributed policy optimization. The `pytorch/` top-level directory is reserved for frameworks that use PyTorch directly (FSDP, DDP, TorchTitan, etc.), while Megatron-derived frameworks belong under `3.test_cases/megatron/`. Other Megatron-ecosystem test cases already live there.

I'd suggest moving the entire test case to `3.test_cases/megatron/nemo-rl/grpo/`.

### Missing license header on all files

The repo convention requires a copyright header on all contributed files:

```
```

I'd suggest adding this to the top of each file (after shebang lines where applicable).

### Kubernetes manifest not in a platform-specific subdirectory

The repo convention places Kubernetes manifests in a `kubernetes/` subdirectory within each test case. Currently `rayjob-grpo.yaml` is at the root level. I'd suggest moving it to `kubernetes/rayjob-grpo.yaml` to follow the pattern in other test cases like `FSDP/kubernetes/`.

KeitaW (commented):
## Review Batch 2/3 — Deployment Pipeline + Training Code + Infrastructure

3.test_cases/pytorch/nemo-rl/grpo/rayjob-grpo.yaml
```diff
@@ -0,0 +1,213 @@
...
+# ============================================================
+apiVersion: ray.io/v1
+kind: RayJob
+metadata:
+  name: grpo-nvrx-rayjob
+  namespace: antonai
+  labels:
+    app: grpo-nvrx-rayjob
+spec:
+  entrypoint: bash /shared/nvrx-demo/scripts/rayjob_countdown_entrypoint.sh
```
KeitaW: ### Hardcoded namespace and personal paths in RayJob manifest

The manifest hardcodes `namespace: antonai` and the entrypoint references `/shared/nvrx-demo/scripts/rayjob_countdown_entrypoint.sh` — a path on your personal shared filesystem. Additionally, the entrypoint script included in this PR (`rayjob_entrypoint.sh`) is not what the manifest points to.

I'd suggest:
1. Change the namespace to a placeholder like `<your-namespace>` or a generic name like `nemo-rl`
2. Update the entrypoint to reference the included `rayjob_entrypoint.sh` (e.g., bake it into the container image or mount it)
3. Add a comment noting which values users need to customize

The same `/shared/nvrx-demo/` prefix appears in 10+ places across the entrypoint script and manifest env vars. Centralizing this as a documented variable (e.g., `$SHARED_DIR`) would make adaptation clearer.

3.test_cases/megatron/nemo-rl/grpo/eval_nemotron_goldilocks.py
```diff
@@ -0,0 +1,265 @@
...
+            with open(cfg_file, "w") as f:
+                json.dump(cfg, f, indent=2)
+            print("  Fixed adapter prefix")
+
+    try:
+        from peft import PeftModel
+    except ImportError:
+        import subprocess
+        subprocess.check_call([sys.executable, "-m", "pip", "install", "peft", "-q"])
+        from peft import PeftModel
```
KeitaW: ### Runtime `pip install peft` fallback

This `try/except` runs `pip install peft -q` at runtime if the package isn't installed. Runtime pip installs are non-reproducible (unpinned version, network dependency) and can fail in air-gapped environments. I'd suggest either:
- Adding `peft` to the Dockerfile's Python dependencies, or
- Documenting `peft` as a prerequisite in the README and removing the runtime fallback

3.test_cases/pytorch/nemo-rl/grpo/README.md
```diff
@@ -0,0 +1,192 @@
...
+| NVRx straggler detection deadlocks with DTensor | All | Disabled (process_group fix pending) | [NVIDIA/nvidia-resiliency-ext#277](https://github.com/NVIDIA/nvidia-resiliency-ext/issues/277) |
+
+## SBOM
+
+Software Bill of Materials is available in the `sbom/` directory:
+- `sbom.txt` — Machine-readable package inventory
+- `pip-freeze.txt` — Python packages (235 packages)
+- `dpkg.txt` — System packages (356 packages)
+
+Key versions: PyTorch 2.9.0+cu129, CUDA 12.9, NCCL 2.27.5, NeMo RL 0.5.0rc0, Ray 2.49.2, vLLM (in worker venvs), NVRx 0.4.1
```
KeitaW: ### NCCL version mismatch between README and SBOM

This line states "NCCL 2.27.5" but the SBOM (`sbom/dpkg.txt`) shows `libnccl2,2.26.5-1+cuda12.9`. Could you confirm which version was actually used in testing and update accordingly?

3.test_cases/pytorch/nemo-rl/grpo/rayjob-grpo.yaml
```diff
@@ -0,0 +1,213 @@
+# ============================================================
+# RayJob: Qwen2.5-1.5B GRPO + NVRx on P5en (H200)
```
KeitaW: ### Model name mismatch between README, manifest, and entrypoint

There's a three-way inconsistency in which model is being trained:

1. **This comment** (line 2): says "Qwen2.5-1.5B GRPO"
2. **GRPO_MODEL env var** (line 115): sets `Qwen/Qwen2.5-0.5B-Instruct` (0.5B, not 1.5B)
3. **README title and results**: describes training "Nemotron-Mini-4B-Instruct"
4. **Entrypoint script**: references Nemotron and a phase 1 merge workflow

It looks like the manifest was adapted from a Qwen experiment but the README and entrypoint describe the Nemotron workflow. I'd suggest aligning these — a user following the README would encounter unexpected behavior from the manifest as-is.

3.test_cases/pytorch/nemo-rl/grpo/rayjob_entrypoint.sh
```diff
@@ -0,0 +1,89 @@
...
+export TORCH_HOME="/shared/nvrx-demo/torch_cache"
+export TRITON_CACHE_DIR="/shared/nvrx-demo/triton_cache"
+mkdir -p "$HF_HOME" "$HF_DATASETS_CACHE" "$TORCH_HOME" "$TRITON_CACHE_DIR"
+
+# Step 1: Merge phase 1 LoRA if not already done
+MERGED="/shared/nvrx-demo/nemotron-phase1-merged"
+if [ ! -f "$MERGED/model.safetensors" ]; then
+  echo "Merging phase 1 LoRA checkpoint..."
+  pip install peft -q 2>&1 | tail -1
+  python3 /shared/nvrx-demo/scripts/merge_lora_checkpoint.py \
```
KeitaW: ### Entrypoint references non-existent files

This line references `/shared/nvrx-demo/scripts/merge_lora_checkpoint.py`, which isn't included in the PR. Other missing references:
- `/shared/nvrx-demo/goldilocks/train.jsonl` (dataset, line 80)
- `/shared/nvrx-demo/miracle-checkpoints` (phase 1 checkpoints, line 34)

These are on your FSx volume — other users won't have them. I'd suggest either:
1. Including `merge_lora_checkpoint.py` in the PR (or documenting how to create it)
2. Adding a "Data Preparation" section to the README explaining how to set up these paths
3. Using a Hugging Face dataset directly, or documenting how to generate the JSONL

3.test_cases/pytorch/nemo-rl/grpo/eval_nemotron_goldilocks.py
```diff
@@ -0,0 +1,265 @@
...
+import os
+import re
+import sys
+import time
+from pathlib import Path
+
+os.environ.setdefault("HF_HOME", "/shared/nvrx-demo/hf_cache")
+
+MODEL = "nvidia/Nemotron-Mini-4B-Instruct"
+DATASET = "/shared/nvrx-demo/goldilocks/train.jsonl"
```
KeitaW: ### Eval script hardcodes local paths

The `DATASET` path points to `/shared/nvrx-demo/goldilocks/train.jsonl` on your shared filesystem. Users following the README won't have this file. I'd suggest making `DATASET` a command-line argument (like `--model` already is) with a helpful error message if the file doesn't exist.

```suggestion
os.environ.setdefault("HF_HOME", "/shared/hf_cache")

MODEL = "nvidia/Nemotron-Mini-4B-Instruct"
DATASET = os.environ.get("EVAL_DATASET", "/shared/goldilocks/train.jsonl")
```

KeitaW (commented):

---

## Things That Look Great

- **Multi-stage Dockerfile with excellent layer caching**: The strategy of copying only `pyproject.toml` + `uv.lock` first to cache the expensive dependency install stage, then copying full source later, is a production-quality pattern that saves significant rebuild time.
- **Dual EFA install mode**: Supporting both `full` and `minimal` EFA installation modes makes the image adaptable across instance types (P5en RDMA vs. socket-only).
- **NRL_GIT_REF pinned to commit hash**: Using a specific commit (`c40dba37789c`) rather than a branch name ensures build reproducibility for the NeMo RL source.
- **Comprehensive SBOM**: Including `dpkg.txt`, `pip-freeze.txt`, and `sbom.txt` gives full supply chain visibility — this is above and beyond what most test cases provide.
- **Excellent README**: The verified instance types table, architecture diagram, instance-specific configuration snippets, and known issues table with upstream issue links make this one of the most thorough READMEs in the repo.
- **NVRx fault tolerance documentation**: The fault injection test procedure and checkpoint resume behavior are well-documented with concrete steps, making the fault tolerance claims verifiable.
- **YAML anchors for DRY configuration**: Using `&common-env`, `&common-mounts`, and `&common-volumes` with YAML aliases avoids config drift between head and worker specs.

KeitaW (changes requested):
Nice addition few comments.

KeitaW:
### Question: Why include the SBOM files in the repo?

I noticed the `sbom/` directory adds ~1,200 lines of package inventory (`dpkg.txt`, `pip-freeze.txt`, `sbom.txt`). These are a snapshot of one specific container build and will drift from the actual image contents as the Dockerfile evolves — the SBOM would need to be regenerated and recommitted on every dependency change to stay accurate.

The Dockerfile already has `generate_sbom.sh` baked in, so the SBOM is generated at build time inside the image itself. Could you help me understand the motivation for also checking it into the repo? If the goal is supply chain visibility, the in-image SBOM (already produced by the build) may be sufficient without committing a static copy that can go stale.

3.test_cases/pytorch/nemo-rl/grpo/Dockerfile
paragao: ### GDRCopy: pin to v2.5.1

This clones HEAD of GDRCopy with no tag or commit pin. Should be pinned to the tested version:

```suggestion
    git clone --depth 1 --branch v2.5.1 https://github.com/NVIDIA/gdrcopy.git && \
```

3.test_cases/pytorch/nemo-rl/grpo/Dockerfile
paragao: ### EFA installer: pin version and clean up old path

Two issues here:

1. `aws-efa-installer-latest.tar.gz` is not reproducible. Pin to a specific version.
2. After the EFA install, the old path `/opt/amazon/aws-ofi-nccl/` from the base image still exists alongside the new `/opt/amazon/ofi-nccl/`. Add a cleanup step:

```bash
rm -rf /opt/amazon/aws-ofi-nccl
```

This prevents the old `libnccl-net-aws-ofi.so` from being loaded accidentally.

3.test_cases/megatron/nemo-rl/grpo/Dockerfile
```diff
@@ -0,0 +1,326 @@
...
+fi
+ldconfig
+EOF
+
+ENV LD_LIBRARY_PATH="/opt/amazon/ofi-nccl/lib:/opt/amazon/efa/lib${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
+
+# EFA / NCCL defaults
+ENV FI_PROVIDER=efa
+ENV FI_EFA_USE_DEVICE_RDMA=1
+ENV NCCL_NET_PLUGIN=ofi
```
paragao, line 368: ### Container bakes wrong `NCCL_NET_PLUGIN` default

The container has `NCCL_NET_PLUGIN=aws-ofi` baked in (from the base image). With EFA installer 1.47.0, the correct value is `ofi`. The PR YAML correctly overrides this, but the container default should also be fixed.

Also, `AWS_OFI_NCCL_VERSION=1.14.0` is stale — the actual installed version is 1.18.0.

paragao (commented):
## Hands-On Testing Report — P5en H200 Cluster (Slurm + Pyxis)

I tested this PR on a 16-node P5en.48xlarge cluster (8x NVIDIA H200 143GB per node, EFA, FSx Lustre) running Slurm with Pyxis/Enroot.

### What works

- **Container image pulls and runs correctly** on H200 hardware via `srun --container-image='docker://public.ecr.aws#v9l4g5s4/nemo-rl-workshop:p5en'`
- **Verified versions inside container**: PyTorch 2.9.0+cu129, CUDA 12.9, Ray 2.49.2, NeMo RL 0.5.0rc0, NCCL 2.27.5 (pip/runtime)
- **EFA stack is functional**: libfabric 2.4.0amzn1.0, EFA installer 1.47.0, `libnccl-ofi` 1.18.0, GDRCopy present
- **1-node GRPO training (8x H200) completes successfully** — I ran 3 steps of the default NeMo RL GRPO example (Qwen2.5-1.5B + OpenMathInstruct-2) and all steps completed without errors. Step 1: 33s (cold), Steps 2-3: ~10.3s each.

### What doesn't work

- **PR entrypoint fails immediately** — all 4 data dependencies are missing (`goldilocks/train.jsonl`, `merge_lora_checkpoint.py`, `miracle-checkpoints/`, `nemotron-phase1-merged/`). These live on the author's FSx volume and are not included in the PR or documented for reproduction.
- **`/etc/hosts` write in entrypoint fails** — `echo "$MY_IP $(hostname)" >> /etc/hosts` returns `Permission denied` in containers that don't run as privileged/writable. The K8s YAML sets `privileged: true` but this assumption should be documented.

### Container EFA/NCCL Audit

I inspected the container's NCCL and EFA configuration in detail and found several issues:

#### 1. Old OFI-NCCL path still exists

Both paths are present in the container:

| Path | Status |
|------|--------|
| `/opt/amazon/aws-ofi-nccl/` | **OLD** — should not exist, leftover from base image |
| `/opt/amazon/ofi-nccl/` | **NEW** — correct path from EFA installer 1.47.0 |

The old path contains `libnccl-net-aws-ofi.so` and `libnccl-tuner-aws-ofi.so` which could be loaded accidentally. The Dockerfile should clean up `/opt/amazon/aws-ofi-nccl/` after the EFA install.

#### 2. `NCCL_NET_PLUGIN` container default is wrong

The container bakes `NCCL_NET_PLUGIN=aws-ofi` (old naming), but the correct value for EFA installer 1.47.0 is `ofi`. The PR YAML correctly overrides this to `ofi`, but anyone running the container without the YAML override gets the wrong plugin.

#### 3. `NCCL_TUNER_PLUGIN` not set

The YAML does not set `NCCL_TUNER_PLUGIN`. The EFA tuner plugin exists at `/opt/amazon/ofi-nccl/lib/libnccl-tuner-ofi.so` and should be explicitly enabled for optimal performance.

#### 4. `AWS_OFI_NCCL_VERSION` env var is stale

The container has `AWS_OFI_NCCL_VERSION=1.14.0` baked in (from the base image), but the actually installed version is **1.18.0** (confirmed via `dpkg -s libnccl-ofi`).

#### 5. Dual NCCL versions

| Source | Version |
|--------|----------|
| pip `nvidia-nccl-cu12` (what PyTorch loads at runtime) | **2.27.5** |
| dpkg `libnccl2` (system package) | **2.26.5** |
| `NCCL_VERSION` env var | **2.26.5** |
| `nccl.h` header | **2.26.5** |

PyTorch uses the pip wheel version (2.27.5) at runtime across all three Ray venvs (main, DTensor, vLLM). The system package 2.26.5 is dead weight. The README correctly states 2.27.5 for the runtime version, but this should be clarified since the SBOM and env var show 2.26.5.

#### 6. GDRCopy should be pinned to v2.5.1

The Dockerfile clones GDRCopy HEAD with no tag (`git clone --depth 1`). This should be pinned to `--branch v2.5.1`.

3.test_cases/megatron/nemo-rl/grpo/kubernetes/rayjob-grpo.yaml
```diff
@@ -0,0 +1,213 @@
...
+                  value: "0"
+                - name: FI_MR_CACHE_MONITOR
+                  value: "disabled"
+                - name: FI_HMEM_CUDA_USE_GDRCOPY
+                  value: "1"
+                - name: NCCL_NET_GDR_LEVEL
+                  value: "PHB"
+                - name: NCCL_CUMEM_ENABLE
+                  value: "0"
+                - name: NCCL_NVLS_ENABLE
```
paragao, line 108: ### Missing `NCCL_TUNER_PLUGIN`

The EFA tuner plugin is installed at `/opt/amazon/ofi-nccl/lib/libnccl-tuner-ofi.so` but `NCCL_TUNER_PLUGIN` is not set in the env vars. Add:

```yaml
- name: NCCL_TUNER_PLUGIN
  value: "ofi"
```

3.test_cases/pytorch/nemo-rl/grpo/rayjob_entrypoint.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+set -e
+echo "=============================================="
+echo "PHASE 2: Accuracy Sprint (merged base + protective LoRA)"
+echo "Starting from step_60 merged model (has format knowledge)"
+echo "=============================================="
+MY_IP=$(hostname -I | awk '{print $1}')
+echo "$MY_IP $(hostname)" >> /etc/hosts
```
paragao: ### Replace `/etc/hosts` write with Ray IP configuration

Tested on H200 cluster:
```
/etc/hosts: Permission denied
```

This line writes to `/etc/hosts` to ensure Ray resolves the container's hostname to its real network IP instead of loopback. However, this requires `privileged: true` on the container and fails in non-privileged environments.

Instead, use Ray's built-in IP configuration:
- Set `RAY_ADDRESS` environment variable, or
- Pass `--node-ip-address=$MY_IP` to `ray start`

This avoids the need for privileged containers just for hostname resolution, and is more portable across container runtimes (K8s, Enroot/Pyxis, etc.). With `hostNetwork: true` already set in the YAML, the container shares the host's `/etc/hosts` anyway, making this write redundant in the K8s case.

3.test_cases/pytorch/nemo-rl/grpo/rayjob_entrypoint.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+set -e
```
paragao: ### Use `set -euo pipefail`

Currently only `set -e`. Missing `-u` (unset variable detection) and `-o pipefail` (piped command failure detection). Recommend:

```suggestion
set -euo pipefail
```

3.test_cases/pytorch/nemo-rl/grpo/README.md
paragao: ### NCCL version clarification

Adding context to @KeitaW's comment: I verified inside the container that there are **two NCCL versions**:

- **pip `nvidia-nccl-cu12`: 2.27.5** — this is what PyTorch loads at runtime (confirmed via `torch.cuda.nccl.version()` on H200)
- **dpkg `libnccl2`: 2.26.5** — system package, not used by PyTorch

The README's "NCCL 2.27.5" is correct for the runtime version. I'd suggest adding a note: "NCCL 2.27.5 (PyTorch wheel; system package is 2.26.5)" to avoid confusion with the SBOM.

3.test_cases/pytorch/nemo-rl/grpo/Dockerfile
```diff
@@ -0,0 +1,326 @@
...
+#   BASE_IMAGE              CUDA base image (default: cuda-dl-base 25.05)
+#   NRL_GIT_REF             NeMo RL commit/branch to build from
+#   TORCH_CUDA_ARCH_LIST    Target GPU SM versions (default: "9.0")
+#   UV_VERSION              uv package manager version
+#   PYTHON_VERSION          Python version (default: 3.12)
+#   OFI_NCCL_VERSION        aws-ofi-nccl version (default: v1.18.0)
+#   MAX_JOBS                Parallel compilation limit (prevent OOM)
+#   EFA_INSTALL_MODE        "full" or "minimal" (default: full for P5en RDMA)
+
+# syntax=docker/dockerfile:1.10
```
paragao: ### `syntax` directive must be line 1

`# syntax=docker/dockerfile:1.10` is on line 43 (after the header comments). For BuildKit to recognize it, it must be the very first line of the Dockerfile. Move it above all comments.

dmvevents:
Thanks @KeitaW and @paragao for the thorough reviews! All feedback has been addressed in the latest push:

**Structural:**
- ✅ Moved to `3.test_cases/megatron/nemo-rl/grpo/` (NeMo RL uses Megatron Core)
- ✅ Removed `sbom/` directory — added extraction instructions in README instead

**Dockerfile:**
- ✅ `# syntax=docker/dockerfile:1.10` moved to line 1
- ✅ GDRCopy pinned to `v2.5.1`
- ✅ EFA installer pinned to `1.47.0`
- ✅ Added `NCCL_TUNER_PLUGIN=ofi`

**RayJob YAML:**
- ✅ Namespace changed to `default` with comment to customize
- ✅ Model name consistently `nvidia/Nemotron-Mini-4B-Instruct`
- ✅ Removed hardcoded personal paths
- ✅ Added `NCCL_TUNER_PLUGIN` env var

**Entrypoint:**
- ✅ `set -euo pipefail` with `${VAR:-default}` pattern throughout
- ✅ Removed `/etc/hosts` write (Ray GCS handles IP discovery)
- ✅ Removed dependency on `merge_lora_checkpoint.py` — falls back to HuggingFace model

**Eval script:**
- ✅ Paths configurable via `--dataset`, `--checkpoint-dir` CLI args and `SHARED_DIR` env var

**README:**
- ✅ NCCL version clarified: pip 2.27.5 (PyTorch runtime) vs system 2.26.5 (dpkg base image)
- ✅ Model name consistent
- ✅ SBOM extraction instructions added

@paragao — thanks for testing on the P5en Slurm cluster! This PR targets EKS (RayJob), but the container image works on Slurm+Pyxis as you confirmed. Happy to add a Slurm section in a follow-up PR if that would be useful.

3.test_cases/megatron/nemo-rl/grpo/kubernetes/rayjob-grpo.yaml
```diff
@@ -0,0 +1,213 @@
+# ============================================================
```
KeitaW, line 4: ### RayJob manifest not in a `kubernetes/` subdirectory

The repo convention places Kubernetes manifests in a `kubernetes/` subdirectory within each test case. I'd suggest moving this to `3.test_cases/megatron/nemo-rl/grpo/kubernetes/rayjob-grpo.yaml` to follow the pattern in other test cases like `FSDP/kubernetes/`.

KeitaW (commented):
## Re-review 1/2 — Structure & Repository Hygiene

Great progress since the first review! Most findings addressed. A couple of structural items remain.

### Missing license headers on all files

None of the 5 files include the repo-required copyright header:

```
```

For the Dockerfile, place it after the `# syntax=` directive. For Python, after the shebang. For YAML, at the top. For the README, use an HTML comment.

3.test_cases/megatron/nemo-rl/grpo/eval_nemotron_goldilocks.py
```diff
@@ -0,0 +1,285 @@
...
```
KeitaW: ### Runtime `pip install peft` fallback still present

This `try/except` runs `pip install peft -q` at runtime if the package isn't installed. Runtime pip installs are non-reproducible (unpinned version, network dependency) and fail in air-gapped environments. I'd suggest either:
- Documenting `peft` as a prerequisite in the README's evaluation section, or
- Adding `peft` to the Dockerfile

KeitaW (commented):
## Re-review 2/2 — Training Code Quality

### Dataset preparation not documented

The entrypoint expects a Goldilocks math dataset at `$DATA_PATH` (default: `${SHARED_DIR}/goldilocks/train.jsonl`), but the README doesn't explain how to obtain or generate this file. A brief "Data Preparation" section would help — even just noting the expected JSONL format (`{"problem": "...", "answer": "..."}`) and where to source the data.

---

## Things That Look Great

- **Excellent response to first review**: Moved to `megatron/`, pinned EFA (1.47.0) and GDRCopy (v2.5.1), fixed model name alignment, parameterized namespace with inline `# <-- change` markers, and explained the dual NCCL version situation clearly.
- **`SHARED_DIR` env var pattern**: The entrypoint and eval script now derive all paths from a single `SHARED_DIR` variable with sensible defaults — easy to adapt to different FSx mounts.
- **Configurable training parameters**: `GRPO_MODEL`, `GRPO_MAX_STEPS`, `GRPO_NUM_NODES`, `CKPT_DIR`, and `DATA_PATH` are all overridable via environment variables.
- **Clear setup instructions in manifest comments**: Numbered setup guide and `# <-- change` markers on every user-configurable value.
- **SBOM moved to build-time**: No stale inventory to maintain.
- **NCCL version explanation**: The README note about pip wheel 2.27.5 vs system 2.26.5 preempts common confusion.
- **Proper `set -euo pipefail`** in the entrypoint script.
- **`NCCL_TUNER_PLUGIN=ofi`**: Good addition for EFA-aware NCCL algorithm selection.

dmvevents:
## Summary of changes (April 16-17, 2026)

All KeitaW round-2 review items are now addressed, plus comprehensive CVE remediation:

### Review items (commits ae1a1b3, d166b18)
- ✅ License headers (MIT-0) on all 7 files
- ✅ Moved `rayjob-grpo.yaml` → `kubernetes/rayjob-grpo.yaml`
- ✅ Dataset preparation documented in README (Data Preparation section)
- ✅ Removed runtime `pip install peft` — baked into Dockerfile
- ✅ Cleaned stale `/opt/amazon/aws-ofi-nccl` path from container
- ✅ `AWS_OFI_NCCL_VERSION` set to 1.18.0 (was stale 1.14.0)

### CVE remediation (commits c8fcfa1, 2425f2a, 23322f9)
Starting from 24 CRITICAL (Inspector April 10):
1. Base image bump `cuda-dl-base:25.05` → `25.06`
2. `apt-get upgrade` in Dockerfile
3. Python CVE pins: h11≥0.16, Pillow≥10.3, certifi≥2024.7.4, urllib3≥2.2.2, requests≥2.32, jinja2≥3.1.5, aiohttp≥3.13.3, tornado≥6.4.2, setuptools≥75.6, cryptography≥46.0.5, zipp≥3.19.1, tqdm≥4.66.4, werkzeug≥3.0.6, mlflow≥3.8.1, cbor2≥5.9.0, pyasn1≥0.6.3, xgrammar≥0.1.32
4. Cross-venv CVE application (main env + 2 Ray worker venvs)
5. wandb 0.25→0.26 + ray 2.49→2.52 upgrades
6. Deleted `nic_sampler` Go binary (CVE-2025-68121)
7. `.trivyignore` for 2 accepted-risk CVEs:
   - `CVE-2023-48022` (Ray dashboard, disputed by design)
   - `CVE-2026-22778` (vllm 0.11.2 → 0.14.1 breaks NeMo RL APIs)

**Result: Trivy reports 0 CRITICAL with .trivyignore. Without it, 1 CRITICAL (vllm, accepted risk).**

Runtime verified: `torch=2.9.0+cu129, ray=2.52.0, mlflow=3.11.1, peft OK, nemo_rl OK`

### Pending
- Full training smoke test on P5en (in progress — deploying now)
- ECR push for distribution review

@KeitaW ready for re-review when you have a chance.

3.test_cases/megatron/nemo-rl/grpo/.trivyignore
```diff
@@ -0,0 +1,8 @@
+# CVE-2023-48022: Ray Dashboard access — disputed by Ray project (by design, not a bug)
+# https://github.com/ray-project/ray/issues/38720
+CVE-2023-48022
+
+# CVE-2026-22778: vllm 0.11.2 — fix requires 0.14.1 which has breaking API changes
+# NeMo RL is tightly coupled to vllm 0.11.x generation APIs. Upgrading breaks training.
+# Accepted risk: vllm is not exposed to untrusted input in this container.
+CVE-2026-22778
```
KeitaW, line 11: ### Missing license header on `.trivyignore`

`.trivyignore` is the only file in the test case still missing the repo's
copyright header — every other file got one in the round-2 fix. Trivy treats
`#`-prefixed lines as comments, so adding the header at the top doesn't
affect parsing.

```suggestion
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT-0

# CVE-2023-48022: Ray Dashboard access — disputed by Ray project (by design, not a bug)
# https://github.com/ray-project/ray/issues/38720
CVE-2023-48022

# CVE-2026-22778: vllm 0.11.2 — fix requires 0.14.1 which has breaking API changes
# NeMo RL is tightly coupled to vllm 0.11.x generation APIs. Upgrading breaks training.
# Accepted risk: vllm is not exposed to untrusted input in this container.
CVE-2026-22778
```

3.test_cases/megatron/nemo-rl/grpo/kubernetes/rayjob-grpo.yaml
```diff
@@ -0,0 +1,216 @@
...
+                # --- Shared storage paths (configurable) ---
+                - name: SHARED_DIR
+                  value: "/shared/nvrx-demo"
+                - name: HF_HOME
+                  value: "/shared/nvrx-demo/hf_cache"
+                - name: HF_DATASETS_CACHE
+                  value: "/shared/nvrx-demo/hf_datasets"
+                - name: TORCH_HOME
+                  value: "/shared/nvrx-demo/torch_cache"
+                - name: TRITON_CACHE_DIR
```
KeitaW, line 128: ### `SHARED_DIR` and downstream paths still default to `/shared/nvrx-demo`

The block comment above this section says `# --- Shared storage paths (configurable) ---`,
but the five env vars then hardcode `/shared/nvrx-demo` — your internal NVRx
demo directory name. A user copy-pasting this manifest gets a path that
almost certainly doesn't exist on their FSx mount, with no `# <-- change`
marker like the image refs elsewhere in this same file have.

I'd suggest one of:

1. **Single-source the path** — set only `SHARED_DIR` and derive the rest.
   Kubernetes expands `$(VAR)` references in env values when the referenced
   var is defined earlier in the same container's `env:` list, so this stays
   DRY without a templating step:
   ```yaml
   - name: SHARED_DIR
     value: "/fsx"  # <-- change to your FSx mount path
   - name: HF_HOME
     value: "$(SHARED_DIR)/hf_cache"
   - name: HF_DATASETS_CACHE
     value: "$(SHARED_DIR)/hf_datasets"
   - name: TORCH_HOME
     value: "$(SHARED_DIR)/torch_cache"
   - name: TRITON_CACHE_DIR
     value: "$(SHARED_DIR)/triton_cache"
   ```
2. **Or leave the literal paths but add `# <-- change`** on each, matching
   the image-ref convention already in this file.

3.test_cases/megatron/nemo-rl/grpo/rayjob_entrypoint.sh
```diff
@@ -0,0 +1,106 @@
...
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+set -euo pipefail
+echo "=============================================="
+echo "NeMo RL GRPO Training Entrypoint"
+echo "=============================================="
+
+# --- Configurable paths (override via environment variables) ---
+SHARED_DIR="${SHARED_DIR:-/shared/nvrx-demo}"
```
KeitaW: ### `SHARED_DIR` default leaks the personal `nvrx-demo` path

Same issue as in the K8s manifest: when a user runs this entrypoint without
setting `SHARED_DIR` (or via `kubectl exec` for debugging), the fallback
path is `/shared/nvrx-demo` rather than something neutral. Could you change
the default to `/fsx` (or whatever the canonical FSx mount path is in your
docs) so this isn't a footgun for adapters?

```suggestion
SHARED_DIR="${SHARED_DIR:-/fsx}"
```

KeitaW (commented):
## Re-review (round 3) Batch 1/2 — Resolution status + Structure findings

Anton — most of my prior findings are now closed. Thanks for the thorough
follow-through, especially on the CVE remediation and SBOM/license artifacts.
Resolution status table below, then two structural items that are still open
or partially resolved.

### Resolution status of prior findings

| # | Finding | Status |
|---|---------|--------|
| 1 | License headers on all files | ✅ Resolved on 7/8; `.trivyignore` still missing one (inline comment below) |
| 2 | Dataset preparation not documented | ✅ Resolved — README §"Data Preparation" added |
| 3 | GDRCopy unpinned | ✅ Resolved — `GDRCOPY_VERSION=v2.5.1` |
| 4 | EFA installer unpinned | ✅ Resolved — `EFA_INSTALLER_VERSION=1.47.0` |
| 5 | Hardcoded `antonai` namespace + personal paths | 🟡 Partial — namespace gone, but `/shared/nvrx-demo` is still the literal default (inline comments below) |
| 6 | Runtime `pip install peft` fallback in eval script | ✅ Resolved — `peft` baked into Dockerfile, no try/except |
| 7 | NCCL version mismatch README vs SBOM | ✅ Resolved |
| 8 | Model name mismatch | ✅ Resolved — all references use `Nemotron-Mini-4B-Instruct` |
| 9 | Entrypoint referenced non-existent `merge_lora_checkpoint.py` | ✅ Resolved |
| 10 | Eval script hardcoded `DATASET` path | ✅ Resolved — uses `SHARED_DIR` env var |

Two remaining structural items inline below. Batch 2 will cover one new
infrastructure finding plus the "Things That Look Great" list.

3.test_cases/megatron/nemo-rl/grpo/Dockerfile
```diff
@@ -0,0 +1,381 @@
...
+
+# vLLM — deep_ep and deep_gemm use Hopper-only TMA instructions
+# Force SM 9.0 regardless of TORCH_CUDA_ARCH_LIST for these
+TORCH_CUDA_ARCH_LIST="9.0" uv sync --locked --extra vllm --no-install-project
+
+# Automodel — flash-attn and transformer-engine compile for target archs
+uv sync --locked --extra automodel --no-install-project
+
+# peft — needed by eval script for LoRA adapter merging
+uv pip install --python ${UV_PROJECT_ENVIRONMENT}/bin/python peft
```
KeitaW: ### `peft` installed without a version pin

This is the package whose runtime install we removed in round 1, and it
landed in the Dockerfile without a version constraint. Every other pip
block in this file uses CVE-floor pins (`"h11>=0.16.0"`, etc.) — `peft` is
the lone unpinned install. The next `docker build` could pick up a
major-version bump that breaks the GRPO LoRA path (the eval script's
`peft.PeftModel.from_pretrained(...)` call is API-sensitive).

Could you pin to whatever version you've validated against?

```suggestion
uv pip install --python ${UV_PROJECT_ENVIRONMENT}/bin/python "peft>=0.13,<1.0"
```

(Adjust the lower bound to the version you've actually built and tested
with — the upper bound at `<1.0` is just a major-version cap.)

KeitaW (commented):
## Re-review (round 3) Batch 2/2 — Infrastructure + Things That Look Great

One inline finding on the Dockerfile, plus a discussion note on CVE pin
strategy and the recap of what's working well.

### Floor-only CVE pins are a deliberate tradeoff — flagging only as discussion

The CVE remediation block in the Dockerfile (around lines 163-194) uses
lower-bound-only pins (`"h11>=0.16.0"`, `"Pillow>=10.3.0"`, etc., no upper
bounds). The repo checklist asks for upper bounds in general, but for CVE
floors specifically the tradeoff flips — capping at `<X+1.0` means a future
patch release that fixes another CVE silently fails to install. So I read
this as intentional and don't think it needs to change.

If you want both, a layered approach works: use `>=` here for "must be at
least the patched version" and rely on `uv.lock` (already COPY'd into the
image) to pin the actual resolved version. That gives both reproducibility
and explicit CVE floors, with the lock file as the source of truth. Not a
blocker — just calling it out in case the right call here is to add a
short comment explaining why these are floor-only.

---

## Things That Look Great

- **Round-2 feedback fully addressed**: dataset preparation section, license
  headers on Python/YAML/Dockerfile/README, eval script using `SHARED_DIR`,
  no more `pip install peft` fallback. Only `.trivyignore` slipped through.
- **`.trivyignore` justifications are exemplary**: each ignored CVE has a
  comment explaining *why* it's safe to ignore (Ray dashboard "disputed by
  upstream", vllm "tight API coupling, no untrusted input"). This is the
  right bar for any future contributor adding a `.trivyignore`.
- **CVE remediation is thorough**: base image bump to
  `nvcr.io/nvidia/cuda-dl-base:25.06-cuda12.9-devel-ubuntu24.04`, Python pins
  for `h11`, `Pillow`, `certifi`, `urllib3`, `requests`, `idna`, `jinja2`,
  `aiohttp`, `tornado`, `setuptools`, `cryptography`, plus apt upgrades and
  cache cleanup at every layer. The Ray-worker venv pins (`mlflow 3.5.1 →
  3.11.1`, `wandb`, `nic_sampler`) caught a class of CVE that's easy to miss
  inside Ray's auto-installed runtime envs.
- **`generate_sbom.sh`**: build-time SBOM beats stale checked-in artifacts.
  `set -euo pipefail`, proper variable quoting, license header all present.
- **`THIRD-PARTY-LICENSES`** follows the repo format and points readers at
  the SBOM extraction method, so the inventory and license disclosures stay
  aligned via a single source.
- **NCCL_SOCKET_IFNAME**: `^lo,docker,veth,eni` (rayjob-grpo.yaml L82, L156)
  uses exclusion patterns and survives EFA interfaces — matches the repo
  convention.
- **`backoffLimit: 2`** (L37) is conservative for GPU workloads, matching
  the repo guidance of ≤3.
- **Entrypoint hardening**: `set -euo pipefail` on line 5, all variables
  quoted, parameterized via env-var fallbacks. Once the default is changed
  off `/shared/nvrx-demo`, this is solid.
- **YAML anchors for DRY config** (`&common-env`, `&common-mounts`,
  `&common-volumes`) are still in place and prevent head-vs-worker drift.
- **Pinned base image and source SHA**: `BASE_IMAGE=nvcr.io/nvidia/cuda-dl-base:25.06-...`
  and `NRL_GIT_REF=c40dba37789c64d03fcce72b152e5657ae6aacd7` give a fully
  reproducible build. Easy to bisect if upstream regresses.

Thanks again — this is in much better shape than it was at round 2.

KeitaW (approved):
Thanks a lot @dmvevents . Approved

PR #1029, opened 2026-03-20
Add DeepSpeed CI regression tests for QLoRA and GPT-103B

paragao:
## Summary

- Add GitHub Actions workflow (`deepspeed-regression-test.yml`) for automated DeepSpeed regression testing on the p5en Slurm cluster
- Update QLoRA sbatch scripts for EFA-enabled p5en instances (NCCL_DEBUG, EFA settings, LD_PRELOAD)

## Workflow Design

**Two independent jobs** running in parallel:

### Job 1: `qlora` (matrix: zero2, zero3)
- QLoRA fine-tuning of **Qwen/Qwen3-8B** on medical-o1-reasoning-SFT dataset
- Builds container from `qlora/Dockerfile`, converts to enroot `.sqsh`
- Pre-caches model weights and dataset on `/fsx/.cache/huggingface`
- Runs with `--max_samples 100` for quick CI validation
- Tests both DeepSpeed ZeRO-2 and ZeRO-3 configurations

### Job 2: `gpt-103b`
- Megatron-DeepSpeed **GPT-103B pretraining** on synthetic BookCorpus data
- Builds container from `0.deepspeed.dockerfile`, converts to enroot `.sqsh`
- Sets up Megatron-DeepSpeed repo + GPT-2 tokenizer + preprocessed data (idempotent)
- CI config: 4 nodes, TP=8, PP=4, ZeRO=0, 10 iterations, fusions enabled

### Sbatch Changes
- `NCCL_DEBUG=INFO` → `NCCL_DEBUG=WARN` (reduce log noise)
- Uncommented EFA settings block (p5en has EFA)
- Added `LD_PRELOAD=/lib/x86_64-linux-gnu/libnccl.so` for system NCCL with EFA support

### Triggers
- **Scheduled:** Every Saturday at 14:00 UTC
- **Push/PR:** Changes to `3.test_cases/pytorch/deepspeed/**`
- **Manual:** `workflow_dispatch`

### Patterns
Follows same SSH/monitor/cleanup patterns as the FSDP regression workflows (PR #1024): keyscan retry, `sacct`-based job status detection, error log dump on failure, artifact upload, cleanup with scancel + rm.

## Testing
- `actionlint` passes (only shellcheck info/style warnings, consistent with existing workflows)
- Workflow will be validated on first manual dispatch after merge

### DIFF
--- a/.github/workflows/deepspeed-regression-test.yml
+++ b/.github/workflows/deepspeed-regression-test.yml
@@ -53,7 +53,7 @@ jobs:
           echo "${{ secrets.SLURM_SSH_KEY }}" > ~/.ssh/slurm_key
           chmod 600 ~/.ssh/slurm_key
           for i in {1..5}; do
-            if ssh-keyscan -H ${{ env.SLURM_HOST }} >> ~/.ssh/known_hosts 2>/dev/null; then
+            if ssh-keyscan -H "${{ env.SLURM_HOST }}" >> ~/.ssh/known_hosts 2>/dev/null; then
               echo "SSH keyscan successful"
               break
             fi
@@ -66,9 +66,9 @@ jobs:
         run: |
           BUILD_ID="${{ github.run_id }}"
           REMOTE_CODE_PATH="${{ env.BASE_PATH }}/deepspeed-tests/code-${BUILD_ID}"
-          echo "remote_code_path=$REMOTE_CODE_PATH" >> $GITHUB_OUTPUT
-          echo "REMOTE_CODE_PATH=$REMOTE_CODE_PATH" >> $GITHUB_ENV
-          echo "BUILD_ID=$BUILD_ID" >> $GITHUB_ENV
+          echo "remote_code_path=$REMOTE_CODE_PATH" >> "$GITHUB_OUTPUT"
+          echo "REMOTE_CODE_PATH=$REMOTE_CODE_PATH" >> "$GITHUB_ENV"
+          echo "BUILD_ID=$BUILD_ID" >> "$GITHUB_ENV"
 
       - name: Transfer Code to Cluster
         run: |
@@ -90,8 +90,8 @@ jobs:
         run: |
           QLORA_SQSH="/fsx/apps/qwen3-qlora-${{ env.BUILD_ID }}.sqsh"
           GPT_SQSH="/fsx/apps/deepspeed-${{ env.BUILD_ID }}.sqsh"
-          echo "qlora_sqsh=$QLORA_SQSH" >> $GITHUB_OUTPUT
-          echo "gpt_sqsh=$GPT_SQSH" >> $GITHUB_OUTPUT
+          echo "qlora_sqsh=$QLORA_SQSH" >> "$GITHUB_OUTPUT"
+          echo "gpt_sqsh=$GPT_SQSH" >> "$GITHUB_OUTPUT"
 
           # Build QLoRA container first
           ssh -i ~/.ssh/slurm_key -o StrictHostKeyChecking=no -o ConnectTimeout=30 \
@@ -174,7 +174,7 @@ jobs:
           echo "${{ secrets.SLURM_SSH_KEY }}" > ~/.ssh/slurm_key
           chmod 600 ~/.ssh/slurm_key
           for i in {1..5}; do
-            if ssh-keyscan -H ${{ env.SLURM_HOST }} >> ~/.ssh/known_hosts 2>/dev/null; then
+            if ssh-keyscan -H "${{ env.SLURM_HOST }}" >> ~/.ssh/known_hosts 2>/dev/null; then
               echo "SSH keyscan successful"
               break
             fi
@@ -189,10 +189,10 @@ jobs:
           REMOTE_CODE_PATH="${{ needs.build-containers.outputs.remote_code_path }}"
           LOG_DIR="${{ env.HOME_PATH }}/ds-qlora-logs-${BUILD_ID}-${{ matrix.zero_config }}"
           SQSH_PATH="${{ needs.build-containers.outputs.qlora_sqsh }}"
-          echo "REMOTE_CODE_PATH=$REMOTE_CODE_PATH" >> $GITHUB_ENV
-          echo "LOG_DIR=$LOG_DIR" >> $GITHUB_ENV
-          echo "BUILD_ID=$BUILD_ID" >> $GITHUB_ENV
-          echo "SQSH_PATH=$SQSH_PATH" >> $GITHUB_ENV
+          echo "REMOTE_CODE_PATH=$REMOTE_CODE_PATH" >> "$GITHUB_ENV"
+          echo "LOG_DIR=$LOG_DIR" >> "$GITHUB_ENV"
+          echo "BUILD_ID=$BUILD_ID" >> "$GITHUB_ENV"
+          echo "SQSH_PATH=$SQSH_PATH" >> "$GITHUB_ENV"
 
       - name: Create Log Directory
         run: |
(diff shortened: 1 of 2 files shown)


3.test_cases/pytorch/deepspeed/qlora/requirements.txt
```diff
@@ -12,7 +12,7 @@ peft>=0.10.0,<1.0.0
 bitsandbytes>=0.42.0,<1.0.0
 
 # Dataset and Hugging Face
-datasets>=2.18.0,<3.0.0
+datasets>=2.18.0
```
KeitaW: ### `datasets` upper bound removed

The `<3.0.0` upper bound was removed. The repo convention is to include upper-bound version pins to prevent surprise breakage from major releases. If `datasets` 3.x is now needed, I'd suggest updating to `datasets>=3.0.0,<4.0.0`. If this was just to unblock a compatibility issue, a comment explaining why the bound was removed would help.

.github/workflows/deepspeed-regression-test.yml
```diff
@@ -0,0 +1,545 @@
...
+        id: setup
+        run: |
+          BUILD_ID="${{ github.run_id }}"
+          REMOTE_TEST_PATH="${{ env.BASE_PATH }}/deepspeed-tests/qlora-${BUILD_ID}-${{ matrix.zero_config }}"
+          LOG_DIR="${{ env.HOME_PATH }}/ds-qlora-logs-${BUILD_ID}-${{ matrix.zero_config }}"
+          echo "remote_test_path=$REMOTE_TEST_PATH" >> $GITHUB_OUTPUT
+          echo "log_dir=$LOG_DIR" >> $GITHUB_OUTPUT
+          echo "REMOTE_TEST_PATH=$REMOTE_TEST_PATH" >> $GITHUB_ENV
+          echo "LOG_DIR=$LOG_DIR" >> $GITHUB_ENV
+          echo "BUILD_ID=$BUILD_ID" >> $GITHUB_ENV
```
KeitaW: ### Unquoted `$GITHUB_OUTPUT` and `$GITHUB_ENV` redirects

PR #1024 specifically fixed this pattern across all existing workflows (SC2086 — word splitting on unquoted variables). I'd suggest quoting for consistency:

```suggestion
          echo "remote_test_path=$REMOTE_TEST_PATH" >> "$GITHUB_OUTPUT"
          echo "log_dir=$LOG_DIR" >> "$GITHUB_OUTPUT"
          echo "REMOTE_TEST_PATH=$REMOTE_TEST_PATH" >> "$GITHUB_ENV"
          echo "LOG_DIR=$LOG_DIR" >> "$GITHUB_ENV"
          echo "BUILD_ID=$BUILD_ID" >> "$GITHUB_ENV"
```

Same pattern appears in the gpt-103b job (~lines 317-325) and elsewhere.

.github/workflows/deepspeed-regression-test.yml
```diff
@@ -0,0 +1,545 @@
...
+          role-to-assume: ${{ env.AWS_ROLE_ARN }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Setup SSH Key
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.SLURM_SSH_KEY }}" > ~/.ssh/slurm_key
+          chmod 600 ~/.ssh/slurm_key
+          for i in {1..5}; do
+            if ssh-keyscan -H ${{ env.SLURM_HOST }} >> ~/.ssh/known_hosts 2>/dev/null; then
```
KeitaW: ### Unquoted `${{ env.SLURM_HOST }}` in ssh-keyscan

PR #1024 fixed this same pattern in all existing workflows. I'd suggest:

```suggestion
            if ssh-keyscan -H "${{ env.SLURM_HOST }}" >> ~/.ssh/known_hosts 2>/dev/null; then
```

Same at ~line 303 in the gpt-103b job.

.github/workflows/deepspeed-regression-test.yml
```diff
@@ -0,0 +1,545 @@
...
+            sleep 10
+          done
+
+      - name: Build QLoRA Container
+        run: |
+          QLORA_DIR="${{ env.REMOTE_TEST_PATH }}/3.test_cases/pytorch/deepspeed/qlora"
+          SQSH_PATH="/fsx/apps/qwen3-qlora-${{ env.BUILD_ID }}-${{ matrix.zero_config }}.sqsh"
+          echo "SQSH_PATH=$SQSH_PATH" >> $GITHUB_ENV
+
+          ssh -i ~/.ssh/slurm_key -o StrictHostKeyChecking=no -o ConnectTimeout=30 \
```
KeitaW, line 97: ### Missing SSH keepalive on long-running container build step

This `ssh -o ConnectTimeout=30` runs Docker build + enroot import which can take 10+ minutes. PR #1024 added `-o ServerAliveInterval=60 -o ServerAliveCountMax=5` to the equivalent FSDP and Megatron build steps to prevent SSH timeout. I'd suggest adding the same here and on the gpt-103b build step (~line 295).

.github/workflows/deepspeed-regression-test.yml
```diff
@@ -0,0 +1,545 @@
...
+            ${{ env.SLURM_USER }}@${{ env.SLURM_HOST }} << EOF
+          set -ex
+          export HF_HOME=/fsx/.cache/huggingface
+          export HF_TOKEN=${HF_TOKEN}
+          mkdir -p \$HF_HOME
+          python3 -c "
+          from huggingface_hub import snapshot_download
+          from datasets import load_dataset
+          print('Downloading Qwen/Qwen3-8B...')
+          snapshot_download('Qwen/Qwen3-8B', token='${HF_TOKEN}')
```
KeitaW: ### HF_TOKEN interpolated directly into Python string literal

Since the heredoc is unquoted (`<< EOF`), the shell expands `${HF_TOKEN}` before Python sees it. If the token contains a single quote, this breaks the Python syntax. Since `HF_TOKEN` is already exported as an environment variable two lines above, `snapshot_download` and `load_dataset` will automatically pick it up via the `huggingface_hub` library. I'd suggest dropping the explicit `token=` parameter:

```suggestion
          from huggingface_hub import snapshot_download
          from datasets import load_dataset

          snapshot_download('Qwen/Qwen3-8B')
```

KeitaW:
### CI is failing — 3/3 DeepSpeed regression jobs red

All three jobs from the latest run ([#23436674590](https://github.com/awslabs/awsome-distributed-training/actions/runs/23436674590)) are failing:

#### QLoRA (zero2 + zero3): `pip install` dependency conflict

Both QLoRA jobs fail at `RUN pip install --no-cache-dir -r requirements.txt` with:

```
datasets 2.18.0 depends on dill<0.3.9 and >=0.3.0
The user requested (constraint) dill==0.3.9
ERROR: ResolutionImpossible
```

The root cause: `requirements.txt` pins `dill==0.3.9` but `datasets>=2.18.0` (with the upper bound removed in this PR) requires `dill<0.3.9`. The fix that removed `<3.0.0` from the `datasets` constraint actually surfaced this — `datasets` 3.x still depends on `dill<0.3.9`, so `dill==0.3.9` is an exact boundary violation.

Fix options:
1. Pin `dill==0.3.8` (or `dill>=0.3.0,<0.3.9`) to match what `datasets` requires
2. Or restore the upper bound: `datasets>=2.18.0,<3.0.0` with `dill<0.3.9`

#### GPT-103B: SSH timeout during `enroot import`

The GPT-103B job fails with `client_loop: send disconnect: Broken pipe` during the `enroot import` step (extracting the Docker image to squashfs). This is the SSH keepalive issue I flagged in the review — the `enroot import` takes ~5 minutes with no output, causing the SSH session to drop. Adding `-o ServerAliveInterval=60 -o ServerAliveCountMax=5` to the `ssh` command should fix this.

PR #1063, opened 2026-04-14
Adding OpenEnv Wordle GRPO sample

allela-roy:
## Purpose

Adding a sample for OpenEnv Wordle GRPO Training on SageMaker HyperPod (EKS)

## Changes

New sample

## Test Plan

**Environment:**
- AWS Service: SageMaker HyperPod (EKS)
- Instance type: ml.g6e.12xlarge (4x NVIDIA L40S, 48 GB each)
- Number of nodes: 2 

**Test commands:**
```bash
# 1. Set up environment variables
source env_vars

# 2. Build and push container image
aws ecr get-login-password --region $AWS_REGION | docker login --username AWS --password-stdin $REGISTRY
docker build -t ${REGISTRY}${IMAGE}:${TAG} .
docker push ${REGISTRY}${IMAGE}:${TAG}

# 3. Deploy Wordle environment
envsubst < kubernetes/openenv-wordle-env.yaml | kubectl apply -f -
kubectl wait --for=condition=Available deployment/openenv-wordle --timeout=300s

# 4. Verify Wordle env health
kubectl run test-env --rm -it --restart=Never --image=curlimages/curl -- \
  curl -s http://openenv-wordle:7860/health

# 5. Launch single-GPU colocate training
envsubst < kubernetes/train-grpo-wordle.yaml | kubectl apply -f -
kubectl wait --for=condition=Ready pod/grpo-wordle --timeout=300s
```

## Test Results

Single-node colocate mode verified with 6+ training steps, 0 pod restarts, all 4 reward signals active:

```
Step 1 (epoch 0.011): loss=1.696e-32, reward=0.9385, reward_correct=0.2688, greens=0.1875, yellows=0.175, repetition=0.3073, step_time=152.9s
Step 2 (epoch 0.022): loss=3.387e-32, reward=0.9281, reward_correct=0.25,   greens=0.1531, yellows=0.2125, repetition=0.3125, step_time=153.0s
Step 3 (epoch 0.032): loss=6.355e-33, reward=0.887,  reward_correct=0.2641, greens=0.1906, yellows=0.1562, repetition=0.276,  step_time=154.4s
Step 4 (epoch 0.043): loss=-2.26e-32, reward=0.9156, reward_correct=0.2594, greens=0.1625, yellows=0.1969, repetition=0.2969, step_time=152.4s
Step 5 (epoch 0.054): loss=1.503e-32, reward=0.9208, reward_correct=0.2688, greens=0.1719, yellows=0.2094, repetition=0.2708, step_time=157.3s
Step 6 (epoch 0.065): loss=-6.63e-32, reward=1.013,  reward_correct=0.3063, greens=0.2062, yellows=0.2063, repetition=0.2943, step_time=151.1s
```

Pod status throughout training:
```
NAME          READY   STATUS    RESTARTS   AGE
grpo-wordle   1/1     Running   0          25m
```

## Directory Structure

```
3.test_cases/
```

## Checklist

- [X ] I have read the [contributing guidelines](https://github.com/awslabs/awsome-distributed-training/blob/main/CONTRIBUTING.md).
- [ X] I am working against the latest `main` branch.
- [X ] I have searched existing open and recently merged PRs to confirm this is not a duplicate.
- [ X] The contribution is self-contained with documentation and scripts.
- [X ] External dependencies are pinned to a specific version or tag (no `latest`).
- [X ] A README is included or updated with prerequisites, instructions, and known issues.
- [ X] New test cases follow the [expected directory structure](#directory-structure).


KeitaW (commented):
## Review 1/5 — Structure & Repository Hygiene

Thanks for this contribution — the OpenEnv Wordle GRPO sample is a great addition to the TRL test cases. I have some findings across several categories; posting them as themed batches for easier navigation.

This batch covers image pinning and Dockerfile robustness.

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM public.ecr.aws/hpc-cloud/nccl-tests:latest
```
KeitaW: ### Base image uses `:latest` tag

Per repo conventions, container image tags must use fixed versions, never `latest`. This makes builds non-reproducible — a new push to `nccl-tests:latest` could silently break the build. I'd suggest pinning to a specific digest or tag. You can find available tags with:
```bash
aws ecr-public describe-image-tags --repository-name hpc-cloud/nccl-tests --region us-east-1
```

Reference: [review-checklist](https://github.com/awslabs/awsome-distributed-training/blob/main/CONTRIBUTING.md), "Fixed version tags."

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/openenv-wordle-env.yaml
```diff
@@ -0,0 +1,77 @@
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: sagemaker.amazonaws.com/node-health-status
+                    operator: In
+                    values:
+                      - Schedulable
+      containers:
+      - name: wordle-env
+        image: registry.hf.space/burtenshaw-wordle:latest
```
KeitaW, line 43: ### Environment image uses `:latest`

`registry.hf.space/burtenshaw-wordle:latest` — if there's no versioned tag for this HuggingFace Space image, I'd suggest adding a comment noting that this is externally managed and may change without notice.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle-multigpu.yaml
```diff
@@ -0,0 +1,169 @@
...
+        nodeSelectorTerms:
+          - matchExpressions:
+              - key: sagemaker.amazonaws.com/node-health-status
+                operator: In
+                values:
+                  - Schedulable
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          preference:
+            matchExpressions:
```
KeitaW, line 42: ### Init container image uses `:latest`

I'd suggest pinning to a specific version:

```suggestion
      image: curlimages/curl:8.12.1
```

3.test_cases/pytorch/trl/openenv-wordle-grpo/env_vars.example
```diff
@@ -0,0 +1,32 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+# Docker / ECR
+export AWS_REGION=$(aws ec2 describe-availability-zones --output text --query 'AvailabilityZones[0].[RegionName]')
+export ACCOUNT=$(aws sts get-caller-identity --query Account --output text)
+export REGISTRY=${ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/
+export IMAGE=openenv-wordle-grpo
+export TAG=latest
```
KeitaW: ### Default TAG should not be `latest`

Combined with the `:latest` base image, this creates a double-unpinned build. I'd suggest defaulting to a version-based tag:

```suggestion
export TAG=v1.0
```

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+old = 'from transformers.utils.import_utils import _is_package_available'; \
+new = 'from transformers.utils.import_utils import _is_package_available as _orig_is_pkg_avail\n\ndef _is_package_available(pkg_name, return_version=False):\n    result = _orig_is_pkg_avail(pkg_name, return_version=return_version)\n    if return_version:\n        return result if isinstance(result, tuple) else (result, None)\n    return result[0] if isinstance(result, tuple) else result'; \
+p.write_text(src.replace(old, new))"
+
+# Fix GRPOTrainer compatibility with transformers 5.x (warnings_issued attr)
+# and vLLM 0.8.x (logprobs_mode kwarg not supported).
+RUN python -c "\
+import pathlib; \
+p = pathlib.Path('/opt/miniconda3/lib/python3.12/site-packages/trl/trainer/grpo_trainer.py'); \
+src = p.read_text(); \
```
KeitaW: ### Inline TRL patches are fragile across versions

These `python -c` blocks patch TRL source files at exact paths using string replacement. If TRL updates these lines (even whitespace changes), the patches will silently fail to match and the `str.replace()` will be a no-op, leaving the bugs in place.

I'd suggest:
1. Adding verification after each patch: `python -c "assert 'patched_text' in open('/path/to/file').read()"` 
2. Adding a comment noting which TRL issue/PR tracks the upstream fix so future maintainers know when the patches can be removed

or fix TRL directory (nice PR opportunity).

KeitaW (changes requested):
Nice PR few nits.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle-multigpu.yaml
```diff
@@ -0,0 +1,169 @@
...
+      - name: HF_HOME
+        value: "/fsx/hf_cache"
+      command:
+        - trl
+        - vllm-serve
+        - --model
+        - $MODEL_NAME
+        - --host
+        - "0.0.0.0"
+        - --port
```
KeitaW, line 89: ### Trainer container requests 1 GPU but claims to use GPUs 1-3

The manifest header (lines 6-8) describes "GPU 0: vLLM server, GPU 1-3: FSDP training" and line 121 uses `accelerate launch --num_processes 1`. But Kubernetes GPU device plugin allocates GPUs at the container level — requesting 1 GPU here means the trainer gets exactly 1 GPU, not 3.

To achieve the described 1+3 split, I'd suggest:

```suggestion
        requests:
          nvidia.com/gpu: 3
          memory: "60Gi"
```

And updating `accelerate launch --num_processes 1` on line 121 to `--num_processes 3`.

Alternatively, if the intent is really 1+1 (total 2 GPUs), the README section "4. Multi-GPU Training" and the manifest header comments should be updated to match.

This is the most significant functional issue I noticed.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle.yaml
```diff
@@ -0,0 +1,116 @@
...
+kind: Pod
+metadata:
+  name: grpo-wordle
+  namespace: ${NAMESPACE}
+  labels:
+    app: grpo-wordle
+  annotations:
+    sagemaker.amazonaws.com/enable-job-auto-resume: "true"
+    sagemaker.amazonaws.com/job-max-retry-count: "3"
+spec:
```
KeitaW, line 26: ### `$HF_TOKEN` exposed as plain-text env var

After `envsubst`, this embeds the actual token in the manifest YAML. Anyone with `kubectl get pod -o yaml` access can read it. I'd suggest using a Kubernetes Secret instead:

```suggestion
      - name: HF_TOKEN
        valueFrom:
          secretKeyRef:
            name: hf-token
            key: token
```

This applies to all manifests in this PR (`train-grpo-wordle.yaml`, `train-grpo-wordle-multigpu.yaml`, `inference-wordle.yaml`).

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
KeitaW: ### Missing copyright header

The Dockerfile is missing the standard repo copyright header. All other files in this PR include it. I'd suggest adding at the top:

```
```

(The training script uses Apache 2.0 from the original HuggingFace source, which is fine for copied code — but the Dockerfile is original.)

KeitaW (commented):
## Review 2/5 — Deployment Pipeline

This batch covers the multi-GPU manifest GPU allocation issue, envsubst usage, missing copyright header, and credential handling.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle.yaml
```diff
@@ -0,0 +1,116 @@
...
+      imagePullPolicy: Always
+      resources:
+        requests:
+          nvidia.com/gpu: $NUM_GPU_PER_NODE
+          memory: "180Gi"
+        limits:
+          nvidia.com/gpu: $NUM_GPU_PER_NODE
+          memory: "180Gi"
+      env:
+      - name: PYTORCH_CUDA_ALLOC_CONF
```
KeitaW, line 63: ### `/dev/shm` mounted as hostPath instead of emptyDir with Memory medium

Using `hostPath: /dev/shm` shares the host's shared memory with the pod, which means multiple pods on the same node can interfere with each other. The preferred Kubernetes pattern is an isolated emptyDir:

```suggestion
    - name: shmem
      emptyDir:
        medium: Memory
        sizeLimit: 16Gi
```

This gives the pod its own isolated shared memory allocation. The same applies to the other manifests in this PR (`train-grpo-wordle-multigpu.yaml`, `inference-wordle.yaml`).

KeitaW (commented):
## Review 3/5 — Infrastructure & NCCL Configuration

Shared memory mounting pattern.

KeitaW (commented):
## Review 4/5 — Documentation Consistency & Training Code Quality

Dangling references, missing newline, and dependency pinning.

3.test_cases/pytorch/trl/openenv-wordle-grpo/README.md
```diff
@@ -0,0 +1,387 @@
...
+# Watch pod status
+kubectl get pod grpo-wordle -w
+
+# Follow training logs
+kubectl logs grpo-wordle -f
+
+# Check checkpoints
+kubectl exec fsx-storage-manager -- ls -la /fsx/checkpoints/wordle-grpo/
+```
+
```
KeitaW, line 272: ### README references `fsx-storage-manager` pod that isn't created by the PR

```
kubectl exec fsx-storage-manager -- ls -la /fsx/checkpoints/wordle-grpo/
```

There's no manifest or instructions for creating an `fsx-storage-manager` pod. A user following the README would get `Error from server (NotFound)`. I'd suggest replacing with a command that uses the training pod itself:

```suggestion
kubectl exec grpo-wordle -- ls -la /fsx/checkpoints/wordle-grpo/
```

3.test_cases/pytorch/trl/openenv-wordle-grpo/src/wordle_prompt.txt
```diff
@@ -0,0 +1,105 @@
...
+- Use lowercase only
+- One guess per response
+- Must be exactly 5 letters
+- Must be a real English word from standard dictionaries
+- Never repeat a previous guess
+- Always include brief reasoning before your guess
+
+## YOUR GOAL
+
+Solve the Wordle in as few guesses as possible by strategically using feedback to eliminate impossible words and narrow down the solution space efficiently.
```
KeitaW: ### Missing newline at end of file

Per the repo's `.editorconfig` (`insert_final_newline = true`), all files should end with a newline character.

3.test_cases/pytorch/trl/openenv-wordle-grpo/src/requirements.txt
```diff
@@ -0,0 +1,23 @@
...
+packaging
+ninja
+torch==2.6.0
+transformers>=4.48.0
+datasets>=2.17.0
+accelerate>=0.27.0
+peft>=0.9.0
+deepspeed>=0.14.0
+wandb
+trl[vllm]
```
KeitaW: ### Dependencies use open-ended lower bounds without upper pins

Several dependencies use `>=` without upper bounds, which can pull incompatible major versions over time. I'd suggest adding upper bounds:

```suggestion
transformers>=4.48.0,<5.0
datasets>=2.17.0,<3.0
accelerate>=0.27.0,<2.0
peft>=0.9.0,<1.0
deepspeed>=0.14.0,<1.0
```

Also note: `trl[vllm]` is installed here (pulling in a vllm version), then TRL is upgraded to 0.26.2 with `--no-deps` in the Dockerfile. The vllm version is therefore determined by whatever `trl[vllm]` pulled initially, not by TRL 0.26.2's constraints. This may be intentional but is worth documenting.

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+src = src.replace(\
+    'model.warnings_issued[\"estimate_tokens\"] = True', \
+    'if hasattr(model, \"warnings_issued\"): model.warnings_issued[\"estimate_tokens\"] = True'); \
+src = src.replace('logprobs_mode=\"processed_logprobs\",\n', ''); \
+p.write_text(src)"
+
+# Flash Attention (requires torch, must install separately with --no-build-isolation)
+RUN pip install flash-attn>=2.5.0 --no-build-isolation
+
```
KeitaW, line 49: ### `flash-attn` has no upper bound pin

Flash Attention builds from source and is sensitive to CUDA/PyTorch version combinations. An unpinned upper bound could pull a version incompatible with torch 2.6.0. I'd suggest:

```suggestion
RUN pip install flash-attn>=2.5.0,<3.0 --no-build-isolation
```

KeitaW (commented):
## Review 5/5 — Things That Look Great

- **Excellent README structure.** The pipeline architecture diagram, prerequisite list, step-by-step instructions with expected outputs, configuration reference table, and reward function documentation are all well-done. This is one of the more polished READMEs I've seen in a test case contribution.
- **OpenEnv is a compelling choice for demonstrating agentic RL.** The client-server architecture (environment as a Kubernetes service, training as a separate pod) showcases real Kubernetes-native ML workflows. The "What is OpenEnv?" section does a great job explaining why this matters.
- **Two deployment modes** (single-GPU colocate and multi-GPU server) give users flexibility. Colocate is great for quick iteration; server mode demonstrates a production-like split.
- **HyperPod-aware manifests.** The tolerations, node selectors, health-check affinity, and auto-resume annotations follow HyperPod best practices consistently across all manifests.
- **The training script is well-structured.** Clean separation between rollout logic (`rollout_once`), reward functions, and the main entrypoint. The argparse interface makes it easy to experiment.
- **Four distinct reward signals** (correctness, greens, yellows, repetition) provide rich training signal and make the training dynamics interpretable from logs.
- **Copyright headers present** on all shell scripts, YAML manifests, and the env_vars example.
- **The Wordle environment as a Deployment with readiness/liveness probes** ensures the environment is healthy before training starts — much better than a bare pod.

---

### Cross-cutting note: `envsubst` variable whitelist

The README instructs users to run `envsubst < kubernetes/*.yaml | kubectl apply -f -` without an explicit variable whitelist. Without a whitelist, `envsubst` substitutes every `$VAR` in the YAML. I'd suggest using explicit whitelists in the README commands, e.g.:
```bash
envsubst '$NAMESPACE $REGISTRY $IMAGE $TAG $HF_TOKEN $MODEL_NAME $VLLM_MODE $NUM_GPU_PER_NODE $NUM_GENERATIONS $GRADIENT_ACCUMULATION_STEPS $LEARNING_RATE $INSTANCE_TYPE $FSX_PVC' < kubernetes/train-grpo-wordle.yaml | kubectl apply -f -
```

paragao (changes requested):
Thanks for the contribution — the README quality and overall structure are solid. A few items to address before we can merge:

## 1. HF_TOKEN plain-text exposure (Security — Blocking)

All Kubernetes manifests (`train-grpo-wordle.yaml`, `train-grpo-wordle-multigpu.yaml`, `inference-wordle.yaml`) pass `HF_TOKEN` as a plain-text `value:` field after `envsubst` processing. Anyone with `kubectl get pod -o yaml` access can read the token.

Please replace the plain-text value with a Kubernetes Secret reference, for example:

```yaml
- name: HF_TOKEN
  valueFrom:
    secretKeyRef:
      name: hf-token
      key: token
```

And add instructions to the README for creating the secret:
```bash
kubectl create secret generic hf-token --from-literal=token=$HF_TOKEN
```

## 2. Multi-GPU manifest: GPU allocation mismatch (Blocking)

The comments in `train-grpo-wordle-multigpu.yaml` describe:
> GPU 0: Dedicated vLLM server
> GPU 1-3: FSDP-sharded GRPO training

But both the `vllm-server` and `trainer` containers request only **1 GPU each** (`nvidia.com/gpu: 1`), and the training command uses `accelerate launch --num_processes 1`.

This means only 2 GPUs are used total, not 4 as described. Could you clarify the intended configuration?

- If the intent is 1 GPU for vLLM + 3 GPUs for FSDP: the `trainer` container should request `nvidia.com/gpu: 3` and use `--num_processes 3`.
- If the intent is 1 GPU for vLLM + 1 GPU for training: the comments and documentation should be updated to match.

## 3. Add upper bounds to dependency versions in `requirements.txt`

The current dependencies use open-ended `>=` bounds (e.g., `transformers>=4.48.0`). Given that the Dockerfile already patches TRL for transformers 5.x compatibility issues, unbounded dependencies are a real reproducibility risk.

Please add upper bounds, for example:
```
transformers>=4.48.0,<6.0
datasets>=2.17.0,<4.0
accelerate>=0.27.0,<2.0
peft>=0.9.0,<1.0
deepspeed>=0.14.0,<1.0
```

## 4. Missing copyright header in `Dockerfile`

The Dockerfile is the only file in this PR missing the standard copyright header. Please add the following two lines at the very top of the file (before the `FROM` line):

```dockerfile
```

All other files in the PR already have this header.

## 5. Double-check `env_vars.example` for sensitive data

Please review the `env_vars.example` file and confirm it does not contain any real sensitive data such as AWS account IDs, FSx filesystem IDs, credentials, or tokens. The current placeholders look correct (e.g., `<your-huggingface-token>`, `<your-hyperpod-cluster-name>`), but please verify before merging.

## 6. `fsx-storage-manager` pod reference in README (Non-blocking)

The README references a pod called `fsx-storage-manager` in two places:

**Checking checkpoints:**
```bash
kubectl exec fsx-storage-manager -- ls -la /fsx/checkpoints/wordle-grpo/
```

**Cleanup section:**
```bash
kubectl delete pod grpo-wordle grpo-wordle-multigpu inference-wordle fsx-storage-manager 2>/dev/null
```

However, no manifest in this PR creates a `fsx-storage-manager` pod. Users following the README will get a "pod not found" error on these commands.

Could you clarify the intent here?
- If `fsx-storage-manager` is a pod users are expected to have in their cluster already, please add a note explaining that.
- If it was included by mistake, please remove it from the README commands and use the `grpo-wordle` pod instead (or another existing pod that has the FSx volume mounted).
- If there should be a manifest for it, please add one.

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
allela-roy: @KeitaW , don't we want to use the latest since this contains the recommended versions of EFA, NCCL libs?

paragao: @allela-roy it doesn't, does it? As of today (April 16th 2026), these are the latest versions on that docker image:

CUDA 12.8
EFA Installer 1.43.0
aws-ofi-nccl 1.16.3
libfabric 1.27.0

So, either you make sure you tag a specific version and on your Dockerfile you update the libraries, or rely on latest anyway and rebuild everything even if that means downgrading a library. 

Either, it is a best practice, not a blocker. Chose your own way and do it, but make it reproducible and that it does not break in the near future.

allela-roy: ACK, fixed.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/openenv-wordle-env.yaml
allela-roy: Added comment

3.test_cases/pytorch/trl/openenv-wordle-grpo/env_vars.example
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle-multigpu.yaml
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle.yaml
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle.yaml
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/README.md
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/src/wordle_prompt.txt
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/src/requirements.txt
allela-roy: Fixed

3.test_cases/pytorch/trl/openenv-wordle-grpo/Dockerfile
allela-roy: Fixed

allela-roy:
Thanks @KeitaW, @paragao for the detailed review. I've fixed the issues. Please review.

paragao (approved):
Thank you for reviewing the changes requested! LGTM.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/openenv-wordle-env.yaml
```diff
@@ -0,0 +1,78 @@
...
+              - matchExpressions:
+                  - key: sagemaker.amazonaws.com/node-health-status
+                    operator: In
+                    values:
+                      - Schedulable
+      containers:
+      - name: wordle-env
+        # NOTE: This image is externally managed by the HuggingFace Space owner
+        # and may change without notice. No versioned tags are available.
+        image: registry.hf.space/burtenshaw-wordle:latest
```
KeitaW, line 43: ### Following up — Wordle env image is still on `:latest`

In the previous round I asked about pinning `registry.hf.space/burtenshaw-wordle:latest`, and the response was to add the inline note above. The comment helps a future reader understand the situation, but it doesn't actually solve the reproducibility problem — anyone running this test case in 6 months can still get a silently-different environment. Since `requirements.txt` already pins the wordle Space to a specific commit (`@77bdfedb9144b341db1c37449328709ccbed7da3`), the image side could match. Two concrete options:

- Pin to a digest: `image: registry.hf.space/burtenshaw-wordle@sha256:<digest>` (resolve once with `docker buildx imagetools inspect registry.hf.space/burtenshaw-wordle:latest`).
- Or build a small image locally from the pinned commit, push it to ECR alongside the trainer image, and reference that.

I don't want to block on this — happy to treat it as a known-limitation if you'd rather, in which case could the inline comment be expanded to spell out the reproducibility caveat (e.g. "if this image breaks the test case, regenerate from the commit pinned in `src/requirements.txt`")? That gives the next maintainer a recovery path. Reference: repo precedent in [`3.test_cases/pytorch/megatron/megatron-lm/Dockerfile`](https://github.com/awslabs/awsome-distributed-training/blob/main/3.test_cases/pytorch/megatron/megatron-lm/Dockerfile) which pins by digest.

KeitaW (commented):
## Re-review batch 1/3 — Structure & Repository Hygiene

Thanks for the quick turnaround on the previous round. Significant progress: `HF_TOKEN` now flows through a `Secret`, the multi-GPU GPU allocation is correct (1+3=4), the `Dockerfile` has its copyright header, the `fsx-storage-manager` dangling reference is gone, and the README's `envsubst` calls use whitelists. This batch covers one follow-up that I think is still open from the first round.

3.test_cases/pytorch/trl/openenv-wordle-grpo/kubernetes/train-grpo-wordle.yaml
```diff
@@ -0,0 +1,109 @@
...
+
+apiVersion: v1
+kind: Pod
+metadata:
+  name: grpo-wordle
+  namespace: ${NAMESPACE}
+  labels:
+    app: grpo-wordle
+  annotations:
+    sagemaker.amazonaws.com/enable-job-auto-resume: "true"
```
KeitaW, line 24: ### HyperPod auto-resume on a plain `Pod` — does it actually fire?

(Same applies to `train-grpo-wordle-multigpu.yaml`.)

Both training manifests are plain `Pod`s in `${NAMESPACE}` (defaulted to `default` in `env_vars.example`) and carry `sagemaker.amazonaws.com/enable-job-auto-resume: "true"` plus `sagemaker.amazonaws.com/job-max-retry-count: "3"`. The README sells this as one of the reasons to run the sample on HyperPod.

I want to flag this as a verification ask rather than a claim: AWS's HyperPod EKS docs scope auto-resume to **PyTorchJob via the Kubeflow Training Operator**, with the additional constraint that *"the job must be running under the `kubeflow` namespace or a namespace prefixed with `hyperpod`"* ([Running jobs using the SageMaker HyperPod CLI](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-hyperpod-eks-run-jobs-hyperpod-cli.html), [Introducing Amazon EKS support in SageMaker HyperPod](https://aws.amazon.com/blogs/machine-learning/introducing-amazon-eks-support-in-amazon-sagemaker-hyperpod/)). The docs don't explicitly say "annotation is a no-op on a plain Pod in the default namespace" — but they also don't document a path where it isn't. So a few possibilities:

1. The annotation works on plain Pods in any namespace, and the docs are just incomplete → could you point me at where you confirmed this? A line in the README would help future readers.
2. The annotation is silently ignored under the current setup → in which case this is a real gap, and the minimal fix is to wrap each manifest as a single-replica `PyTorchJob` (the operator's PodTemplate supports multiple containers, so the vLLM-server + trainer two-container layout in `train-grpo-wordle-multigpu.yaml` carries over unchanged).
3. There's a hybrid path I'm missing.

Note that PyTorchJob's "distributed env var injection" (`MASTER_ADDR`, `WORLD_SIZE`, `RANK`) is irrelevant here because Accelerate's `--num_processes 3` does its own intra-node spawn — so this isn't the generic "use PyTorchJob for distributed training" rule from the checklist. The concrete benefit for *this* PR is HyperPod auto-resume integration plus the namespace requirement, nothing else.

KeitaW (commented):
## Re-review batch 2/3 — Deployment Pipeline

One new concern that I didn't catch in the first round, around HyperPod auto-resume. Inline comment anchored to `train-grpo-wordle.yaml`, but the same applies to `train-grpo-wordle-multigpu.yaml`.

KeitaW (commented):
## Re-review batch 3/3 — Documentation Consistency + Things That Look Great

Three small documentation nits, then a list of things that genuinely look great about this iteration.

---

## Things That Look Great

- **Significant progress since last review.** `HF_TOKEN` now flows through a `Secret`, the multi-GPU GPU allocation is correct (1 + 3 = 4), the `Dockerfile` has its copyright header, the `fsx-storage-manager` dangling reference is gone, and the README's `envsubst` calls use whitelists. That's a tight turnaround on substantive fixes.
- **The TRL patch verification step.** The `assert 'logprobs_mode="processed_logprobs"' not in src` check in the `Dockerfile` makes the build fail loudly if upstream changes formatting, instead of silently shipping a broken image. That's a genuinely good pattern that I'd love to see more widely in this repo.
- **Wordle environment as a Deployment with readiness/liveness probes** plus the `wait-for-env` initContainer in the multi-GPU manifest gives a clean ordering guarantee: the trainer can't start before the env is healthy. This is exactly the right Kubernetes idiom.
- **Auto-resume annotations and HyperPod tolerations are consistent across all three manifests.** Easy to copy as a template for future HyperPod EKS test cases (modulo the verification ask in batch 2).
- **README quality.** The "What is OpenEnv?" section, the architecture diagram, the side-by-side colocate vs. server-mode trade-offs, the configuration reference table, and the reward function table together make this one of the more polished READMEs in `3.test_cases/`. The actual training output snippet showing 6 steps with all four reward signals is a particularly nice touch — readers can sanity-check their own runs against it.
- **Pinned upstream commit for the `burtenshaw/wordle` Space** (`@77bdfedb9144b341db1c37449328709ccbed7da3`) in `requirements.txt`. Even though the env image still floats on `latest`, the Python client side is reproducible.
- **Clean separation of concerns in `train_wordle_grpo.py`** between `rollout_once`, the four reward functions, `rollout_func`, and `main`. Easy to read, and easy to modify a single reward without touching the trainer wiring.

3.test_cases/pytorch/trl/openenv-wordle-grpo/src/train_wordle_grpo.py
```diff
@@ -0,0 +1,519 @@
...
+    parser.add_argument(
+        "--num-epochs",
+        type=int,
+        default=1,
+        help="Number of training epochs.",
+    )
+    parser.add_argument(
+        "--save-interval",
+        type=int,
+        default=10,
```
KeitaW, line 171: ### `--save-interval` default mismatch between argparse and YAMLs

The argparse default here is `10`, but both training manifests pass `--save-interval 25`, and the README's Configuration Reference table also says `25`. It's not a bug because the YAMLs override the argparse default, but a user reading the script source first will see `10` and be surprised. Could you align the argparse default to `25`, or note in the help string that the YAMLs override it?

```suggestion
        default=25,
```

3.test_cases/pytorch/trl/openenv-wordle-grpo/README.md
```diff
@@ -0,0 +1,404 @@
...
+```bash
+docker build -t ${REGISTRY}${IMAGE}:${TAG} .
+```
+
+Expected output (abbreviated):
+
+```
+[+] Building 850.2s (12/12) FINISHED
+ => [stage-0  1/11] FROM public.ecr.aws/hpc-cloud/nccl-tests:latest
```
KeitaW, line 184: ### Build output snippet doesn't match the actual `FROM` line

The expected build output here shows `FROM public.ecr.aws/hpc-cloud/nccl-tests:latest`, but the real `Dockerfile` pins to a digest: `nccl-tests:latest@sha256:4498c1fa9e76...`. Builders relying on the README's expected output to double-check their build will see a different first line. Could you regenerate the snippet from an actual build, or just shorten it so it doesn't claim to be exact?

3.test_cases/pytorch/trl/openenv-wordle-grpo/README.md
```diff
@@ -0,0 +1,404 @@
...
+
+```bash
+# Check both containers
+kubectl get pod grpo-wordle-multigpu -w
+
+# Follow vLLM server logs
+kubectl logs grpo-wordle-multigpu -c vllm-server --tail=20
+
+# Follow training logs
+kubectl logs grpo-wordle-multigpu -c trainer -f
```
KeitaW, line 331: ### Multi-GPU and single-GPU output dirs diverge but the README only documents one

`train-grpo-wordle.yaml` writes to `/fsx/checkpoints/wordle-grpo`, while `train-grpo-wordle-multigpu.yaml` writes to `/fsx/checkpoints/wordle-grpo-multigpu`. Section 3.2's monitoring snippet correctly points readers at `/fsx/checkpoints/wordle-grpo/`, but section 4.1 doesn't have an equivalent — so users running the multi-GPU path may look in the wrong directory. Could you add a snippet here that points at the `wordle-grpo-multigpu` directory? For example:

```bash
kubectl exec grpo-wordle-multigpu -c trainer -- ls -la /fsx/checkpoints/wordle-grpo-multigpu/
```

KeitaW (approved):
LGTM. Thanks for addressing the issues.

PR #1069, opened 2026-04-29
Add Extensions framework for SageMaker HyperPod OnInitComplete

austincherian:
## Summary

Adds a modular Extensions framework under `1.architectures/5.sagemaker-hyperpod/Extensions/` for SageMaker HyperPod Slurm clusters using AMI-based configuration. Extensions run via `OnInitComplete` and can be independently enabled or disabled.

### run_extensions.sh (entrypoint)
Combined OnInitComplete entrypoint that orchestrates extension scripts. Each extension can be toggled on/off via feature flags. Handles logging, error handling, and execution ordering.

### detect-node/
Utility that detects node type (controller, compute, or login) using HyperPod platform config files (`provisioning_parameters.json` and `resource_config.json`). Writes results to `/opt/ml/config/nodeinfo.json` for use by downstream extensions. Runs first -- other extensions depend on its output.

### add-users/
User provisioning for HyperPod Slurm clusters:
- Creates POSIX users with consistent UIDs across all nodes
- Sets up home directories on shared filesystems (FSx for Lustre or OpenZFS)
- Generates SSH keypairs for passwordless inter-node access
- Registers users in Slurm accounting (controller only)
- Supports legacy CSV format (`shared_users.txt`) and YAML format (`shared_users.yaml`) with groups and per-group Slurm accounts

### observability/
Observability stack for HyperPod Slurm clusters:
- Node Exporter (system metrics)
- DCGM Exporter (GPU metrics)
- EFA Exporter (network fabric metrics)
- Slurm Exporter (job scheduler metrics)
- OpenTelemetry Collector with Amazon Managed Prometheus (AMP) remote write
- Node-type-aware configuration (different OTel configs for controller, compute, login nodes)

## Execution Order

```
run_extensions.sh
  |-- detect-node/detect_node.sh    (always runs first)
  |-- add-users/add_users.sh        (if ENABLE_ADD_USERS=true)
  |-- observability/setup_observability.sh  (if ENABLE_OBSERVABILITY=true)
```

## Deployment

Upload the Extensions directory to S3 and specify `run_extensions.sh` as the OnInitComplete target:
```json
{
    LifeCycleConfig: {
        OnInitComplete: run_extensions.sh,
        SourceS3Uri: s3://<bucket>/Extensions/
    }
}
```

## Testing

Tested on SageMaker HyperPod Slurm clusters with AMI-based configuration across controller, compute, and login node types.

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/Extensions/add-users/setup_ssh_keys.sh
+++ b/1.architectures/5.sagemaker-hyperpod/Extensions/add-users/setup_ssh_keys.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+#
+# Generates SSH keypairs on the shared filesystem for passwordless inter-node SSH.
+# First node to run creates the keypair; subsequent nodes find it already there.
+#
+# Usage: setup_ssh_keys.sh <users_file> <shared_fs_mount>
+
+set -e
+
+USERS_FILE="$1"
+SHARED_FS_MOUNT="$2"
+
+if [[ -z "$USERS_FILE" ]] || [[ -z "$SHARED_FS_MOUNT" ]]; then
+    echo "Usage: setup_ssh_keys.sh <users_file> <shared_fs_mount>"
+    exit 1
+fi
+
+setup_ssh() {
+    local username="$1"
+    local ssh_dir="${SHARED_FS_MOUNT}/${username}/.ssh"
+
+    if ! id -u "$username" >/dev/null 2>&1; then
+        echo "[WARN] User $username does not exist. Skipping SSH setup."
+        return 0
+    fi
+
+    mkdir -p "$ssh_dir"
+
+    if [[ ! -f "$ssh_dir/id_rsa" ]]; then
+        echo "[INFO] Generating SSH keypair for $username on shared filesystem"
+        ssh-keygen -t rsa -b 4096 -q -f "$ssh_dir/id_rsa" -N "" 2>/dev/null || true
+        cat "$ssh_dir/id_rsa.pub" >> "$ssh_dir/authorized_keys"
+    else
+        echo "[INFO] SSH keypair for $username already exists on shared filesystem"
+        if [[ -f "$ssh_dir/id_rsa.pub" ]]; then
+            if ! grep -qF "$(cat "$ssh_dir/id_rsa.pub")" "$ssh_dir/authorized_keys" 2>/dev/null; then
+                cat "$ssh_dir/id_rsa.pub" >> "$ssh_dir/authorized_keys"
+            fi
+        fi
+    fi
+
+    chmod 700 "$ssh_dir"
+    chmod 600 "$ssh_dir/id_rsa" 2>/dev/null || true
+    chmod 644 "$ssh_dir/id_rsa.pub" 2>/dev/null || true
+    touch "$ssh_dir/authorized_keys"
+    chmod 600 "$ssh_dir/authorized_keys"
+    chown -R "$username:$username" "$ssh_dir"
+
+    local user_home
+    user_home=$(getent passwd "$username" | cut -d: -f6)
+
+    if [[ "$user_home" == "${SHARED_FS_MOUNT}/${username}" ]]; then
+        local local_home="/home/$username"
+        if [[ -d "$local_home" ]] && [[ "$local_home" != "$user_home" ]]; then
+            if [[ -L "$local_home/.ssh" ]]; then
+                echo "[INFO] Symlink $local_home/.ssh already exists"
+            elif [[ -d "$local_home/.ssh" ]]; then
+                rm -rf "$local_home/.ssh"
+                ln -s "$ssh_dir" "$local_home/.ssh"
+                chown -h "$username:$username" "$local_home/.ssh"
+                echo "[INFO] Replaced $local_home/.ssh with symlink to $ssh_dir"
+            else
+                ln -s "$ssh_dir" "$local_home/.ssh"
+                chown -h "$username:$username" "$local_home/.ssh"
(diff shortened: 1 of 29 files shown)


1.architectures/5.sagemaker-hyperpod/Extensions/add-users/setup_ssh_keys.sh
```diff
@@ -0,0 +1,81 @@
...
+    local ssh_dir="${SHARED_FS_MOUNT}/${username}/.ssh"
+
+    if ! id -u "$username" >/dev/null 2>&1; then
+        echo "[WARN] User $username does not exist. Skipping SSH setup."
+        return 0
+    fi
+
+    mkdir -p "$ssh_dir"
+
+    if [[ ! -f "$ssh_dir/id_rsa" ]]; then
```
PremiumSpider, line 31: **Update: RETRACTED my SSH comment above** cross-node SSH works correctly.

My initial test used IP addresses (`ssh -o BatchMode=yes 10.0.78.0 hostname`) which failed with `Host key verification failed`. However, the AMI platform already writes a system-wide SSH config at `/etc/ssh/ssh_config.d/initsmhp-ssh.conf` that disables host key checking for Slurm node **hostnames** (via `Match host * exec "grep '^NodeName=%h' ... slurm.conf"`). When I retested using hostnames (`ssh -o BatchMode=yes ip-10-0-78-0 hostname`), it works perfectly — including in `BatchMode` (non-interactive), which is what MPI/torchrun use.

Since Slurm, `srun`, and distributed training launchers all resolve nodes by hostname (from `sinfo`), not by raw IP, this is not an issue in practice. Sorry for the noise!

## Retrack comment below
I tested cross-node SSH on a fresh cluster and it fails on first use:

```
testuser1@ip-10-0-72-90:~$ ssh 10.0.78.0 hostname
The authenticity of host '10.0.78.0 (10.0.78.0)' can't be established.
ED25519 key fingerprint is SHA256:piRruBl6oAgP779HAOA9m6iPhslMZMfeSZ8F3b9p7IA.
Are you sure you want to continue connecting (yes/no/[fingerprint])?
Host key verification failed.
```

The script generates keypairs and sets up `authorized_keys`, but doesn't write an SSH client config to skip host key verification for cluster-internal hosts. This means:

- Interactive SSH prompts for host key acceptance on every first connection to a new node
- Non-interactive SSH (MPI, `srun`, distributed training launchers like `torchrun`) fails silently because there's no TTY to accept the prompt
- Every time a node is replaced or added via UpdateCluster, the host keys change and SSH breaks again until the user manually clears `known_hosts`

The base lifecycle scripts in this repo handle this by writing an SSH config — for example, `LifecycleScripts/base-config/utils/motd.sh` sets up `StrictHostKeyChecking no` for the ubuntu user.

Would it make sense to add something like this after the keypair generation?

```bash
if [[ ! -f "$ssh_dir/config" ]]; then
    cat > "$ssh_dir/config" << 'SSHEOF'
Host *
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null
SSHEOF
    chmod 600 "$ssh_dir/config"
    chown "$username:$username" "$ssh_dir/config"
fi
```

If `Host *` is too broad, it could be scoped to the VPC CIDR (e.g., `Host 10.*`) or to hostnames matching the cluster pattern.

---

1.architectures/5.sagemaker-hyperpod/Extensions/observability/install_slurm_exporter.sh
```diff
@@ -0,0 +1,63 @@
...
+    echo "This was identified as the controller node because Slurmctld is running. Begining SLURM Exporter Installation"
+
+    SLURM_EXPORTER_DIR="/tmp/slurm_exporter_build"
+
+    if [ -d "$SLURM_EXPORTER_DIR" ]; then
+        echo "Previous build directory found. Removing for clean build..."
+        rm -rf "$SLURM_EXPORTER_DIR"
+    fi
+
+    git clone -b v1.1.0 https://github.com/SckyzO/slurm_exporter.git "$SLURM_EXPORTER_DIR"
```
PremiumSpider, line 35: Question: the slurm exporter is the only component that builds from source during provisioning (git clone + Go compile). The other exporters all pull pre-built Docker images from ECR. Is there a plan to containerize this one too? In VPC-isolated clusters without internet access, this `git clone` would fail.

PremiumSpider (approved):
**Update: RETRACTED my SSH comment above**, cross-node SSH works correctly.

My initial test used IP addresses (`ssh -o BatchMode=yes 10.0.78.0 hostname`) which failed with `Host key verification failed`. However, the AMI platform already writes a system-wide SSH config at `/etc/ssh/ssh_config.d/initsmhp-ssh.conf` that disables host key checking for Slurm node **hostnames** (via `Match host * exec "grep '^NodeName=%h' ... slurm.conf"`). When I retested using hostnames (`ssh -o BatchMode=yes ip-10-0-78-0 hostname`), it works perfectly — including in `BatchMode` (non-interactive), which is what MPI/torchrun use.

Since Slurm, `srun`, and distributed training launchers all resolve nodes by hostname (from `sinfo`), not by raw IP, this is not an issue in practice. Sorry for the noise!


## RETRACTED COMMENT BELOW
Tested this end-to-end on a fresh HyperPod Slurm cluster in us-west-2 (controller + 2 compute + login, ml.t3.medium, FSx OpenZFS at /home, AMI-based Extended mode with OnInitComplete).

Everything works well, node detection is correct on all 4 nodes, users are created with consistent UIDs, home directories land on the shared filesystem, Slurm accounting is set up on the controller, and the observability stack (node-exporter, efa-exporter, slurm-exporter, otel-collector) comes up cleanly with metrics flowing to AMP.

One issue I hit: cross-node SSH fails out of the box for the provisioned users. Details in the inline comment on `setup_ssh_keys.sh`.

---- 

Approving
The SSH config issue is real but not a blocker, it's a missing feature, not a regression. Everything the PR claims to do works correctly. The race condition is low-probability, and the  
  slurm exporter build-from-source is a pre-existing pattern choice.

PR #1070, opened 2026-04-29
nccl-tests: bump to CUDA 13.0.2 / NCCL 2.30.4 and add sm_103 (B300)

KeitaW:
## Summary

Brings the shared `public.ecr.aws/hpc-cloud/nccl-tests` image up to current
upstream and adds NVCC `sm_103` so collectives run natively on Blackwell Ultra
(P6-B300, compute_cap 10.3) instead of falling back to PTX-JIT.

This is the prerequisite for the FSDP test case bump (#1073), which `FROM`s
this image.

## Stack delta

| Component         | Before               | After                 |
|-------------------|----------------------|-----------------------|
| CUDA              | 12.9.1               | **13.0.2**            |
| GDRCopy           | v2.5.1               | v2.5.2                |
| EFA installer     | 1.47.0               | 1.48.0                |
| NCCL              | v2.29.3-1            | **v2.30.4-1**         |
| nccl-tests        | v2.17.9              | v2.18.3               |
| NVCC GENCODE      | sm_80/86/89/90/100   | + **sm_103**          |

The CUDA bump also brings the image above the repo's CI floor
(`MIN_CUDA="13.0"` in `.github/workflows/pr-review-and-slurm-test.yml`); the
prior 12.9.1 was below it.

## Performance

Hardware: SageMaker HyperPod Slurm cluster, 36 × p6-b300.48xlarge
(B300, sm_103). Same allocation for pre and post. Command:
`all_reduce_perf -b 8 -e 16G -f 2 -g 1 -c 1 -n 100` (default in
`nccl-tests-container.sbatch`).

**Avg busbw (geometric mean across all sizes, GB/s):**

| Nodes | GPUs | Pre (`:latest`) | Post (this PR) | Δ          |
|------:|-----:|----------------:|---------------:|-----------:|
| 1     | 8    | 234.86          | 234.08         | -0.3%      |
| 4     | 32   | 155.75          | 175.97         | **+13.0%** |
| 8     | 64   | 133.63          | 158.18         | **+18.4%** |
| 16    | 128  | 120.38          | 144.76         | **+20.3%** |
| 32    | 256  | 112.37          | 136.03         | **+21.1%** |
| 36    | 288  | 112.10          | 131.75         | **+17.5%** |

**Peak-payload busbw (`16 GiB`, OOP, GB/s):**

| Nodes | Pre    | Post   | Δ          |
|------:|-------:|-------:|-----------:|
| 1     | 848.26 | 846.36 | -0.2%      |
[...]

PR #1073, opened 2026-04-29
FSDP: pin nccl-tests base, bump torch to cu130, log TFLOPS/MFU

KeitaW:
## Summary

Three changes:

1. **Pin the nccl-tests base image** instead of `:latest` (which violates the
   repo's "no `:latest`" rule and was a moving target). Pinned to the new tag
   from the companion bump PR #1070. The new image is built with `sm_103` so
   collectives run natively on B300.
2. **Bump torch from 2.7.1 (cu128) to 2.9.1 (cu130)** + matching
   torchaudio/torchvision/transformers. Aligns with the new base image's
   CUDA 13.
3. **Add tokens-per-sec, per-GPU TFLOPS, and MFU** to the existing
   `samples/sec` log line in `src/train.py`. Adds `--peak_tflops_per_gpu`
   (default 3000 for B300; H100/B200 numbers in the help text) so MFU is
   computed correctly without users doing it externally.

Drops the `LD_PRELOAD=/usr/local/cuda-12.8/lib/libnccl.so` block from all 10
sbatch scripts and `training-sub.template` — the new base image already
exports NCCL via `LD_LIBRARY_PATH`, and the hardcoded DLAMI path was a known
footgun.

## Stack delta

| Component         | Before                | After                 |
|-------------------|-----------------------|-----------------------|
| Base image tag    | `:latest` ⚠️           | `cuda13.0.2-efa1.48.0-ofiv1.19.0-ncclv2.30.4-1-testsv2.18.3` |
| torch             | 2.7.1 (cu128)         | **2.9.1 (cu130)**     |
| torchaudio        | 2.7.1                 | 2.9.1                 |
| torchvision       | 0.22.1                | 0.24.1                |
| transformers      | 4.53.0                | 4.57.6                |

## Dependency

This PR can merge only after the new `public.ecr.aws/hpc-cloud/nccl-tests`
tag from PR #1070 is published. Until then, container builds for this PR
will fail with "image not found".

## Performance

Hardware: SageMaker HyperPod Slurm cluster, p6-b300.48xlarge (B300, sm_103),
8 GPUs/node. Llama 3.1 8B, FSDP2 full-shard, BF16, batch_size=4 per rank,
seq_len=8192, sm_103 NCCL collectives via the new base image.

| Nodes | GPUs | Tokens/s/GPU | TFLOPS/GPU | True MFU |
|------:|-----:|-------------:|-----------:|---------:|
| 1     | 8    | 6,609        | 293.1      | 9.8%     |
| 4     | 32   | 6,681        | 296.4      | 9.9%     |
| 8     | 64   | 6,672        | 295.9      | 9.9%     |
| 16    | 128  | 6,617        | 293.5      | 9.8%     |
| 32    | 256  | 6,325        | 280.5      | 9.4%     |

(36-node row queued; will append once it lands.)

True MFU = `TFLOPS_per_GPU / 3000` (B300 dense BF16 peak, 3 PFLOPS/GPU).
The new `--peak_tflops_per_gpu` flag drives the MFU column directly in the
log line — no external recomputation needed. **Scaling is essentially flat
through 32 nodes** (293 → 296 → 296 → 294 → 281 TFLOPS/GPU = -4% drift at
the largest scale), which is the kind of plateau the new sm_103 NCCL
collectives were supposed to hit.

The MFU on this workload is low (~10%) because Llama 3.1 8B at
batch_size=4/seq_len=8192 isn't compute-bound on B300 — the 6,500 tps/GPU
floor here is bound by per-step optimizer + FSDP all-gather overhead at
batch_size=4. Larger models (70B, MoE) and FP8 paths will push closer to
peak; they're future work for a follow-up PR. The point of *this* sweep is
to confirm the bumped stack scales — it does.

### Methodology note (bench-only, NOT in shipped sbatches)

For these runs, the dataset env was overridden to use a pre-cached
HuggingFace dataset (`wikitext-103-raw-v1`) with `HF_DATASETS_OFFLINE=1` —
multi-rank streaming of `c4` from compute nodes hit HF Hub rate limits (HTTP
429) under 256+ concurrent ranks. **The shipped sbatch scripts in this PR
still default to `c4`** so single-node and small-scale users get the
out-of-the-box experience. The wikitext workaround is a benchmarking
expedient for high-rank runs, not a recommended config — that's why it isn't
pinned into the source sbatches. (For future high-rank-count benches, the
right fix is a one-time dataset prefetch into the FSx mount, not a streaming
override.)

## Test plan

- [ ] PR #1070 image published.
- [x] `docker build` succeeds against the new base.
- [x] `enroot import` succeeds.
- [x] `python -c "import torch; print(torch.cuda.is_available(), torch.version.cuda)"`
      → `True 13.0`.
- [x] `llama3_1_8b-training.sbatch` smoke run, confirm new TFLOPS/MFU log
      line appears.
- [x] Llama 3.1 8B sweeps at 1/4/8/16/32 nodes (above).
- [x] 32-node Llama 3.1 8B sweep (above) is the cap; user limited to ≤32n.
- [ ] 70B + Mistral 8x7B sweeps (future PR — see note above).


allela-roy (approved):
LGTM

PR #1082, opened 2026-05-05
chore: Making observability setup false by default

kn0wthing:
## Purpose

* Observability is advanced feature where customer needs to do some prerequisites.
* This will be consistent with old LCS from Base Config LifeCycle scripts

## Changes

Making observability setup false by default

-

## Test Plan

Create cluster with same LCS by putting this change in S3 bucket and using it.

## Test Results

Screenshot 2026-05-04 at 5 19 59 PM

Screenshot 2026-05-04 at 5 23 54 PM

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/Extensions/run_extensions.sh
+++ b/1.architectures/5.sagemaker-hyperpod/Extensions/run_extensions.sh
@@ -28,7 +28,7 @@ set -ex
 # Feature toggles -- set to "true" or "false"
 # ===========================================================================
 ENABLE_ADD_USERS="true"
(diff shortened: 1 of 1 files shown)


svnisar:
Testing cluster creation without observability flag true results successful cluster creation

svnisar (approved):

PremiumSpider (approved):
User needs to follow readme and supply users.txt or users.yaml as ENABLE_ADD_USERS="true" however for ENABLE_OBSERVABILITY="true" users need to know they need to setup AMP and setup IAM role to have ECR permission AmazonEC2ContainerRegistryReadOnly

PR #1100, opened 2026-05-16
fix(parallelcluster-prereqs): default FSx OpenZFS to SINGLE_AZ_HA_2

### LINKED ISSUE #1097: [Bug]: Prereqs template hardcodes FSx OpenZFS `SINGLE_AZ_HA_1`, which is unavailable in us-east-1 and most major regions
## Environment

- AWS Service: ParallelCluster
- Template: `1.architectures/2.aws-parallelcluster/infra-templates/parallelcluster-prerequisites.yaml`
- Region/AZs tested: us-east-1a (use1-az6), us-east-1b (use1-az1)
- Date observed: 2026-05-16

## Bug Description

**Expected behavior:** The prereqs stack deploys cleanly with default parameters in any AZ that supports the requested compute instance types.

**Actual behavior:** Stack creation fails inside the `OpenZFSFileSystem` resource with:

```
The file system creation request failed because OpenZFS file systems with the
provided deployment type and storage type are not available in the requested
Availability Zone.
```

`parallelcluster-prerequisites.yaml` L323 hard-codes:

```yaml
DeploymentType: SINGLE_AZ_HA_1
```

Per the AWS [Availability by AWS Region](https://docs.aws.amazon.com/fsx/latest/OpenZFSGuide/available-aws-regions.html) table for FSx for OpenZFS, `SINGLE_AZ_HA_1` is **not offered at all** in many major regions including:

- US East (N. Virginia) — `us-east-1`
- US East (Ohio) — `us-east-2`
- US West (Oregon) — `us-west-2`
- Europe (Frankfurt), (Ireland), (Spain), (Zurich)
- Asia Pacific (Tokyo), (Singapore), (Sydney), (Mumbai)
- Canada (Central)

In all of these regions, `SINGLE_AZ_HA_2` is the only Single-AZ (HA) SSD option. The official table footnote notes that "*Due to differences in infrastructure capabilities and configurations, your file system type may be unavailable in specific Availability Zones within these regions.*" — so even in regions where HA_1 *is* offered, AZ-level availability may be partial.

The AWS [`CreateFileSystemOpenZFSConfiguration` API reference](https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemOpenZFSConfiguration.html) describes the two options as:

> `SINGLE_AZ_HA_2` — Creates file systems with high availability and throughput capacities of 160 - 10,240 MB/s using an NVMe L2ARC cache by deploying a primary and standby file system within the same Availability Zone.
>
> `SINGLE_AZ_HA_1` — Creates file systems with high availability and throughput capacities of 64 - 4,096 MB/s by deploying a primary and standby file system within the same Availability Zone.

We empirically reproduced the regional-unavailability case during a `us-east-1` deploy: the AZ-unavailable error appeared in `us-east-1a` (AZ ID `use1-az6`) and `us-east-1b` (AZ ID `use1-az1`); switching to HA_2 succeeded.

## Steps to Reproduce

```bash
aws cloudformation create-stack \
  --stack-name fsx-ha1-repro \
  --template-body file://1.architectures/2.aws-parallelcluster/infra-templates/parallelcluster-prerequisites.yaml \
  --parameters \
    ParameterKey=SubnetsAZ,ParameterValue=us-east-1a \
  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM
```

Wait for `CREATE_FAILED`, then:

```bash
aws cloudformation describe-stack-events --stack-name fsx-ha1-repro \
  --query 'StackEvents[?ResourceStatus==`CREATE_FAILED` && ResourceType==`AWS::FSx::FileSystem`]'
```

## Suggested Fix

Two options:

**Option A — flip the default (smallest change):**

```diff
-        DeploymentType: SINGLE_AZ_HA_1
+        DeploymentType: SINGLE_AZ_HA_2
```

Also bump the `HomeThroughput` description / default to reflect HA_2's minimum (160 MB/s — note the existing 512 default is already valid for both HA_1 and HA_2 so the default value doesn't need to change).

**Option B — expose the deployment type as a parameter (more flexible):**

```yaml
Parameters:
  HomeDeploymentType:
    Description: >-
      FSx OpenZFS deployment type for /home. SINGLE_AZ_HA_2 is recommended
      (Graviton-based, broader AZ availability). SINGLE_AZ_HA_1 is retained
      for backward compatibility but is unavailable in many AZs.
    Type: String
    Default: SINGLE_AZ_HA_2
    AllowedValues:
      - SINGLE_AZ_HA_1
      - SINGLE_AZ_HA_2
```

```yaml
OpenZFSFileSystem:
  Properties:
    OpenZFSConfiguration:
      DeploymentType: !Ref HomeDeploymentType
```

Either change is `CREATE_ONLY` (FSx doesn't allow changing deployment type on an existing file system), so no migration story is needed for fresh deploys.

Related observation worth tracking separately: `HomeThroughput` is declared as a plain `Type: Number` without `AllowedValues`. FSx only accepts a discrete set of throughput values that differs by deployment type (HA_1: 64/128/256/512/…; HA_2: 160/320/640/1280/…). A user overriding to an arbitrary value fails late inside FSx. Worth adding `AllowedValues` after this change lands.

Happy to submit a PR.

KeitaW:
Fixes #1097.

`1.architectures/2.aws-parallelcluster/infra-templates/parallelcluster-prerequisites.yaml` hard-coded `DeploymentType: SINGLE_AZ_HA_1` for the FSx OpenZFS file system backing `/home`. Per the official [FSx OpenZFS Availability by AWS Region table](https://docs.aws.amazon.com/fsx/latest/OpenZFSGuide/available-aws-regions.html), `SINGLE_AZ_HA_1` is not offered in us-east-1, us-east-2, us-west-2, eu-central-1, eu-west-1, ap-northeast-1, ap-southeast-1, ap-southeast-2, ap-south-1, ca-central-1, and several other major regions — so the template fails out-of-the-box there.

This flips the default to `SINGLE_AZ_HA_2`, which is the newer generation and broadly available in all the regions where HA_1 is missing. Its 160 MB/s throughput minimum is below the template's existing `HomeThroughput` default of 512 MB/s, so no other parameters need to change. Minimal one-line fix; introducing a configurable `DeploymentType` parameter / Mappings / Rules was considered and explicitly deferred as a separate future enhancement.

### DIFF
--- a/1.architectures/2.aws-parallelcluster/infra-templates/parallelcluster-prerequisites.yaml
+++ b/1.architectures/2.aws-parallelcluster/infra-templates/parallelcluster-prerequisites.yaml
@@ -320,7 +320,7 @@ Resources:
         CopyTagsToBackups: Yes
         CopyTagsToVolumes: Yes
         DailyAutomaticBackupStartTime: '19:00'
-        DeploymentType: SINGLE_AZ_HA_1
+        DeploymentType: SINGLE_AZ_HA_2
         Options: 
           - DELETE_CHILD_VOLUMES_AND_SNAPSHOTS
         RootVolumeConfiguration: 


KeitaW:
Follow-up: opened #1102 to extend this fix all the way to universal regional coverage. `SINGLE_AZ_HA_2` (this PR) covers the major regions where `HA_1` was rejected, but still isn't offered in ~20 smaller regions (ap-osaka, ca-west-1, eu-london, sa-east-1, af-south-1, me-bahrain, GovCloud-East, …). #1102 adds a secondary private subnet and switches OpenZFS to `MULTI_AZ_1` — the only deployment type with a checkmark in every row of the [availability table](https://docs.aws.amazon.com/fsx/latest/OpenZFSGuide/available-aws-regions.html). It also incidentally removes the manual second-subnet workaround that today's accounting-database CFN template forces on users.

PR #1108, opened 2026-05-23
feat: add OpenVLA-OFT LIBERO finetuning test case on EKS

sagarneeldubey:
## Purpose

Add a new test case for fine-tuning [OpenVLA](https://huggingface.co/openvla/openvla-7b) using the [OFT (Optimized Fine-Tuning)](https://github.com/moojink/openvla-oft) recipe on AWS-backed Kubernetes (EKS / SageMaker HyperPod-on-EKS), with an end-to-end LIBERO walkthrough.

## Changes

- New test case under `3.test_cases/pytorch/vla/openvla/openvla-oft/`:
  - `Dockerfile` building an image with `openvla-oft` + the vendored `finetune.py` script.
  - `src/finetune.py` and `src/requirements.txt` pinning extra Python deps layered on top of `openvla-oft`.
  - `kubernetes/libero/` recipe with a one-shot dataset download Job, a `PyTorchJob` manifest (rendered via `envsubst`), optional dynamic and static FSx for Lustre PVC manifests, and a `verify-tfds-layout.sh` sanity check.
  - `tests/libero/` smoke + property tests, manifest renderer, and shellcheck script for local validation.
  - Top-level `README.md` and per-recipe `kubernetes/libero/README.md` covering prerequisites, build → ECR push → PyTorchJob submission, and known issues.
- `.gitignore`: ignore `.kiro/` IDE directory.

## Test Plan

**Environment:**
- AWS Service: SageMaker HyperPod-on-EKS
- Instance type: `p4d.24xlarge` (8x A100 40 GB) — used for validation with `BATCH_SIZE=2` to fit in memory. `p4de.24xlarge` (A100 80 GB) or above is recommended for the upstream-default batch size.
- Number of nodes: 1 (8 GPUs, single-node multi-GPU fine-tuning per upstream LIBERO recipe)

**Test commands:**

```bash
# Render and lint manifests locally
cd 3.test_cases/pytorch/vla/openvla/openvla-oft/tests/libero
./shellcheck.sh
./render_all_manifests.sh
pytest -q

# On-cluster
kubectl apply -f kubernetes/libero/pvc-fsx-lustre-dynamic.yaml  #not needed if fsx-cluster already provisioned
kubectl apply -f kubernetes/libero/libero-download.yaml
envsubst < kubernetes/libero/libero-finetune.yaml | kubectl apply -f -
```

## Test Results

- Local: `pytest`, `shellcheck`, and the manifest renderer all pass. The TFDS layout verifier confirms the staged `openvla/modified_libero_rlds` dataset.
- On-cluster: dataset download Job completes and stages RLDS data onto the shared PVC; `PyTorchJob` starts training with the upstream LIBERO hyperparameters (with `BATCH_SIZE=2` to fit p4d's 40 GB A100s) and writes checkpoints + logs to `/data/runs`. Ran for `Steps=100`. Screenshot of training logs below:

```bash
Saved merged model for Step 100 at: /data/runs/openvla-7b+libero_spatial_no_noops+b8+lr-0.0005+lora-r32+dropout-0.0--image_aug--parallel_dec--8_acts_chunk--continuous_acts--L1_regression--3rd_person_img--wrist_img--proprio_state--100_chkpt
/usr/local/lib/python3.12/site-packages/torch/distributed/distributed_c10d.py:4807: UserWarning: No device id is provided via `init_process_group` or `barrier `. Using the current device set by the user. 
  warnings.warn(  # warn only once
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
Max step 100 reached! Stopping training...
```

## Directory Structure

```
3.test_cases/
└── pytorch/
    └── vla/
        └── openvla/
            └── openvla-oft/
                ├── Dockerfile
                ├── README.md
                ├── src/
                │   ├── finetune.py
                │   └── requirements.txt
                ├── kubernetes/
                │   └── libero/
                │       ├── README.md
                │       ├── env_vars.example
                │       ├── libero-download.yaml
                │       ├── libero-finetune.yaml
                │       ├── pv-fsx-lustre-static.yaml
                │       ├── pvc-fsx-lustre-dynamic.yaml
                │       └── verify-tfds-layout.sh
                └── tests/
                    └── libero/
                        ├── conftest.py
                        ├── fixtures/
                        ├── render_all_manifests.sh
                        ├── shellcheck.sh
                        ├── smoke-verify-tfds-layout.sh
                        ├── test_p8_rlds_dataset_failfast.py
                        └── test_properties.py
```

A `slurm/` sibling folder is intentionally left out for now and will be added in a follow-up once tested end-to-end.

## Checklist


KeitaW (changes requested):
Few comments

KeitaW (commented):

---

This PR adds a substantial, well-documented new test case for fine-tuning OpenVLA-OFT
on LIBERO over AWS-backed Kubernetes (EKS / HyperPod-on-EKS): a DLC-based Dockerfile,
a vendored `finetune.py`, a complete `kubernetes/libero/` recipe (PyTorchJob + dataset
Download Job + FSx PVC variants), and an unusually thorough README with a real
troubleshooting catalogue. The engineering quality is high — the failure modes
documented in the README are exactly the ones you'd otherwise rediscover the hard way.
My main concern is structural: the test case lands at a non-standard nesting depth
(`pytorch/vla/openvla/openvla-oft/`) that has no precedent in the repo and, as a direct
consequence, breaks three relative documentation links. I also reviewed the vendored
`src/finetune.py` training logic in depth: the algorithm is faithful to upstream and
correct for this recipe, but it surfaces a genuine **correctness blocker** — a dependency
interaction that can silently train the wrong model (see *Training Logic & Implementation*).
The rest are mostly hygiene items (license headers) plus a few operational and doc nits — and
I'd recommend dropping the bundled `tests/libero/` suite outright, since it exercises bash
idioms and third-party tools rather than whether the recipe builds and trains.

---

### Directory placement adds a `vla/` grouping level the repo doesn't use

**File:** `3.test_cases/pytorch/vla/openvla/openvla-oft/` (new tree)

I noticed this lands at `3.test_cases/pytorch/vla/openvla/openvla-oft/` — a four-level
nesting under `pytorch/`. Every existing sibling under `3.test_cases/pytorch/` is flat
(`FSDP`, `ddp`, `deepspeed`, `nanoVLM`, `torchtitan`, `verl`, `vjepa2`, `vjepa2.1`, …),
including the vision models, and there is no existing `vla/` directory. So this PR
introduces a brand-new intermediate grouping layer that nothing else follows. I'd
suggest dropping the `vla/` level and placing this at
`3.test_cases/pytorch/openvla/openvla-oft/`, which matches how `nanoVLM` and the
`vjepa2*` cases sit directly under `pytorch/`. Beyond consistency, the extra depth is
what breaks the relative links below — flattening one level makes those paths shorter
and easier to keep correct. Could you confirm whether the `vla/` layer is load-bearing
for anything, or whether it can come out?

---

### The nesting depth breaks three relative documentation links

**Files:** `README.md` (line 31), `kubernetes/libero/README.md` (lines 36, 906)

Resolving the relative links from their actual locations, three of them point one (or
more) levels too shallow — the targets all exist, but the link depth is wrong, and the
miscount comes straight from the extra `vla/openvla` nesting:

- `README.md:31` → `../../../../1.architectures/4.amazon-eks` resolves to
  `3.test_cases/1.architectures/4.amazon-eks` (doesn't exist). From this file the repo
  root is five levels up, so it needs `../../../../../1.architectures/4.amazon-eks`.
- `kubernetes/libero/README.md:36` → `../../../../../../1.architectures` resolves to
  `3.test_cases/1.architectures`. From `kubernetes/libero/` the root is seven levels up:
  `../../../../../../../1.architectures`.
- `kubernetes/libero/README.md:906` → `../../../nvrx/Dockerfile` resolves to
  `3.test_cases/pytorch/vla/openvla/nvrx/Dockerfile`. `nvrx` lives at
  `3.test_cases/pytorch/nvrx/`, so from `kubernetes/libero/` it needs
  `../../../../../nvrx/Dockerfile`.

It might be worth fixing these after settling the directory question above, since
flattening the tree changes the correct depth — recompute them once the final location
is decided.

---

### MIT-0 license header is missing on every new file

**Files:** all 18 new files in the test case

None of the added files carry the repo's required header. This is the single most
common thing that slips through on a new test-case scaffold, so I'd batch-add it to
every shell script, YAML manifest, and Python file (for `src/finetune.py`, which is
vendored from upstream, the header goes above the upstream docstring — the same way other
vendored scripts in the repo handle it).

```
```

(The `.gitignore` change is fine — it's an append to an existing file, not a new one.)

KeitaW (commented):
## Review Batch 2/5 — Deployment Pipeline

### `envsubst` is invoked without an explicit allow-list in the default path

**File:** `kubernetes/libero/README.md`, `kubernetes/libero/*.yaml`

The manifests are carefully written against the `$$`-escape convention and the README
documents the GNU-`envsubst` allow-list form as a fallback — that's genuinely good
defensive work. That said, the default commands throughout the walkthrough are bare
`envsubst < file | kubectl apply -f -`, which substitutes every `$NAME` in the
environment and only behaves because of the `$$` escaping. I'd suggest promoting the
allow-list form (`envsubst '$IMAGE_URI $DATA_PVC_NAME $NAMESPACE $TASK_SUITE $TASK_SUITE_DNS' < …`)
to the *primary* instruction rather than the fallback, so the recipe is robust
regardless of which `envsubst` is first on a contributor's PATH.

---

### The bundled `tests/libero/` suite tests the wrong layer — I'd drop it from this PR

**Files:** the whole `tests/libero/` directory — **~1,650 lines of test code across 8 files**
for a single training example.

I went through each file. The pattern is consistent: the suite validates the cheap,
self-evident layer (bash idioms, static YAML, an obvious helper, third-party tools) and never
touches the one that matters for a training recipe — *does the image build and train?* File by
file:

| File | Lines | What it does | Verdict |
|------|------:|--------------|---------|
| `test_properties.py` | 677 | Property tests P1–P7 (breakdown below) | **Drop** |
| `conftest.py` | 317 | Fixtures: `envsubst` path resolution, render + YAML-parse helpers, well-formed / malformed TFDS-tree factories | **Drop** — exists only to serve the property tests |
| `test_p8_rlds_dataset_failfast.py` | 179 | P8: `RLDSDataset` on an empty/malformed dir raises quickly and names the missing path | **Drop** — exercises upstream's loader, not the contributor's code |
| `fixtures/hf-cli-shim` | 153 | Fake `huggingface-cli download` that idempotently creates the TFDS layout | **Drop** — exists only for the tautological P6 |
| `render_all_manifests.sh` | 118 | Renders every manifest via `envsubst`, validates via `kubectl apply --dry-run=client` or a PyYAML fallback | Closest to useful (envsubst is the fragile, bespoke piece), but subsumed by an integration test |
| `smoke-verify-tfds-layout.sh` | 105 | Bash smoke test of `verify-tfds-layout.sh` (happy path + two malformed cases) | **Drop** — tests an obvious 33-line helper |
| `fixtures/env_vars.test` | 68 | Test-only env-var values so manifests render under `--dry-run=client` (cleanly made — AWS-ID-shaped placeholders, no `:?`, no secrets) | **Drop** — a well-built fixture, but only feeds tests that are themselves going |
| `shellcheck.sh` | 37 | `shellcheck`s the `env_vars` file, skips if shellcheck absent | **Drop** — lints one file; CI already runs `flake8`/`bandit`/`bash -n` |

**Property-by-property (`test_properties.py` + `test_p8`):**
- **P1** — allow-list accept/reject: ~5 lines of standard bash (`grep -qxF`); doesn't stop the allow-list from diverging across the three places it's duplicated.
- **P2** — fail-loud on unset `TASK_SUITE`: asserts `${VAR:?}` aborts — i.e., tests a *bash language feature*.
- **P3** — distinct suites → distinct manifest names: asserts `${TASK_SUITE//_/-}` differs per suite; a one-line property.
- **P4 / P5** — `verify-tfds-layout.sh` accepts/rejects trees: tests the contributor's own obvious 33-line helper.
- **P6** — Download-Job idempotence: stubs `huggingface-cli` *and* `pip` with fakes, runs the wrapper twice, asserts a byte-identical tree. The stub is idempotent by construction → tautology; and it excludes every real download risk (partial write, auth, rate-limit, network).
- **P7** — render → `kubectl dry-run` → YAML preserves the training flags: those flags are *literal text* `envsubst` never substitutes, so this is largely a static-text snapshot.
- **P8** — `RLDSDataset` fails fast on an empty PVC: real operator value in the "actionable message" angle, but it's testing upstream's loader, not this PR's code.

Nothing across the 1,650 lines builds the image or runs a training step — exactly what
`max_steps=100` was sized for — and the whole approach diverges from the repo's
`docker_build` / `docker_run` integration-test convention.

I'd suggest **dropping the entire `tests/libero/` directory from this PR.** If you want
coverage, add it as a focused follow-up in the repo's integration style: a `docker_build` + a
short `docker_run` that builds the image and runs `max_steps=100` to a first loss line /
checkpoint write. That one test is worth more than all eight properties — and because it renders
and applies the manifests as part of running, it also subsumes the only check here with genuine
value (that the `$$`-escaped `envsubst` still produces valid YAML), so nothing real is lost by
removing the unit suite now.

KeitaW (commented):
## Review Batch 3/5 — Infrastructure & NCCL Configuration

3.test_cases/pytorch/vla/openvla/openvla-oft/src/requirements.txt
```diff
@@ -0,0 +1,12 @@
+# Python deps for OpenVLA-OFT finetune.py
+# The `prismatic` / `experiments` packages are installed from the upstream
+# openvla-oft repo via `pip install -e .` in the Dockerfile, which pulls
+# most of the heavy ML deps. Anything listed below is an explicit top-up.
+
+draccus==0.8.0
+accelerate>=0.30.0
+peft>=0.11.0
+huggingface-hub>=0.23.0
+transformers>=4.40.1
```
KeitaW: ### `requirements.txt` re-declares upstream deps and may shadow the required `transformers` fork

**File:** `src/requirements.txt` (line 10)

Checking upstream `openvla-oft`'s `pyproject.toml` at the pinned commit, it already
declares `accelerate`, `peft==0.11.1`, `huggingface_hub`, `draccus==0.8.0`, `wandb`,
and — importantly — `transformers @ git+https://github.com/moojink/transformers-openvla-oft.git`,
a fork required for the bidirectional attention that OFT parallel decoding depends on.
Because the Dockerfile runs `pip install -e /opt/openvla-oft` first and *then*
`pip install -r requirements.txt`, the plain `transformers>=4.40.1` line here can pull
stock PyPI `transformers` over the fork, silently disabling parallel decoding — the model
then trains under causal attention with no error, producing a structurally wrong policy.
This is the **correctness blocker** detailed under *Training Logic & Implementation* below. The
other lines are redundant (the "explicit top-up" comment is misleading — almost
everything listed is already pinned upstream) and the lower-bound-only / unpinned
entries (`tqdm`, `wandb`) can drift on rebuild. I'd suggest trimming this file to only
genuinely-additional packages and dropping `transformers` entirely so the fork stays
installed; if a top-up genuinely needs a floor, pair it with an upper bound.

3.test_cases/pytorch/vla/openvla/openvla-oft/kubernetes/libero/libero-finetune.yaml
```diff
@@ -0,0 +1,216 @@
...
+  namespace: ${NAMESPACE}
+  labels:
+    app.kubernetes.io/part-of: openvla-oft-libero
+    app.kubernetes.io/component: finetune
+spec:
+  elasticPolicy:
+    rdzvBackend: c10d
+    minReplicas: 1
+    maxReplicas: 64
+    maxRestarts: 100
```
KeitaW: ### `elasticPolicy` restart/replica ceilings look like template defaults for a single-node recipe

**File:** `kubernetes/libero/libero-finetune.yaml` (line 41)

The recipe is single-node (`NUM_NODES=1`, all ranks in one pod), but `elasticPolicy`
sets `maxReplicas: 64` and `maxRestarts: 100`. A restart ceiling of 100 on a GPU job
can burn a lot of GPU-hours retrying a deterministic failure (bad dataset, OOM) before
anyone notices. I'd lower `maxRestarts` to something conservative and bring
`maxReplicas` in line with the recipe's actual scale.

```suggestion
    maxRestarts: 3
```

3.test_cases/pytorch/openvla-oft/src/finetune.py
```diff
@@ -0,0 +1,876 @@
...
+            torch.save(action_head.state_dict(), checkpoint_dir / f"action_head--{checkpoint_name_suffix}")
+
+        if cfg.use_film:
+            torch.save(
+                vla.module.vision_backbone.state_dict(), checkpoint_dir / f"vision_backbone--{checkpoint_name_suffix}"
+            )
+
+    dist.barrier()
+
+    if cfg.use_lora and cfg.merge_lora_during_training:
```
KeitaW, line 502: ### Every rank reloads and merges the 7B at each checkpoint

**File:** `src/finetune.py` (lines 476–481)

This is **inherited verbatim from upstream `moojink/openvla-oft`** — not introduced by this
PR (the only edits the vendored copy makes are the two `main_process_first()` wrappers on the
model-*load* calls; this merge block is unchanged). Flagging it as an upstream quirk worth
fixing here, not as a contributor mistake. The `merge_lora_during_training` block runs
`AutoModelForVision2Seq.from_pretrained(...)` + `PeftModel.from_pretrained(...)` +
`merge_and_unload()` on **every** rank, but only rank 0 writes the result; the `dist.barrier()`
above already guarantees the adapter is on disk first. That is 7× redundant 7B loads + merges
per checkpoint — slow, with extra shared-filesystem reads (not an OOM on the 80 GB / ~1–2
TB-RAM target nodes). I'd move the load/merge inside `if distributed_state.is_main_process:`,
keeping the trailing `dist.barrier()` reachable by all ranks so there's no deadlock — and it's
worth sending the same fix upstream.

KeitaW (commented):
## Review Batch 4/5 — Training Logic & Implementation

---

`src/finetune.py` is vendored from upstream `moojink/openvla-oft` (`vla-scripts/finetune.py` @ `e4287e9`). Its training algorithm is faithful to upstream and
correct for this recipe — DDP + LoRA with a continuous L1-regression action head, so the
usual ZeRO-3 traps (`device_map`, sharded `save_pretrained`) don't apply, and the
continuous-head forward path correctly extracts the action-token hidden states and computes
L1 over the full `(B, 8, 7)` chunk. The issues below live in the dependency and config
context the script runs in, not in its algorithm.






---

---

### The `transformers`-fork shadowing is the one training-correctness blocker

This is the training-time face of the `requirements.txt` finding above. At
`finetune.py:626` the model loads via
`AutoModelForVision2Seq.from_pretrained(..., trust_remote_code=True)`; OFT parallel decoding
only works if the `moojink/transformers-openvla-oft` fork (which makes `LlamaModel.forward`
attention bidirectional) is the installed `transformers`. If the stock wheel wins the
install race, the model still loads and trains — under **causal** attention — with no error
and no NaN: the loss falls but the policy is structurally wrong and incompatible with OFT
inference. Fix it at the source (drop the `transformers` line from `requirements.txt`); a
startup guard in `finetune.py` is reasonable defense-in-depth, but tie it to a *verified*
fork marker — the fork changes `forward()` logic, not a persistent attribute, so a naive
assertion can pass on a broken install.

---

### `merge_lora_during_training` writes ~150 GB of throwaway merged checkpoints in the smoke run

**Files:** `kubernetes/libero/libero-finetune.yaml` (the `command:` save flags),
`src/finetune.py` (`save_training_checkpoint`, lines 436–487; save trigger at line 839)

Yes — I traced this through the code. The chain, step by step:

**The four flags interact like this:**

| Flag (from the manifest) | Value | Effect |
|---|---|---|
| `--max_steps` | 100 | training stops at step 100 |
| `--save_freq` | 10 | a checkpoint is written every 10 steps |
| `--save_latest_checkpoint_only` | False | each checkpoint goes to its **own** dir (`…--<step>_chkpt/`); nothing is overwritten |
| `--merge_lora_during_training` | True | each checkpoint **also** merges LoRA into the base and writes the **full 7B model** |

**How many checkpoints?** The loop saves when `gradient_step_idx > 0 and log_step % save_freq == 0`
(`finetune.py:839`), and the `max_steps` break (`:870`) runs *after* the save. With
`grad_accumulation_steps=1`, `log_step` advances by 1 per step, so saves fire at steps
**10, 20, 30, …, 100 — ten of them** (step 100 saves, then breaks).

**Why each one is ~15 GB, not a few hundred MB.** This is the crux. A LoRA checkpoint is
normally tiny — just the rank-32 adapter (a few hundred MB) plus the small action/proprio
heads. But `--merge_lora_during_training=True` makes `save_training_checkpoint` reload the base
model, call `merge_and_unload()`, then `merged_vla.save_pretrained(checkpoint_dir)` — writing
the **entire merged OpenVLA-7B** (~15 GB in bf16) into that step's directory. The merge inflates
each checkpoint roughly **50×**, and `save_latest_checkpoint_only=False` keeps all ten.

**The tally on the PVC.** Ten merged models × ~15 GB ≈ **150 GB of checkpoints**, sharing the
volume with everything else the run writes to `/data`:
- `/data/hf-cache` — the base `openvla-7b` snapshot the run downloads (~15 GB),
- `/data/datasets/rlds` — the suite the Download Job staged (~2.5 GB for one LIBERO suite),
- `/data/runs/<run_id>` — these ten checkpoints.

≈ **165–170 GB total.** The README recommends a PVC of **"at least 100 GiB"** (and notes
checkpoints colocate with the dataset on the same volume), so on a minimum-spec PVC the run hits
**`ENOSPC` partway through** — and it dies *late*, after burning most of the smoke test's
wall-clock, the worst time to fail. (It does fit the `DATA_PVC_SIZE=1200Gi` *default* in
`env_vars.example`, so the inconsistency is specifically the README's 100 GiB minimum vs. what
the shipped flags actually write.)

**Why it's all wasted:** this is a 100-step smoke test — ~0.07 % of the ~150k steps upstream
uses for a real run — so the ten merged 7B checkpoints are throwaway artifacts of an
unconverged model. The test only needs to prove the save/merge path works *once*.

**Fix (any one):** `--save_latest_checkpoint_only=True` (one dir, overwritten — ~15 GB instead
of ~150 GB), or `--save_freq` ≥ `max_steps` (a single save at the end), or drop
`--merge_lora_during_training` for the smoke run (keep only the small adapter). Separately, the
README's 100 GiB minimum should be reconciled with what the shipped flags actually write.

---

### `src/finetune.py` is modified, not "vendored verbatim" — and shouldn't be vendored at all

`src/finetune.py`'s header states it is *"vendored verbatim from the openvla-oft
repository."* It is not. Stripping docstrings and comments from both files and diffing the
pure code:

- Raw line counts: upstream **1142** → vendored **876** (−266 lines).
- After removing docstrings/comments: upstream **383** vs vendored **385** code lines.
- **The entire code delta is two edits — both the same change:** wrapping the model-load
  calls in `main_process_first()` (`main_process_first` appears 3× in the vendored file, 0×
  upstream).

**1. `snapshot_download` race fix** (`finetune.py:593–596`):
```python
# upstream:                                  # vendored:
if model_is_on_hf_hub(cfg.vla_path):         if model_is_on_hf_hub(cfg.vla_path):
    vla_download_path = snapshot_download(       with distributed_state.main_process_first():
        repo_id=cfg.vla_path)                        vla_download_path = snapshot_download(repo_id=cfg.vla_path)
```

**2. `from_pretrained` race fix** (`finetune.py:624–631`):
```python
# upstream:                                  # vendored:
processor = AutoProcessor.from_pretrained(   with distributed_state.main_process_first():
    cfg.vla_path, trust_remote_code=True)        processor = AutoProcessor.from_pretrained(cfg.vla_path, trust_remote_code=True)
vla = AutoModelForVision2Seq.from_pretrained(    vla = AutoModelForVision2Seq.from_pretrained(
    cfg.vla_path, ...).to(device_id)                 cfg.vla_path, ...).to(device_id)
```

These fix the two distributed races the README troubleshooting documents (the 8× HF-Hub-hit
/ `update_auto_map` race, and the `trust_remote_code` dynamic-modules-cache partial read).
Everything else in the 266-line delta is docstrings/comments — including the module
docstring, which was rewritten to *add* the self-refuting "vendored verbatim" sentence that
isn't in upstream at all.

**Why it matters:** "verbatim" invites a future maintainer to re-vendor from upstream to
"resync," which would silently drop both race fixes and reintroduce the exact crashes the
README documents.

**Recommended path — don't carry a patched `finetune.py` in this repo at all.** The two
`main_process_first()` fixes are general-purpose (any multi-rank `openvla-oft` user hits these
races), so their home is upstream, not a vendored fork that won't track future upstream
changes and will silently skew against the pinned package (the script and the
`pip install -e`'d package must move in lockstep). Concretely:

1. **Open a PR against `moojink/openvla-oft`** with the two race fixes (and, while there, the
   `merge_lora_during_training` all-rank-merge fix above — same root, same beneficiaries).
2. **Once merged, drop the vendored `src/finetune.py` entirely:** bump `OPENVLA_OFT_COMMIT` to
   the merged commit and invoke the script the image already installs
   (`/opt/openvla-oft/vla-scripts/finetune.py`) instead of `COPY`-ing a divergent copy. That
   removes the fork, the drift, the mislabel, and the script/package version-skew in one move.


---

### Latent items (dormant in this recipe; worth fixing before the script is generalized)

All upstream-inherited and not triggered by the current flags: the manual LR-warmup path is
overwritten by `MultiStepLR.step()` (dormant — `lr_warmup_steps=0`); `action_head` would
raise `NameError` in pure discrete mode and no assert guards it (dormant —
`use_l1_regression=True`); and the `log_step == cfg.max_steps` stop should be `>=` for safety
(dormant — infinite RLDS + `grad_accumulation_steps=1`). Three further items were checked and
dismissed as harmless: `labels` not moved to device (transformers moves it internally), an
empty-validation `IndexError`, and a one-step LR-logging lag.

KeitaW (commented):


### A few doc-accuracy nits worth a quick pass

**Files:** `Dockerfile` (line 72), `kubernetes/libero/README.md` (clone step),
`kubernetes/libero/env_vars.example` (line 27)

Grouping the small ones so they don't scatter:

- **`Dockerfile:72`** — the patch comment says "The DLC base ships PyTorch 2.9 already
  version-pinned with NCCL 2.22+", but the `FROM` pins the PyTorch 2.8.0 DLC. I'd align
  the comment with the actual tag to avoid confusing a future maintainer deciding
  whether a bump is safe.
- **Clone instructions** — the walkthrough clones
  `https://github.com/awslabs/awsome-distributed-training/` and `cd`s into that path.
  The repo is now `awsome-distributed-ai`; GitHub redirects, but it's worth updating
  for accuracy (and the `cd` path will change anyway if the directory is flattened).
- **`env_vars.example:27`** — `IMAGE_URI` defaults to `…/openvla-oft:latest`, and the
  README builds/pushes `:latest`. The repo convention is fixed tags over `latest`; even
  for a user-built image, a pinned tag (or a clear "replace with your tag" placeholder)
  avoids stale-image confusion — which your own `ncclDevCommDestroy` troubleshooting
  entry calls out as a real failure mode.

---

## Things That Look Great

- **The troubleshooting catalogue in `kubernetes/libero/README.md` is exceptional** —
  `ncclDevCommDestroy`, the `set_num_images_in_input` Hub-stub mismatch, the
  `PrismaticProcessor` concurrent-write race, Gloo `connectFullMesh`, the
  `libcudart.so.12` flash-attn mismatch, NVML on CPU nodes. Each has symptom → cause →
  fix. This is the kind of hard-won knowledge that usually lives in someone's head.
- **The `main_process_first()` fix** wrapping `from_pretrained` / `snapshot_download` is
  a correct, minimal fix for the dynamic-modules-cache race across 8 ranks — and the
  reasoning is documented inline and in the README.
- **NCCL/Gloo networking is right:** `NCCL_SOCKET_IFNAME=^lo` uses the correct
  exclusion pattern (not positive selection), and the `GLOO_SOCKET_IFNAME=lo` choice is
  both correct for single-pod and explicitly flagged as needing to change for
  multi-node — exactly the nuance that usually gets missed.
- **Fail-loud env-var handling:** `set -u` plus the allow-list check in the Download
  Job, and the `${TASK_SUITE:?…}` parameter expansion in `env_vars`, make
  misconfiguration fail at render time with an actionable message instead of mid-run.
- **Reproducibility discipline:** both `openvla-oft` and `dlimp` are pinned to explicit
  commits, the `sed` patches `grep`-verify their own results so a future upstream change
  fails the build loudly, and the build ends with an import sanity check.
- **Thoughtful storage story:** three clearly-delineated PVC paths (dynamic / static /
  pre-existing `fsx-claim`) with `Retain` on the static PV so a teardown can't nuke a
  pre-provisioned filesystem.

---

3.test_cases/pytorch/vla/openvla/openvla-oft/src/requirements.txt
sagarneeldubey: Removed  `requirements.txt`  entirely (all deps were redundant with upstream's editable install) and dropped the corresponding pip install -r step from the Dockerfile, so stock transformers can no longer shadow the required fork.

sagarneeldubey:
> ## Review Batch 5/5 — Documentation Consistency
> ### A few doc-accuracy nits worth a quick pass
> **Files:** `Dockerfile` (line 72), `kubernetes/libero/README.md` (clone step), `kubernetes/libero/env_vars.example` (line 27)
> 
> Grouping the small ones so they don't scatter:
> 
> * **`Dockerfile:72`** — the patch comment says "The DLC base ships PyTorch 2.9 already
>   version-pinned with NCCL 2.22+", but the `FROM` pins the PyTorch 2.8.0 DLC. I'd align
>   the comment with the actual tag to avoid confusing a future maintainer deciding
>   whether a bump is safe.
> * **Clone instructions** — the walkthrough clones
>   `https://github.com/awslabs/awsome-distributed-training/` and `cd`s into that path.
>   The repo is now `awsome-distributed-ai`; GitHub redirects, but it's worth updating
>   for accuracy (and the `cd` path will change anyway if the directory is flattened).
> * **`env_vars.example:27`** — `IMAGE_URI` defaults to `…/openvla-oft:latest`, and the
>   README builds/pushes `:latest`. The repo convention is fixed tags over `latest`; even
>   for a user-built image, a pinned tag (or a clear "replace with your tag" placeholder)
>   avoids stale-image confusion — which your own `ncclDevCommDestroy` troubleshooting
>   entry calls out as a real failure mode.
> 
> ## Things That Look Great
> * **The troubleshooting catalogue in `kubernetes/libero/README.md` is exceptional** —
>   `ncclDevCommDestroy`, the `set_num_images_in_input` Hub-stub mismatch, the
>   `PrismaticProcessor` concurrent-write race, Gloo `connectFullMesh`, the
>   `libcudart.so.12` flash-attn mismatch, NVML on CPU nodes. Each has symptom → cause →
>   fix. This is the kind of hard-won knowledge that usually lives in someone's head.
> * **The `main_process_first()` fix** wrapping `from_pretrained` / `snapshot_download` is
>   a correct, minimal fix for the dynamic-modules-cache race across 8 ranks — and the
>   reasoning is documented inline and in the README.
> * **NCCL/Gloo networking is right:** `NCCL_SOCKET_IFNAME=^lo` uses the correct
>   exclusion pattern (not positive selection), and the `GLOO_SOCKET_IFNAME=lo` choice is
>   both correct for single-pod and explicitly flagged as needing to change for
>   multi-node — exactly the nuance that usually gets missed.
> * **Fail-loud env-var handling:** `set -u` plus the allow-list check in the Download
>   Job, and the `${TASK_SUITE:?…}` parameter expansion in `env_vars`, make
>   misconfiguration fail at render time with an actionable message instead of mid-run.
> * **Reproducibility discipline:** both `openvla-oft` and `dlimp` are pinned to explicit
>   commits, the `sed` patches `grep`-verify their own results so a future upstream change
>   fails the build loudly, and the build ends with an import sanity check.
> * **Thoughtful storage story:** three clearly-delineated PVC paths (dynamic / static /
>   pre-existing `fsx-claim`) with `Retain` on the static PV so a teardown can't nuke a
>   pre-provisioned filesystem.

`Dockerfile`:  Fixed the stale "PyTorch 2.9" in the patch comment to "PyTorch 2.8" to match the pinned 2.8.0-...-cu129 DLC base; the other 2.9.0/cu130 mentions are intentional (they describe the newer DLC the recipe deliberately avoids).
`Readme.md`: updated the clone instructions to use the new canonical repo URL https://github.com/awslabs/awsome-distributed-ai/, and updated the cd path to awsome-distributed-ai/... to match the directory the clone creates.
`env_vars.example`: Pinned the image tag to the Dockerfile's OPENVLA_OFT_COMMIT (openvla-oft:e4287e9) instead of :latest across env_vars.example, the README build/rebuild commands (via an IMAGE_TAG var), and the download-job comment, matching the repo's fixed-tag convention.

sagarneeldubey:
> ## Review Batch 2/5 — Deployment Pipeline
> ### `envsubst` is invoked without an explicit allow-list in the default path
> **File:** `kubernetes/libero/README.md`, `kubernetes/libero/*.yaml`
> 
> The manifests are carefully written against the `$$`-escape convention and the README documents the GNU-`envsubst` allow-list form as a fallback — that's genuinely good defensive work. That said, the default commands throughout the walkthrough are bare `envsubst < file | kubectl apply -f -`, which substitutes every `$NAME` in the environment and only behaves because of the `$$` escaping. I'd suggest promoting the allow-list form (`envsubst '$IMAGE_URI $DATA_PVC_NAME $NAMESPACE $TASK_SUITE $TASK_SUITE_DNS' < …`) to the _primary_ instruction rather than the fallback, so the recipe is robust regardless of which `envsubst` is first on a contributor's PATH.
> 
> ### The bundled `tests/libero/` suite tests the wrong layer — I'd drop it from this PR
> **Files:** the whole `tests/libero/` directory — **~1,650 lines of test code across 8 files** for a single training example.
> 
> I went through each file. The pattern is consistent: the suite validates the cheap, self-evident layer (bash idioms, static YAML, an obvious helper, third-party tools) and never touches the one that matters for a training recipe — _does the image build and train?_ File by file:
> 
> File	Lines	What it does	Verdict
> `test_properties.py`	677	Property tests P1–P7 (breakdown below)	**Drop**
> `conftest.py`	317	Fixtures: `envsubst` path resolution, render + YAML-parse helpers, well-formed / malformed TFDS-tree factories	**Drop** — exists only to serve the property tests
> `test_p8_rlds_dataset_failfast.py`	179	P8: `RLDSDataset` on an empty/malformed dir raises quickly and names the missing path	**Drop** — exercises upstream's loader, not the contributor's code
> `fixtures/hf-cli-shim`	153	Fake `huggingface-cli download` that idempotently creates the TFDS layout	**Drop** — exists only for the tautological P6
> `render_all_manifests.sh`	118	Renders every manifest via `envsubst`, validates via `kubectl apply --dry-run=client` or a PyYAML fallback	Closest to useful (envsubst is the fragile, bespoke piece), but subsumed by an integration test
> `smoke-verify-tfds-layout.sh`	105	Bash smoke test of `verify-tfds-layout.sh` (happy path + two malformed cases)	**Drop** — tests an obvious 33-line helper
> `fixtures/env_vars.test`	68	Test-only env-var values so manifests render under `--dry-run=client` (cleanly made — AWS-ID-shaped placeholders, no `:?`, no secrets)	**Drop** — a well-built fixture, but only feeds tests that are themselves going
> `shellcheck.sh`	37	`shellcheck`s the `env_vars` file, skips if shellcheck absent	**Drop** — lints one file; CI already runs `flake8`/`bandit`/`bash -n`
> **Property-by-property (`test_properties.py` + `test_p8`):**
> 
> * **P1** — allow-list accept/reject: ~5 lines of standard bash (`grep -qxF`); doesn't stop the allow-list from diverging across the three places it's duplicated.
> * **P2** — fail-loud on unset `TASK_SUITE`: asserts `${VAR:?}` aborts — i.e., tests a _bash language feature_.
> * **P3** — distinct suites → distinct manifest names: asserts `${TASK_SUITE//_/-}` differs per suite; a one-line property.
> * **P4 / P5** — `verify-tfds-layout.sh` accepts/rejects trees: tests the contributor's own obvious 33-line helper.
> * **P6** — Download-Job idempotence: stubs `huggingface-cli` _and_ `pip` with fakes, runs the wrapper twice, asserts a byte-identical tree. The stub is idempotent by construction → tautology; and it excludes every real download risk (partial write, auth, rate-limit, network).
> * **P7** — render → `kubectl dry-run` → YAML preserves the training flags: those flags are _literal text_ `envsubst` never substitutes, so this is largely a static-text snapshot.
> * **P8** — `RLDSDataset` fails fast on an empty PVC: real operator value in the "actionable message" angle, but it's testing upstream's loader, not this PR's code.
> 
> Nothing across the 1,650 lines builds the image or runs a training step — exactly what `max_steps=100` was sized for — and the whole approach diverges from the repo's `docker_build` / `docker_run` integration-test convention.
> 
> I'd suggest **dropping the entire `tests/libero/` directory from this PR.** If you want coverage, add it as a focused follow-up in the repo's integration style: a `docker_build` + a short `docker_run` that builds the image and runs `max_steps=100` to a first loss line / checkpoint write. That one test is worth more than all eight properties — and because it renders and applies the manifests as part of running, it also subsumes the only check here with genuine value (that the `$$`-escaped `envsubst` still produces valid YAML), so nothing real is lost by removing the unit suite now.

1/ Dropped the entire `tests/libero/` directory
2/ I dug into envsubst comment and the allow-list form actually can't be promoted to the default: an allow-list only controls which names GNU substitutes — it can't make GNU honor the $$ escape or ${VAR:-default} expansion these manifests depend on (the latter is used by ${RUN_ID_NOTE:-…} in `libero-finetune.yaml`). That incompatibility is exactly why the recipe standardized on a8m/envsubst. Promoting the bare allow-list as the default would hand GNU users broken YAML.

Instead I documented the nuance in README §0.4 and the libero-download.yaml header:

Called out up front that GNU gettext supports neither $$ nor ${VAR:-default}, with a8m/envsubst as the recommended tool that renders cleanly out of the box.
Reframed the GNU path as a real fallback that requires editing the manifests first (collapse $$NAME → $NAME, expand ${RUN_ID_NOTE:-…} by hand), then rendering with an allow-list.
Added correct per-manifest allow-lists (the original example was missing TASK_SUITE and NAMESPACE); the two FSx manifests have no $$/:-default and work with an allow-list alone.

3.test_cases/pytorch/vla/openvla/openvla-oft/kubernetes/libero/libero-finetune.yaml
sagarneeldubey: Done

sagarneeldubey:
> ## Review Batch 4/5 — Training Logic & Implementation
> `src/finetune.py` is vendored from upstream `moojink/openvla-oft` (`vla-scripts/finetune.py` @ `e4287e9`). Its training algorithm is faithful to upstream and correct for this recipe — DDP + LoRA with a continuous L1-regression action head, so the usual ZeRO-3 traps (`device_map`, sharded `save_pretrained`) don't apply, and the continuous-head forward path correctly extracts the action-token hidden states and computes L1 over the full `(B, 8, 7)` chunk. The issues below live in the dependency and config context the script runs in, not in its algorithm.
> 
> ### The `transformers`-fork shadowing is the one training-correctness blocker
> This is the training-time face of the `requirements.txt` finding above. At `finetune.py:626` the model loads via `AutoModelForVision2Seq.from_pretrained(..., trust_remote_code=True)`; OFT parallel decoding only works if the `moojink/transformers-openvla-oft` fork (which makes `LlamaModel.forward` attention bidirectional) is the installed `transformers`. If the stock wheel wins the install race, the model still loads and trains — under **causal** attention — with no error and no NaN: the loss falls but the policy is structurally wrong and incompatible with OFT inference. Fix it at the source (drop the `transformers` line from `requirements.txt`); a startup guard in `finetune.py` is reasonable defense-in-depth, but tie it to a _verified_ fork marker — the fork changes `forward()` logic, not a persistent attribute, so a naive assertion can pass on a broken install.
> 
> ### `merge_lora_during_training` writes ~150 GB of throwaway merged checkpoints in the smoke run
> **Files:** `kubernetes/libero/libero-finetune.yaml` (the `command:` save flags), `src/finetune.py` (`save_training_checkpoint`, lines 436–487; save trigger at line 839)
> 
> Yes — I traced this through the code. The chain, step by step:
> 
> **The four flags interact like this:**
> 
> Flag (from the manifest)	Value	Effect
> `--max_steps`	100	training stops at step 100
> `--save_freq`	10	a checkpoint is written every 10 steps
> `--save_latest_checkpoint_only`	False	each checkpoint goes to its **own** dir (`…--<step>_chkpt/`); nothing is overwritten
> `--merge_lora_during_training`	True	each checkpoint **also** merges LoRA into the base and writes the **full 7B model**
> **How many checkpoints?** The loop saves when `gradient_step_idx > 0 and log_step % save_freq == 0` (`finetune.py:839`), and the `max_steps` break (`:870`) runs _after_ the save. With `grad_accumulation_steps=1`, `log_step` advances by 1 per step, so saves fire at steps **10, 20, 30, …, 100 — ten of them** (step 100 saves, then breaks).
> 
> **Why each one is ~15 GB, not a few hundred MB.** This is the crux. A LoRA checkpoint is normally tiny — just the rank-32 adapter (a few hundred MB) plus the small action/proprio heads. But `--merge_lora_during_training=True` makes `save_training_checkpoint` reload the base model, call `merge_and_unload()`, then `merged_vla.save_pretrained(checkpoint_dir)` — writing the **entire merged OpenVLA-7B** (~15 GB in bf16) into that step's directory. The merge inflates each checkpoint roughly **50×**, and `save_latest_checkpoint_only=False` keeps all ten.
> 
> **The tally on the PVC.** Ten merged models × ~15 GB ≈ **150 GB of checkpoints**, sharing the volume with everything else the run writes to `/data`:
> 
> * `/data/hf-cache` — the base `openvla-7b` snapshot the run downloads (~15 GB),
> * `/data/datasets/rlds` — the suite the Download Job staged (~2.5 GB for one LIBERO suite),
> * `/data/runs/<run_id>` — these ten checkpoints.
> 
> ≈ **165–170 GB total.** The README recommends a PVC of **"at least 100 GiB"** (and notes checkpoints colocate with the dataset on the same volume), so on a minimum-spec PVC the run hits **`ENOSPC` partway through** — and it dies _late_, after burning most of the smoke test's wall-clock, the worst time to fail. (It does fit the `DATA_PVC_SIZE=1200Gi` _default_ in `env_vars.example`, so the inconsistency is specifically the README's 100 GiB minimum vs. what the shipped flags actually write.)
> 
> **Why it's all wasted:** this is a 100-step smoke test — ~0.07 % of the ~150k steps upstream uses for a real run — so the ten merged 7B checkpoints are throwaway artifacts of an unconverged model. The test only needs to prove the save/merge path works _once_.
> 
> **Fix (any one):** `--save_latest_checkpoint_only=True` (one dir, overwritten — ~15 GB instead of ~150 GB), or `--save_freq` ≥ `max_steps` (a single save at the end), or drop `--merge_lora_during_training` for the smoke run (keep only the small adapter). Separately, the README's 100 GiB minimum should be reconciled with what the shipped flags actually write.
> 
> ### `src/finetune.py` is modified, not "vendored verbatim" — and shouldn't be vendored at all
> `src/finetune.py`'s header states it is _"vendored verbatim from the openvla-oft repository."_ It is not. Stripping docstrings and comments from both files and diffing the pure code:
> 
> * Raw line counts: upstream **1142** → vendored **876** (−266 lines).
> * After removing docstrings/comments: upstream **383** vs vendored **385** code lines.
> * **The entire code delta is two edits — both the same change:** wrapping the model-load
>   calls in `main_process_first()` (`main_process_first` appears 3× in the vendored file, 0×
>   upstream).
> 
> **1. `snapshot_download` race fix** (`finetune.py:593–596`):
> 
> ```python
> # upstream:                                  # vendored:
> if model_is_on_hf_hub(cfg.vla_path):         if model_is_on_hf_hub(cfg.vla_path):
>     vla_download_path = snapshot_download(       with distributed_state.main_process_first():
>         repo_id=cfg.vla_path)                        vla_download_path = snapshot_download(repo_id=cfg.vla_path)
> ```
> 
> **2. `from_pretrained` race fix** (`finetune.py:624–631`):
> 
> ```python
> # upstream:                                  # vendored:
> processor = AutoProcessor.from_pretrained(   with distributed_state.main_process_first():
>     cfg.vla_path, trust_remote_code=True)        processor = AutoProcessor.from_pretrained(cfg.vla_path, trust_remote_code=True)
> vla = AutoModelForVision2Seq.from_pretrained(    vla = AutoModelForVision2Seq.from_pretrained(
>     cfg.vla_path, ...).to(device_id)                 cfg.vla_path, ...).to(device_id)
> ```
> 
> These fix the two distributed races the README troubleshooting documents (the 8× HF-Hub-hit / `update_auto_map` race, and the `trust_remote_code` dynamic-modules-cache partial read). Everything else in the 266-line delta is docstrings/comments — including the module docstring, which was rewritten to _add_ the self-refuting "vendored verbatim" sentence that isn't in upstream at all.
> 
> **Why it matters:** "verbatim" invites a future maintainer to re-vendor from upstream to "resync," which would silently drop both race fixes and reintroduce the exact crashes the README documents.
> 
> **Recommended path — don't carry a patched `finetune.py` in this repo at all.** The two `main_process_first()` fixes are general-purpose (any multi-rank `openvla-oft` user hits these races), so their home is upstream, not a vendored fork that won't track future upstream changes and will silently skew against the pinned package (the script and the `pip install -e`'d package must move in lockstep). Concretely:
> 
> 1. **Open a PR against `moojink/openvla-oft`** with the two race fixes (and, while there, the
>    `merge_lora_during_training` all-rank-merge fix above — same root, same beneficiaries).
> 2. **Once merged, drop the vendored `src/finetune.py` entirely:** bump `OPENVLA_OFT_COMMIT` to
>    the merged commit and invoke the script the image already installs
>    (`/opt/openvla-oft/vla-scripts/finetune.py`) instead of `COPY`-ing a divergent copy. That
>    removes the fork, the drift, the mislabel, and the script/package version-skew in one move.
> 
> ### Latent items (dormant in this recipe; worth fixing before the script is generalized)
> All upstream-inherited and not triggered by the current flags: the manual LR-warmup path is overwritten by `MultiStepLR.step()` (dormant — `lr_warmup_steps=0`); `action_head` would raise `NameError` in pure discrete mode and no assert guards it (dormant — `use_l1_regression=True`); and the `log_step == cfg.max_steps` stop should be `>=` for safety (dormant — infinite RLDS + `grad_accumulation_steps=1`). Three further items were checked and dismissed as harmless: `labels` not moved to device (transformers moves it internally), an empty-validation `IndexError`, and a one-step LR-logging lag.

Fixed:

Checkpoint bloat: set --save_latest_checkpoint_only=True in libero-finetune.yaml. One overwritten ~15 GB dir instead of ten merged 7B checkpoints (~150 GB). The smoke test only needs to prove the save/merge path once, so kept --merge_lora_during_training=True.
README capacity: reconciled the 100 GiB minimum with what the flags actually write — added a footprint breakdown (~30–40 GiB steady state) and a heads-up that flipping the flag back to False re-inflates to ~150 GiB.
"vendored verbatim" mislabel: corrected the 
finetune.py
 docstring (and README file-tree label) to "adapted from openvla-oft: +2 race fixes."
On dropping the patched script: agreed it shouldn't live here long-term. Keeping it for now since the script and the pinned package must move in lockstep and the two main_process_first() race fixes are required for multi-rank runs. I've prepared a PR against moojink/openvla-oft with both fixes (branched off the pinned commit e4287e9). Once it merges, a follow-up will delete the vendored copy, bump OPENVLA_OFT_COMMIT, and invoke the installed 
finetune.py
 directly.

Note: the upstream PR link is https://github.com/moojink/openvla-oft/pull/155 but not merged yet

3.test_cases/pytorch/openvla-oft/src/finetune.py
sagarneeldubey: Made the change in the patched copy so that only rank 0 loads the base model, merges the adapter, and save the result. Changes will be made in the upstream file through the same PR (once  created). 
Once again we can drop the patched script once the upstream fixes are applied.

sagarneeldubey:
@KeitaW All comments addressed. Also re-tested the training job and run successful:

Earlier - the model checkpoints were saved multiple times for each step by each rank. With the latest change it is only saved once by the main task (rank 0).

Before the change: 
same step checkpoint saved more than once
before


After the change: 
One checkpoint saved only once

after

KeitaW (approved):
Thanks @sagarneeldubey for addressing the issues! Looking forward to the future PR once https://github.com/moojink/openvla-oft/pull/155 is merged.

Issue #1056, opened 2026-04-06
[RFC]: Repository reorganization — remove numbering, merge awsome-inference


KeitaW:
## Motivation

The current `awsome-distributed-training` repository (now `awsome-distributed-ai`) has grown to cover a broad set of distributed ML workloads on AWS, but the directory layout carries legacy decisions that make it harder to navigate and contribute to:

1. **Numbered prefixes impose a rigid ordering** that no longer reflects how users discover content. Directories like `1.architectures/`, `2.ami_and_containers/`, `3.test_cases/` suggest a sequential workflow, but most users land directly on a specific test case or architecture via search or a shared link. The numbering also creates friction when adding new top-level directories (what number should it get?).

2. **Training and inference are split across two separate repositories.** The [`awsome-inference`](https://github.com/aws-samples/awsome-inference) repo covers inference workloads (SGLang, TRT-LLM, NIMs, Ray Serve, Dynamo, etc.) with its own infrastructure and project structure. Maintaining two repos leads to duplicated IaC (both have VPC/EKS setup), divergent conventions, and a fragmented contributor/user experience. Users doing both training *and* inference on the same cluster shouldn't need two repos.

3. **The `test_cases/` directory has two kinds of content mixed together.** Some examples are *framework-centric* — the demo's subject is the training/inference framework (FSDP, Megatron-LM, NeMo), and model variants underneath illustrate it (`pytorch/FSDP/llama3`, `megatron/nemo/qwen3`). Others are *use-case-centric* — the demo's subject is a specific model or use case, and the framework is incidental (`pytorch/ddp/detr-finetune` is really a DETR-on-COCO fine-tuning example, not a DDP demo). Forcing both into a single framework-first hierarchy buries the latter under the wrong axis and makes them undiscoverable.

### 1. Remove numeric prefixes from all directories

| Current | Proposed |
|---|---|
| `0.docs/` | `docs/` |
| `1.architectures/` | `architectures/` |
| `2.ami_and_containers/` | `ami_and_containers/` |
| `3.test_cases/` | `examples/` |
| `4.validation_and_observability/` | `validation_and_observability/` |
| `micro-benchmarks/` | `micro-benchmarks/` (unchanged) |

Also remove numbering from subdirectories within `architectures/`:

| Current | Proposed |
|---|---|
| `architectures/0.common/` | `architectures/common/` |
| `architectures/1.vpc_network/` | `architectures/vpc_network/` |
[...]

PR #1110, opened 2026-05-27
DeepSeek-V3 Disaggregated Inference with vLLM, UCCL-EP, and NIXL on EKS

pbelevich:
## Purpose

Add a new test case demonstrating **disaggregated DeepSeek-V3 inference with vLLM, UCCL-EP, and NIXL over AWS EFA**. The sample reproduces a 6-node scaling study (1N → 1P+5D) that quantifies the latency and throughput trade-offs of NIXL-based prefill/decode disaggregation for a 671B-parameter MLA + 256-expert MoE model.

Relates to ongoing work consolidating inference samples from `aws-samples/awsome-inference` into `awsome-distributed-ai`.

## Changes

- New test case at `3.test_cases/pytorch/vllm/dsv3-uccl-nixl/` covering DeepSeek-V3 disaggregated serving on EKS / SageMaker HyperPod EKS.
- `Dockerfile` building **vLLM 0.21.0** from source on **CUDA 13** with **NIXL 1.0.1** (`nixl-cu13`), **UCCL-EP** (commit `0dc87eb`) + the `deep_ep` drop-in wrapper, **Ray 2.55.1**, EFA installer 1.48.0, NCCL 2.30.4, GDRCopy 2.5.2 — all version-pinned.
- `manifests/{prefill,decode,proxy,unified,hf-token-secret}.yaml` — `envsubst`-templated pod manifests.
  - Prefill uses `--all2all-backend deepep_high_throughput`, decode uses `deepep_low_latency`.
  - KV transfer via `NixlConnector` + `LIBFABRIC` over EFA (side-channel on port 5600, host network).
  - Per-node parallelism: `TP=4`, `DP=2`, `EP=8` on 8×H200.
- `recipe/deploy.sh` — renders manifests for a chosen topology (`unified|1p1d|1p3d|1p4d|1p5d`) using **scoped** `envsubst` so embedded shell variables (`$endpoint`, `$MY_IP`, …) survive untouched.
- `recipe/benchmark.sh` — warmup + rate sweep at 1/4/8/16/inf RPS via `vllm bench serve`, fixed seed, `--save-result`.
- `recipe/teardown.sh`, `setup/build-push.sh`, `setup/install-prereqs.sh` (NVIDIA + EFA device plugins for vanilla EKS), `setup/env_vars.example`.
- `README.md` with architecture diagram, results table, prerequisites, knob reference, troubleshooting, and cost estimate.
- `3.test_cases/pytorch/vllm/README.md` index entry.
- `.gitignore` keeps `setup/env_vars` (HF token, account ID) and `results/` out of git.

## Test Plan

**Environment:**
- AWS Service: Amazon EKS (also tested on SageMaker HyperPod EKS, ap-southeast-3 and us-west-2)
- Instance type: `p5en.48xlarge` (8× H200 141GB, 16× EFA, 3,200 Gbps)
- Number of nodes: 1 → 6 (covering 1N, 1P+1D, 1P+2D, 1P+3D, 1P+4D, 1P+5D, 2P+3D, 2P+4D, plus an AGRS backend comparison)

**Test commands:**
```bash
cd 3.test_cases/pytorch/vllm/dsv3-uccl-nixl

cp setup/env_vars.example setup/env_vars
$EDITOR setup/env_vars                 # set REGISTRY, HF_TOKEN, node hostnames + IPs
source setup/env_vars

./setup/build-push.sh                  # build + ECR push

kubectl create namespace "$NAMESPACE"
kubectl create secret generic hf-token --namespace "$NAMESPACE" \
    --from-literal=HF_TOKEN="$HF_TOKEN"

./recipe/deploy.sh 1p4d                # 1 prefill + 4 decode (5 nodes, peak)

# wait until proxy logs "All instances ready!"
curl -sf "http://${PREFILL_IP}:8000/health"

./recipe/benchmark.sh 1p4d             # warmup + rate sweep, results -> results/1p4d/

./recipe/teardown.sh
```

## Test Results

Workload: `--random-input-len 1024 --random-output-len 256 --num-prompts 50` via `vllm bench serve`.

| Config       | Nodes | GPUs | Burst RPS | Burst tok/s | P50 TPOT @1RPS | P50 TPOT @burst | P50 TTFT @1RPS | tok/s/GPU |
| ------------ | :---: | :--: | --------: | ----------: | -------------: | --------------: | -------------: | --------: |
| 1N unified   |   1   |   8  |     13.49 |       3,454 |        21.83ms |         44.00ms |          77ms  |    431.55 |
| 1P+1D       |   2   |  16  |     15.94 |       4,082 |        20.93ms |         40.91ms |         313ms  |    255.10 |
| 1P+3D       |   4   |  32  |     17.22 |       4,408 |        18.21ms |         33.76ms |         343ms  |    137.75 |
| **1P+4D**   |   5   |  40  | **19.18** |   **4,910** |        17.81ms |         32.51ms |         303ms  |    122.75 |
| 1P+5D       |   6   |  48  |     17.16 |       4,392 |        17.81ms |         30.34ms |         268ms  |     91.51 |
| 1P+3D AGRS  |   4   |  32  |     10.62 |       2,721 |        23.30ms |         48.20ms |         310ms  |     85.03 |

**Headline findings**

1. **Disaggregation pays off mostly in latency, not throughput.** A single unified node has the best `tok/s/GPU` (431.55); disaggregation lowers burst TPOT 44ms → 30ms but only adds ~42% peak RPS for 5× the GPUs.
2. **The optimal P:D ratio is 1:4** at this model size — beyond 4 decode nodes the single prefill node is the hard bottleneck (1P+5D burst RPS regresses).
3. **UCCL-EP (`deepep_high_throughput` + `deepep_low_latency`) beats `allgather_reducescatter` (AGRS) by ~62%** on burst RPS and ~30% on TPOT for this MLA + 256-expert MoE model.
4. **`hostNetwork: true` + `vpc.amazonaws.com/efa: 16` are mandatory.** With TCP fallback (no EFA), the same workload runs ~28× slower.

The full per-rate breakdown, plots, and software-stack version capture live in the upstream experiment directory used to develop this sample.

## Directory Structure

```
3.test_cases/
└── pytorch/
    └── vllm/
        ├── README.md                  # index of vLLM test cases
        └── dsv3-uccl-nixl/
            ├── Dockerfile             # vLLM 0.21.0 + UCCL-EP + NIXL on CUDA 13 / EFA
            ├── README.md              # overview, prerequisites, usage
            ├── .gitignore             # keeps setup/env_vars and results/ out of git
            ├── manifests/             # Kubernetes pod manifests (envsubst-templated)
            │   ├── prefill.yaml
            │   ├── decode.yaml
            │   ├── proxy.yaml
            │   ├── unified.yaml
            │   └── hf-token-secret.yaml
            ├── recipe/                # deploy / benchmark / teardown helpers
            │   ├── deploy.sh
            │   ├── benchmark.sh
            │   └── teardown.sh
            └── setup/                 # build + cluster prereqs
                ├── env_vars.example
                ├── build-push.sh
                └── install-prereqs.sh
```

This is a Kubernetes-native deployment sample (EKS + SageMaker HyperPod EKS), so the `slurm/` and `hyperpod-eks/` siblings are not applicable — the manifests are generic enough to run on either orchestrator unmodified.

## Checklist

- [x] External dependencies are pinned to a specific version or tag (no `latest`) — vLLM `v0.21.0`, UCCL `0dc87eb…`, NIXL `v1.0.1`, NCCL `v2.30.4-1`, EFA `1.48.0`, GDRCopy `v2.5.2`, Ray `>=2.43`.
- [x] A README is included with prerequisites, instructions, and known issues.

### DIFF
--- a/3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
+++ b/3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
@@ -1,5 +1,5 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: MIT-0
 #
 # vLLM 0.21.0 + UCCL-EP + NIXL container image for DeepSeek-V3 disaggregated
 # inference on AWS EKS / SageMaker HyperPod EKS with EFA.
@@ -10,7 +10,7 @@
 #   - EFA installer 1.48.0
 #   - NCCL 2.30.4 (CUDA arches 8.0/8.6/8.9/9.0/10.0/10.3)
 #   - vLLM 0.21.0 from source
-#   - NIXL 1.0.1 (cu13 build)
+#   - NIXL 1.1.0 (CUDA-13 backend auto-selected via the meta wheel)
 #   - UCCL (commit 0dc87eb) + uccl.ep + deep_ep wrapper
 #   - Ray 2.55.1 for DP coordination
 #
@@ -25,7 +25,7 @@ ARG EFA_INSTALLER_VERSION=1.48.0
 ARG NCCL_VERSION=v2.30.4-1
 ARG NCCL_TESTS_VERSION=v2.18.3
 ARG VLLM_COMMIT=v0.21.0
-ARG NIXL_VERSION=v1.0.1
+ARG NIXL_VERSION=v1.1.0
 ARG UCCL_COMMIT=0dc87eb3b40c372a16b70ef320f37daaa5299ca7
 
 # --- system packages ---------------------------------------------------------
@@ -139,15 +139,14 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 
 # --- Ray (DP coordinator) ----------------------------------------------------
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv pip install --system "ray[default]>=2.43"
+    uv pip install --system "ray[default]==2.55.1"
 
-# --- NIXL (cu13) -------------------------------------------------------------
-# Install nixl-cu13 first, then nixl meta package, then force-reinstall
-# nixl-cu13 to keep cu13 ownership of nixl_ep namespace.
+# --- NIXL --------------------------------------------------------------------
+# v1.1.0 ships cu12 and cu13 backends inside the meta wheel and auto-selects
+# at runtime based on torch.version.cuda — single-line install. (v1.0.1 had
+# an unconditional cu12 dep that required a 3-step reinstall dance.)
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv pip install --system "nixl-cu13==$(echo ${NIXL_VERSION} | sed 's/^v//')" \
-    && uv pip install --system "nixl==$(echo ${NIXL_VERSION} | sed 's/^v//')" \
-    && uv pip install --system --reinstall --no-deps "nixl-cu13==$(echo ${NIXL_VERSION} | sed 's/^v//')"
+    uv pip install --system "nixl==$(echo ${NIXL_VERSION} | sed 's/^v//')"
 
 # --- UCCL-EP + deep_ep wrapper ----------------------------------------------
 RUN git clone https://github.com/uccl-project/uccl.git /opt/uccl \
@@ -166,13 +165,22 @@ ENV PER_EXPERT_BATCHING=1
 RUN TORCH_CUDA_ARCH_LIST="9.0a+PTX;10.0a+PTX;10.3a+PTX" python3 setup.py install
 
 WORKDIR /opt/uccl/ep/deep_ep_wrapper
-RUN pip install --no-deps .
+RUN pip install .
 
 # --- smoke tests -------------------------------------------------------------
 RUN echo "=== Smoke test: vLLM ===" \
     && python3 -c "import vllm; print(f'vLLM {vllm.__version__}')" \
     && echo "=== Smoke test: NIXL ===" \
     && python3 -c "import nixl; print('NIXL OK')" \
+    && echo "=== Smoke test: NIXL backend matches CUDA 13 ===" \
+    && python3 -c "import importlib.util, subprocess, os; \
+spec = importlib.util.find_spec('nixl_ep'); \
+assert spec, 'nixl_ep not found'; \
+so = next(p for p in os.listdir(os.path.dirname(spec.origin)) if p.endswith('.so')); \
+out = subprocess.check_output(['ldd', os.path.join(os.path.dirname(spec.origin), so)], text=True); \
+assert 'libcudart.so.13' in out, 'nixl_ep does not link cu13'; \
+assert 'libcudart.so.12' not in out, 'nixl_ep also links cu12'; \
+print(f'nixl_ep cu13 OK: {so}')" \
     && echo "=== Smoke test: uccl.ep ===" \
     && python3 -c "import torch; import uccl.ep; print('uccl.ep OK')" \
     && echo "=== Smoke test: deep_ep ===" \
--- a/3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
+++ b/3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
@@ -1,0 +1,186 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# vLLM 0.21.0 + UCCL-EP + NIXL container image for DeepSeek-V3 disaggregated
+# inference on AWS EKS / SageMaker HyperPod EKS with EFA.
+#
+# Layered on top of an NCCL-tests style EFA base:
+#   - CUDA 13 devel
+#   - GDRCopy 2.5.2
+#   - EFA installer 1.48.0
+#   - NCCL 2.30.4 (CUDA arches 8.0/8.6/8.9/9.0/10.0/10.3)
+#   - vLLM 0.21.0 from source
+#   - NIXL 1.0.1 (cu13 build)
+#   - UCCL (commit 0dc87eb) + uccl.ep + deep_ep wrapper
+#   - Ray 2.55.1 for DP coordination
+#
+# Build:
+#   ./setup/build-push.sh
+#
+ARG CUDA_VERSION=13.0.3
+FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu24.04
+
+ARG GDRCOPY_VERSION=v2.5.2
+ARG EFA_INSTALLER_VERSION=1.48.0
+ARG NCCL_VERSION=v2.30.4-1
+ARG NCCL_TESTS_VERSION=v2.18.3
+ARG VLLM_COMMIT=v0.21.0
+ARG NIXL_VERSION=v1.0.1
+ARG UCCL_COMMIT=0dc87eb3b40c372a16b70ef320f37daaa5299ca7
+
+# --- system packages ---------------------------------------------------------
+RUN apt-get update -y && apt-get upgrade -y
+RUN apt-get remove -y --allow-change-held-packages \
+    ibverbs-utils libibverbs-dev libibverbs1 libmlx5-1 libnccl2 libnccl-dev \
+    || true
+
+RUN rm -rf /opt/hpcx /usr/local/mpi \
+    && rm -f /etc/ld.so.conf.d/hpcx.conf \
+    && ldconfig
+
+ENV OPAL_PREFIX=
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y --allow-unauthenticated \
+    apt-utils autoconf automake build-essential check cmake curl debhelper \
+    devscripts git gcc gdb kmod libsubunit-dev libtool openssh-client \
+    openssh-server pkg-config vim python3-dev python3-pip python3-venv
+
+RUN apt-get purge -y cuda-compat-* || true
+
+RUN mkdir -p /var/run/sshd
+RUN sed -i 's/[ #]\(.*StrictHostKeyChecking \).*/ \1no/g' /etc/ssh/ssh_config && \
+    echo "    UserKnownHostsFile /dev/null" >> /etc/ssh/ssh_config && \
+    sed -i 's/#\(StrictModes \).*/\1no/g' /etc/ssh/sshd_config
+
+ENV LD_LIBRARY_PATH=/usr/local/cuda/extras/CUPTI/lib64:/opt/amazon/openmpi/lib:/opt/nccl/build/lib:/opt/amazon/efa/lib:/opt/amazon/ofi-nccl/lib:/opt/amazon/ofi-nccl/lib/x86_64-linux-gnu:/usr/local/lib:$LD_LIBRARY_PATH
+ENV PATH=/opt/amazon/openmpi/bin:/opt/amazon/efa/bin:/usr/bin:/usr/local/bin:$PATH
+
+RUN rm -f /usr/lib/python3.12/EXTERNALLY-MANAGED /usr/lib/python3/EXTERNALLY-MANAGED
+RUN pip3 install --break-system-packages awscli pynvml
+
+# --- GDRCopy -----------------------------------------------------------------
+RUN git clone -b ${GDRCOPY_VERSION} https://github.com/NVIDIA/gdrcopy.git /tmp/gdrcopy \
+    && cd /tmp/gdrcopy \
+    && make prefix=/opt/gdrcopy install
+
+ENV LD_LIBRARY_PATH=/opt/gdrcopy/lib:$LD_LIBRARY_PATH
+ENV LIBRARY_PATH=/opt/gdrcopy/lib:$LIBRARY_PATH
+ENV CPATH=/opt/gdrcopy/include:$CPATH
+ENV PATH=/opt/gdrcopy/bin:$PATH
+
+# --- EFA installer -----------------------------------------------------------
+RUN cd $HOME \
+    && curl -O https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz \
+    && tar -xf aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz \
+    && cd aws-efa-installer \
+    && ./efa_installer.sh -y -g -d --skip-kmod --skip-limit-conf --no-verify --disable-ngc \
+    && rm -rf $HOME/aws-efa-installer
+
+RUN echo "Verifying AWS OFI NCCL plugin installation..." && \
+    (ls -la /opt/amazon/ofi-nccl/lib/libnccl-net*.so || \
+     ls -la /opt/amazon/ofi-nccl/lib/x86_64-linux-gnu/libnccl-ofi*.so)
+
+# --- NCCL --------------------------------------------------------------------
+RUN git clone -b ${NCCL_VERSION} https://github.com/NVIDIA/nccl.git /opt/nccl \
+    && cd /opt/nccl \
+    && make -j $(nproc) src.build CUDA_HOME=/usr/local/cuda \
+       NVCC_GENCODE="-gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_86,code=sm_86 -gencode=arch=compute_89,code=sm_89 -gencode=arch=compute_90,code=sm_90 -gencode=arch=compute_100,code=sm_100 -gencode=arch=compute_103,code=sm_103"
+
+# --- NCCL tests (handy for fi_info / health checks) --------------------------
+RUN git clone -b ${NCCL_TESTS_VERSION} https://github.com/NVIDIA/nccl-tests.git /opt/nccl-tests \
+    && cd /opt/nccl-tests \
+    && make -j $(nproc) MPI=1 MPI_HOME=/opt/amazon/openmpi/ \
+       CUDA_HOME=/usr/local/cuda NCCL_HOME=/opt/nccl/build \
+       NVCC_GENCODE="-gencode=arch=compute_80,code=sm_80 -gencode=arch=compute_86,code=sm_86 -gencode=arch=compute_89,code=sm_89 -gencode=arch=compute_90,code=sm_90 -gencode=arch=compute_100,code=sm_100 -gencode=arch=compute_103,code=sm_103"
+
+RUN rm -rf /var/lib/apt/lists/*
+
+ENV OMPI_MCA_pml=^ucx \
+    OMPI_MCA_btl=tcp,self \
+    OMPI_MCA_btl_tcp_if_exclude=lo,docker0,veth_def_agent \
+    OPAL_PREFIX=/opt/amazon/openmpi \
+    NCCL_SOCKET_IFNAME=^docker,lo,veth \
+    PMIX_MCA_gds=hash \
+    LD_PRELOAD=/opt/nccl/build/lib/libnccl.so
+
+# --- Python tooling (uv) -----------------------------------------------------
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+ENV PATH="/root/.local/bin:$PATH"
+ENV UV_HTTP_TIMEOUT=500
+ENV UV_LINK_MODE=copy
+
+RUN apt-get update -y && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    ccache libibverbs-dev libnl-3-dev libnl-route-3-dev libnuma-dev ninja-build \
+    && rm -rf /var/lib/apt/lists/*
+
+# --- vLLM from source --------------------------------------------------------
+ENV CUDA_HOME=/usr/local/cuda
+ENV MAX_JOBS=32
+ENV NVCC_THREADS=1
+ENV TORCH_CUDA_ARCH_LIST="8.0 8.6 8.9 9.0 10.0 10.3"
+ENV VLLM_TARGET_DEVICE=cuda
+
+WORKDIR /opt/vllm
+RUN git clone https://github.com/vllm-project/vllm.git . \
+    && git checkout ${VLLM_COMMIT}
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=cache,target=/root/.cache/ccache \
+    uv pip install --system --index-strategy unsafe-best-match \
+        -r requirements/build/cuda.txt \
+        --extra-index-url https://download.pytorch.org/whl/cu130 \
+    && uv pip install --system --index-strategy unsafe-best-match \
+        -r requirements/cuda.txt \
+        --extra-index-url https://download.pytorch.org/whl/cu130 \
+    && python3 setup.py bdist_wheel --dist-dir=dist --py-limited-api=cp38 \
+    && uv pip install --system --index-strategy unsafe-best-match \
+        dist/*.whl \
+        --extra-index-url https://download.pytorch.org/whl/cu130
+
+# --- Ray (DP coordinator) ----------------------------------------------------
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --system "ray[default]>=2.43"
+
+# --- NIXL (cu13) -------------------------------------------------------------
+# Install nixl-cu13 first, then nixl meta package, then force-reinstall
+# nixl-cu13 to keep cu13 ownership of nixl_ep namespace.
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --system "nixl-cu13==$(echo ${NIXL_VERSION} | sed 's/^v//')" \
+    && uv pip install --system "nixl==$(echo ${NIXL_VERSION} | sed 's/^v//')" \
+    && uv pip install --system --reinstall --no-deps "nixl-cu13==$(echo ${NIXL_VERSION} | sed 's/^v//')"
+
+# --- UCCL-EP + deep_ep wrapper ----------------------------------------------
+RUN git clone https://github.com/uccl-project/uccl.git /opt/uccl \
+    && cd /opt/uccl \
+    && git checkout ${UCCL_COMMIT} \
+    && git submodule update --init --recursive
+
+RUN uv pip install --system nanobind
+
+WORKDIR /opt/uccl
+RUN uv pip install --system .
+
+# UCCL-EP uses SM90+ features (TMA, mbarrier); avoid SM80 in this build.
+WORKDIR /opt/uccl/ep
+ENV PER_EXPERT_BATCHING=1
+RUN TORCH_CUDA_ARCH_LIST="9.0a+PTX;10.0a+PTX;10.3a+PTX" python3 setup.py install
+
+WORKDIR /opt/uccl/ep/deep_ep_wrapper
+RUN pip install --no-deps .
+
+# --- smoke tests -------------------------------------------------------------
+RUN echo "=== Smoke test: vLLM ===" \
+    && python3 -c "import vllm; print(f'vLLM {vllm.__version__}')" \
+    && echo "=== Smoke test: NIXL ===" \
+    && python3 -c "import nixl; print('NIXL OK')" \
+    && echo "=== Smoke test: uccl.ep ===" \
+    && python3 -c "import torch; import uccl.ep; print('uccl.ep OK')" \
+    && echo "=== Smoke test: deep_ep ===" \
+    && python3 -c "import torch; from deep_ep import Buffer, Config; print('deep_ep OK')" \
+    && echo "=== Smoke test: vLLM has_deep_ep ===" \
+    && python3 -c "from vllm.utils.import_utils import has_deep_ep; assert has_deep_ep(); print('vLLM has_deep_ep() = True')" \
+    && echo "=== Smoke test: Ray ===" \
+    && python3 -c "import ray; print(f'Ray {ray.__version__}')" \
+    && echo "=== All smoke tests passed ==="
+
+WORKDIR /workspace
--- a/3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
+++ b/3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
@@ -1,3 +1,8 @@
+<!--
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+-->
+
 # DeepSeek-V3 Disaggregated Inference with vLLM, UCCL-EP, and NIXL on EKS
 
 This sample shows how to serve **DeepSeek-V3-0324** (671B parameters, 256 experts,
@@ -11,13 +16,18 @@ top-8 routing, MLA) with disaggregated **prefill / decode** across multiple
   transfer between prefill and decode pods
 - A simple Python proxy (`toy_proxy_server.py` shipped in vLLM) for routing
 
-It is the deployable version of the experiment
-[`experiments/2026-05-06-dsv3-disagg-mtp`](https://github.com/awslabs/awsome-distributed-ai)
-that benchmarked configs from a single 8-GPU unified node up to a 6-node
+It benchmarks configs from a single 8-GPU unified node up to a 6-node
 1-prefill / 5-decode disaggregated topology, plus a UCCL-EP vs. AGRS
 (`allgather_reducescatter`) backend comparison.
 
-## What you can reproduce
+## Indicative results
+
+> **Caveat.** The numbers below come from single-seed runs at
+> `--num-prompts 50`. They are smoke-benchmark numbers, not statistically
+> tight measurements — re-run with the multi-seed loop in
+> [`recipe/benchmark.sh`](recipe/benchmark.sh) and compute confidence
+> intervals before quoting them anywhere downstream. Treat the rankings
+> below as directionally informative, not as research findings.
 
 | Config       | Nodes | GPUs | Burst RPS | Burst tok/s | P50 TPOT @1RPS | P50 TPOT @burst | tok/s/GPU |
 | ------------ | :---: | :--: | --------: | ----------: | -------------: | --------------: | --------: |
@@ -28,20 +38,20 @@ that benchmarked configs from a single 8-GPU unified node up to a 6-node
 | 1P+5D       |   6   |  48  |     17.16 |       4,392 |        17.81ms |         30.34ms |     91.51 |
 | 1P+3D AGRS  |   4   |  32  |     10.62 |       2,721 |        23.30ms |         48.20ms |     85.03 |
 
-(Workload: `--random-input-len 1024 --random-output-len 256 --num-prompts 50` via
-`vllm bench serve`.)
+Workload: `--random-input-len 1024 --random-output-len 256 --num-prompts 50`
+via `vllm bench serve`.
 
-Headline findings:
+Directional observations:
 
-1. **Disaggregation pays off mostly in latency, not throughput.** A single
-   unified node is the best `tok/s/GPU` config; disaggregation lowers burst
-   TPOT from 44ms → 30ms.
-2. **The optimal P:D ratio is 1:4.** Beyond that, the single prefill node is
-   the hard bottleneck.
-3. **UCCL-EP (`deepep_high_throughput` + `deepep_low_latency`) beats AGRS by
-   ~62%** on burst RPS and ~30% on TPOT for this MLA + 256-expert MoE model.
-
-See `conclusion.md` in the upstream experiment for the full analysis.
+1. **Disaggregation pays off mostly in latency, not throughput.** Unified
+   has the best `tok/s/GPU`; disaggregation lowers burst TPOT roughly
+   44ms → 30ms but only modestly increases peak RPS.
+2. **The 1P:4D ratio appears to be the knee** — beyond that, the single
+   prefill node looks like the bottleneck.
+3. **UCCL-EP (`deepep_high_throughput` + `deepep_low_latency`) outperforms
+   AGRS** for this MLA + 256-expert MoE model in our runs (≈30% TPOT
+   improvement is consistent with public UCCL-EP benchmarks; the burst-RPS
+   delta is from a single-seed run and warrants reproduction).
 
 ## Architecture
 
@@ -79,7 +89,9 @@ See `conclusion.md` in the upstream experiment for the full analysis.
 
 ### Cluster
 
-- Amazon EKS or SageMaker HyperPod EKS, version 1.28+
+- Amazon EKS or SageMaker HyperPod EKS, version **1.33+** (1.32 and below
+  are out of EKS standard support; 1.33 added native EFA traffic in the
+  default EKS security group)
 - 1–6 × `p5en.48xlarge` GPU nodes (8×H200 141GB, 16×EFA, ~30 TB NVMe)
 - NVIDIA device plugin and EFA device plugin installed
 - (Recommended) FSx for Lustre PVC for the model cache, or local NVMe under
@@ -89,13 +101,13 @@ See `conclusion.md` in the upstream experiment for the full analysis.
 
 | Component | Version |
 | --- | --- |
-| Kubernetes | 1.28+ (EKS / HyperPod EKS tested) |
-| kubectl | 1.28+ |
+| Kubernetes | 1.33+ (EKS / HyperPod EKS tested) |
+| kubectl | 1.33+ |
 | Docker / BuildKit | 24.0+ |
 | AWS CLI v2 | latest |
 | vLLM | 0.21.0 |
 | UCCL | commit `0dc87eb` |
-| NIXL | v1.0.1 |
+| NIXL | v1.1.0 |
 | PyTorch | 2.11.0 (cu130) |
 | Ray | 2.55.1 |
 | EFA installer | 1.48.0 |
@@ -111,23 +123,22 @@ See `conclusion.md` in the upstream experiment for the full analysis.
 ## Repository layout
 
 ```
-dsv3-disagg/
-├── Dockerfile                  # vLLM 0.21.0 + NIXL 1.0.1 + UCCL-EP, CUDA 13
+dsv3-uccl-nixl/
+├── Dockerfile                  # vLLM 0.21.0 + NIXL 1.1.0 + UCCL-EP, CUDA 13
 ├── README.md                   # this file
 ├── setup/
 │   ├── env_vars.example        # cluster + image + role config
 │   ├── build-push.sh           # docker build + ECR push
 │   └── install-prereqs.sh      # NVIDIA + EFA device plugins (optional)
 ├── manifests/
-│   ├── prefill.yaml            # 1× prefill pod (deepep_high_throughput)
-│   ├── decode.yaml             # 1× decode pod template (deepep_low_latency)
-│   ├── proxy.yaml              # router pod (toy_proxy_server.py)
-│   ├── unified.yaml            # 1-node baseline (no NIXL, deepep_low_latency)
-│   └── hf-token-secret.yaml    # HuggingFace token secret stub
+│   ├── prefill.yaml            # prefill Deployment (deepep_high_throughput)
+│   ├── decode.yaml             # decode Deployment template (deepep_low_latency)
+│   ├── proxy.yaml              # router Deployment (toy_proxy_server.py)
+│   └── unified.yaml            # 1-node baseline Deployment (no NIXL)
 └── recipe/
     ├── deploy.sh               # render + apply manifests for a chosen topology
-    ├── teardown.sh             # delete pods
-    └── benchmark.sh            # vllm bench serve rate sweep + warmup
+    ├── teardown.sh             # delete Deployments
+    └── benchmark.sh            # vllm bench serve rate sweep + warmup, multi-seed
 ```
 
 ## Quick start
@@ -137,7 +148,8 @@ dsv3-disagg/
 ```bash
 cd 3.test_cases/pytorch/vllm/dsv3-uccl-nixl
 cp setup/env_vars.example setup/env_vars
-$EDITOR setup/env_vars   # set ECR registry, HF_TOKEN, node hostnames, IPs
+$EDITOR setup/env_vars              # replace every REPLACE_ME placeholder
+grep REPLACE_ME setup/env_vars      # should print nothing when fully filled
 source setup/env_vars
 ```
 
@@ -154,7 +166,8 @@ The image bundles:
 
 - vLLM 0.21.0 (built from source for `sm_80 sm_86 sm_89 sm_90 sm_100 sm_103`)
 - UCCL-EP with the `deep_ep` drop-in wrapper (so vLLM auto-detects it)
-- NIXL 1.0.1 (`nixl-cu13`) with the `LIBFABRIC` backend
+- NIXL 1.1.0 (CUDA-13 backend auto-selected from the meta wheel) with the
+  `LIBFABRIC` backend
 - Ray 2.55.1 (DP coordinator)
 - EFA installer 1.48.0, NCCL 2.30.4, AWS OFI NCCL plugin
 - GDRCopy 2.5.2
@@ -221,14 +234,23 @@ curl -s "http://${PREFILL_IP}:8000/v1/completions" \
 ### 7. Benchmark
 
 `recipe/benchmark.sh` runs a warm-up then a rate sweep at 1, 4, 8, 16, and
-`inf` RPS using `vllm bench serve`:
+`inf` RPS using `vllm bench serve`, looping over three seeds (1234, 5678,
+9012) per rate so you can aggregate over multiple runs:
 
 ```bash
 ./recipe/benchmark.sh 1p4d
 ```
 
-Results land in `results/<topology>/openai-<rate>qps-*.json`. Key fields:
-`request_throughput`, `output_throughput`, `p50_tpot_ms`, `p50_ttft_ms`.
+Results land in `results/<topology>/seed-<seed>/openai-<rate>qps-*.json`.
+Key fields per JSON: `request_throughput`, `output_throughput`,
+`p50_tpot_ms`, `p50_ttft_ms`. Aggregate across the three seed dirs and
+report a confidence interval before quoting any number externally.
+
+Override the seeds, rates, or prompt count with environment variables:
+
+```bash
+SEEDS="1234" RATES="1 inf" NUM_PROMPTS=200 ./recipe/benchmark.sh 1p4d
+```
 
 ### 8. Tear down
 
@@ -276,10 +298,18 @@ For DeepSeek-V3 always use `deepep_high_throughput` on prefill and
                        "kv_connector_extra_config":{"backends":["LIBFABRIC"]}}'
 ```
 
-`kv_role: kv_both` lets vLLM act as either sender or receiver based on the
-incoming `do_remote_decode` / `do_remote_prefill` request flag — the proxy
-sets these. `LIBFABRIC` uses EFA via the libfabric provider; alternatives
-like `UCX` are not yet recommended on EFA without GPUDirect RDMA tuning.
+In vLLM 0.21.0, `kv_role` is a required-but-ignored placeholder for
+`NixlConnector` — the [vLLM NixlConnector docs](https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html)
+state that "the actual prefiller/decoder roles are determined by the
+upper-level proxy. Therefore, `kv_role` in `--kv-transfer-config` is
+effectively a placeholder and does not affect NixlConnector's behavior."
+What actually drives prefill vs. decode behavior is the proxy injecting
+`do_remote_prefill` / `do_remote_decode` into per-request `kv_transfer_params`.
+The config validator just refuses to start without *some* `kv_role`; we set
+`kv_both` for that reason only.
+
+`LIBFABRIC` uses EFA via the libfabric provider; alternatives like `UCX`
+are not yet recommended on EFA without GPUDirect RDMA tuning.
 
 `VLLM_NIXL_SIDE_CHANNEL_PORT=5600` and `VLLM_NIXL_SIDE_CHANNEL_HOST=$MY_IP`
 expose the metadata exchange channel on the host network.
@@ -288,11 +318,16 @@ expose the metadata exchange channel on the host network.
 
 The proxy uses
 `/opt/vllm/tests/v1/kv_connector/nixl_integration/toy_proxy_server.py`
-which ships with the vLLM source tree. It does simple round-robin over
-`--decoder-hosts` and routes the prefill phase to `--prefiller-hosts`.
+which ships in the vLLM source tree. The vLLM NixlConnector docs direct
+users to invoke it from that path, but it lives under `tests/` and is named
+`toy_*` — upstream is free to move or delete it on any release. The image
+pins `VLLM_COMMIT=v0.21.0`, so the path is stable for this sample, but
+verify it before bumping that ARG.
 
-For production traffic shaping you will want a smarter router — this proxy
-exists to demonstrate the wire protocol, not to be a load balancer.
+It does simple round-robin over `--decoder-hosts` and routes the prefill
+phase to `--prefiller-hosts`. For production traffic shaping you will want
+a smarter router — this proxy exists to demonstrate the wire protocol, not
+to be a load balancer.
 
 ### Storage
 
@@ -320,7 +355,7 @@ and set `HF_HOME` accordingly. Note: ~680 GB FP8 weights load in
 | `ACCOUNT` | `123456789012` | AWS account ID for ECR |
 | `REGISTRY` | `${ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/` | ECR registry |
 | `IMAGE` | `vllm-uccl-ep` | ECR repo name |
-| `TAG` | `vllm0.21.0-uccl-0dc87eb` | image tag |
+| `TAG` | `vllm0.21.0-uccl-0dc87eb-cu13` | image tag |
 | `NAMESPACE` | `dsv3-disagg` | Kubernetes namespace |
 | `INSTANCE_TYPE` | `p5en.48xlarge` | node selector value |
 | `MODEL` | `deepseek-ai/DeepSeek-V3-0324` | HF model id |
@@ -379,23 +414,39 @@ kubectl logs <pod-name> -n "$NAMESPACE" --tail=80
 CUDA-graph capture can take 2–5 minutes after `/health` returns 200. Be
 patient on the first benchmark.
 
-## Cost estimate (us-west-2 on-demand, May 2026)
+## Cost
 
-| Config | Nodes | Cost / hr | Cost per 1h benchmark |
-| --- | --- | --- | --- |
-| 1N | 1 | $98.32 | $98.32 |
-| 1P+1D | 2 | $196.64 | $196.64 |
-| 1P+3D | 4 | $393.28 | $393.28 |
-| 1P+4D | 5 | $491.60 | $491.60 |
-| 1P+5D | 6 | $589.92 | $589.92 |
+GPU instance pricing changes too often (and varies by region, term, and
+Capacity Block) for a baked-in table to age well. Check the live source:
 
-A full sweep across all topologies (one rate-sweep each) takes 4–6 hours
-end to end including model download.
+- [On-Demand pricing](https://aws.amazon.com/ec2/pricing/on-demand/)
+- [Capacity Block pricing](https://aws.amazon.com/ec2/capacityblocks/pricing/)
+- Or query the AWS pricing API:
+  ```bash
+  aws pricing get-products --service-code AmazonEC2 \
+    --filters Type=TERM_MATCH,Field=instanceType,Value=p5en.48xlarge \
+              Type=TERM_MATCH,Field=location,Value="US West (Oregon)"
+  ```
+
+As a rough sense of scale: the largest topology in this sample uses 6
+nodes, and a full rate sweep across topologies takes 4–6 hours end to end
+(including the first-time ~680 GB model download).
+
+## License
+
+This sample is released under the [MIT-0 License](../../../../LICENSE), the
+same license as the rest of `awsome-distributed-ai`.
 
 ## References
 
+- vLLM NixlConnector usage:
+  https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html
 - vLLM disaggregated serving:
   https://github.com/vllm-project/vllm/blob/main/docs/source/serving/disaggregated_serving.md
 - UCCL-EP: https://github.com/uccl-project/uccl
 - NIXL: https://github.com/ai-dynamo/nixl
+- NIXL libfabric (EFA) plugin:
+  https://github.com/ai-dynamo/nixl/blob/main/src/plugins/libfabric/README.md
 - DeepSeek-V3 paper: https://arxiv.org/abs/2412.19437
+- AWS EFA cheat sheet:
+  https://github.com/awslabs/awsome-distributed-ai/blob/main/architectures/efa-cheatsheet.md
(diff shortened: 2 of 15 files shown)


3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
```diff
@@ -0,0 +1,186 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
```
KeitaW: ### SPDX license header on Dockerfile is Apache-2.0, not MIT-0

I'd flip this to MIT-0 to match the repo's contribution policy — the Dockerfile is the only file in this PR that carries `Apache-2.0`, which I think slipped in from an upstream template.

```suggestion
# SPDX-License-Identifier: MIT-0
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+- Hugging Face token with access to `deepseek-ai/DeepSeek-V3-0324`
+- Container registry (e.g. ECR) writable from your build host
+- `kubectl` configured for your cluster
+- Around ~680 GB of fast storage for the FP8 weights cache
+
+## Repository layout
+
+```
+dsv3-disagg/
```
KeitaW: ### Repo layout diagram in README uses the wrong directory name

The layout diagram shows the root as `dsv3-disagg/` but the actual directory is `dsv3-uccl-nixl/`. Same name drift appears in the PR description's "Directory Structure" section. I'd rename to match the on-disk path so a copy-paste `cd` works.

```suggestion
dsv3-uccl-nixl/
```

KeitaW (commented):

Hi @pbelevich — thanks for putting this together. This is a big, well-organized contribution and the per-pod parallelism / NIXL wiring is unusually clear. I've split my comments into five themed batches so each one is reviewable on its own:

1. **Structure & Repository Hygiene** (this batch)
2. Deployment Pipeline & K8s Operational Correctness
3. Infrastructure, NCCL & Container Security
4. Documentation Consistency
5. Evaluation Methodology Validity

A handful of findings span the whole new directory rather than a single line, so I'm putting them in this body:

---

### Most new files are missing the MIT-0 header entirely

**Files:** all five `manifests/*.yaml`, all three `recipe/*.sh`, both `setup/*.sh`, `setup/env_vars.example`, `.gitignore`, both `README.md` files

Only the Dockerfile has any header (and it's the wrong one — see the inline comment on `Dockerfile` line 2). Since the PR's checklist box for "I have read the contributing guidelines" is ticked, I assume this was an oversight rather than a deliberate choice. Could you add the canonical two-line header at the top of each shell script, YAML manifest, and env-vars file? For YAML, the header goes before any `---` separator.

---

### Placement under `3.test_cases/pytorch/vllm/` vs. RFC #1056

[RFC #1056](https://github.com/awslabs/awsome-distributed-ai/issues/1056) proposes that inference samples land under `examples/inference/<library>/<model>/` — for this PR that would be `examples/inference/vllm/deepseek-v3-disagg/`. The current location reuses the legacy `3.test_cases/pytorch/` tree (which CLAUDE.md describes as "distributed training examples by framework"), so an inference sample doesn't quite fit there.

I noticed the repo doesn't have an `examples/` directory yet, so I don't think this PR can singlehandedly adopt the new layout. Would it be worth either (a) calling that out in the PR description as a known interim placement, or (b) coordinating with whoever is driving #1056 so this lands in the new tree on day one? Happy to defer if you've already discussed this with maintainers.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/setup/install-prereqs.sh
```diff
@@ -0,0 +1,27 @@
...
+NVIDIA_PLUGIN_VERSION="${NVIDIA_PLUGIN_VERSION:-v0.16.2}"
+EFA_PLUGIN_VERSION="${EFA_PLUGIN_VERSION:-v0.5.7}"
+
+echo "==> Installing NVIDIA Kubernetes device plugin ${NVIDIA_PLUGIN_VERSION}"
+kubectl apply -f \
+    "https://raw.githubusercontent.com/NVIDIA/k8s-device-plugin/${NVIDIA_PLUGIN_VERSION}/deployments/static/nvidia-device-plugin.yml"
+
+echo "==> Installing AWS EFA Kubernetes device plugin ${EFA_PLUGIN_VERSION}"
+kubectl apply -f \
+    "https://raw.githubusercontent.com/aws-samples/aws-efa-eks/main/manifest/efa-k8s-device-plugin.yml"
```
KeitaW: ### Unpinned upstream `main` reference for EFA device plugin

The NVIDIA device plugin is pinned with `${NVIDIA_PLUGIN_VERSION}` and that's great. The AWS EFA plugin URL goes to `aws-samples/aws-efa-eks/main/manifest/efa-k8s-device-plugin.yml`, though — `main` will silently shift if upstream pushes a breaking change to the manifest. I'd pin the EFA plugin URL to a tag or commit SHA the same way you do for the NVIDIA one (the `${EFA_PLUGIN_VERSION}` ARG is already declared on line 8, it just isn't wired into the URL).

```suggestion
    "https://raw.githubusercontent.com/aws-samples/aws-efa-eks/${EFA_PLUGIN_VERSION}/manifest/efa-k8s-device-plugin.yml"
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
```diff
@@ -0,0 +1,186 @@
...
+        -r requirements/cuda.txt \
+        --extra-index-url https://download.pytorch.org/whl/cu130 \
+    && python3 setup.py bdist_wheel --dist-dir=dist --py-limited-api=cp38 \
+    && uv pip install --system --index-strategy unsafe-best-match \
+        dist/*.whl \
+        --extra-index-url https://download.pytorch.org/whl/cu130
+
+# --- Ray (DP coordinator) ----------------------------------------------------
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --system "ray[default]>=2.43"
```
KeitaW: ### Ray pin is unbounded and disagrees with documented version

The Dockerfile installs `"ray[default]>=2.43"` (no upper bound), but the README's Software table and the PR description both claim Ray 2.55.1. Whatever PyPI happens to serve at build time will end up in the image, so reproducibility drifts away from the documented version. I'd pin to an exact release that matches the README:

```suggestion
    uv pip install --system "ray[default]==2.55.1"
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
```diff
@@ -0,0 +1,157 @@
...
+        - |
+          set -euo pipefail
+
+          echo "=== DSv3 DECODE ${DECODE_INDEX} (TP=${TENSOR_PARALLEL_SIZE}, DP=${DATA_PARALLEL_SIZE}, EP=8, ${DECODE_BACKEND}, NixlConnector) ==="
+          MY_IP=$(hostname -I | awk '{print $1}')
+          echo "MY_IP: $MY_IP"
+
+          IFACE=$(ip -o -4 addr show | grep "$MY_IP" | awk '{print $2}' | head -1)
+          export GLOO_SOCKET_IFNAME=$IFACE
+          export NCCL_SOCKET_IFNAME=$IFACE
```
KeitaW: ### Runtime `NCCL_SOCKET_IFNAME=$IFACE` overrides the exclusion pattern

The Dockerfile correctly sets `ENV NCCL_SOCKET_IFNAME=^docker,lo,veth` (exclusion pattern, good). However, every pod's `command:` then does:

```bash
IFACE=$(ip -o -4 addr show | grep "$MY_IP" | awk '{print $2}' | head -1)
export NCCL_SOCKET_IFNAME=$IFACE
```

This silently swaps the exclusion pattern for a single positive interface name (typically `eth0` or `ens5`). On `p5en.48xlarge` that hides the EFA interfaces from NCCL — bootstrap still completes, but you lose the EFA fabric. That matches your README's troubleshooting note that "the same workload runs ~28× slower" with the EFA fallback.

I'd drop the `NCCL_SOCKET_IFNAME` export entirely (the Dockerfile env is fine) and keep only the `GLOO_SOCKET_IFNAME=$IFACE` line — Gloo is fine with a positive selection because it isn't on the EFA hot path.

```suggestion
          export NCCL_SOCKET_IFNAME=^lo,docker,veth
```

The same fix applies to `manifests/prefill.yaml` (line 131) and `manifests/unified.yaml` (line 117). See [EFA cheatsheet](https://github.com/awslabs/awsome-distributed-ai/blob/main/architectures/efa-cheatsheet.md).

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/prefill.yaml
```diff
@@ -0,0 +1,156 @@
...
+        - |
+          set -euo pipefail
+
+          echo "=== DSv3 PREFILL (TP=${TENSOR_PARALLEL_SIZE}, DP=${DATA_PARALLEL_SIZE}, EP=8, ${PREFILL_BACKEND}, NixlConnector) ==="
+          MY_IP=$(hostname -I | awk '{print $1}')
+          echo "MY_IP: $MY_IP"
+
+          IFACE=$(ip -o -4 addr show | grep "$MY_IP" | awk '{print $2}' | head -1)
+          export GLOO_SOCKET_IFNAME=$IFACE
+          export NCCL_SOCKET_IFNAME=$IFACE
```
KeitaW: ### Same `NCCL_SOCKET_IFNAME` override in prefill manifest

See the longer write-up on `decode.yaml` line 132 — same fix here.

```suggestion
          export NCCL_SOCKET_IFNAME=^lo,docker,veth
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/unified.yaml
```diff
@@ -0,0 +1,139 @@
...
+        - bash
+        - -c
+        - |
+          set -euo pipefail
+
+          echo "=== DSv3 UNIFIED (TP=${TENSOR_PARALLEL_SIZE}, DP=${DATA_PARALLEL_SIZE}, EP=8, ${DECODE_BACKEND}) ==="
+          MY_IP=$(hostname -I | awk '{print $1}')
+          IFACE=$(ip -o -4 addr show | grep "$MY_IP" | awk '{print $2}' | head -1)
+          export GLOO_SOCKET_IFNAME=$IFACE
+          export NCCL_SOCKET_IFNAME=$IFACE
```
KeitaW: ### Same `NCCL_SOCKET_IFNAME` override in unified manifest

See the longer write-up on `decode.yaml` line 132 — same fix here.

```suggestion
          export NCCL_SOCKET_IFNAME=^lo,docker,veth
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/hf-token-secret.yaml
```diff
@@ -0,0 +1,9 @@
+# Apply with: envsubst < manifests/hf-token-secret.yaml | kubectl apply -f -
```
KeitaW: ### `hf-token-secret.yaml` exists but the README never uses it

The manifest file is shipped, but the Quick Start in the README (step 3) tells the reader to create the secret imperatively:

```bash
kubectl create secret generic hf-token --namespace "$NAMESPACE" \
    --from-literal=HF_TOKEN="$HF_TOKEN"
```

Neither `recipe/deploy.sh` nor `setup/install-prereqs.sh` references the manifest either. I'd either drop the YAML or replace the imperative `kubectl create secret` block in the README with `envsubst < manifests/hf-token-secret.yaml | kubectl apply -f -` so the file is actually wired in. The manifest version also has the nice property of being idempotent (apply replaces).

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
```diff
@@ -0,0 +1,157 @@
...
+        - name: FI_EFA_FORK_SAFE
+          value: "1"
+        - name: UCCL_SOCKET_IFNAME
+          value: "^lo,docker0,veth_def_agent"
+        - name: VLLM_NIXL_SIDE_CHANNEL_PORT
+          value: "5600"
+        - name: NCCL_DEBUG
+          value: "WARN"
+      resources:
+        requests:
```
KeitaW: ### No `livenessProbe` on long-running vLLM pods

vLLM can deadlock (NCCL stalls, CUDA-graph capture stuck) and continue answering `/health` from the API server thread while the actual generator hangs — the pod stays `Ready` forever without a probe to evict it. Could you add a `livenessProbe` that exercises a real generation (or at least an `/v1/models` call that round-trips through the engine)? The same applies to `prefill.yaml` and `unified.yaml`.

A minimal start (hitting `/health`, then upgrade to a `/v1/completions` probe once latency is characterised):

```yaml
livenessProbe:
  httpGet:
    path: /health
    port: 8200
  initialDelaySeconds: 1800   # cold start: model download + CUDA graphs
  periodSeconds: 60
  failureThreshold: 5
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
```diff
@@ -0,0 +1,157 @@
...
+              app: dsv3-disagg
+          topologyKey: kubernetes.io/hostname
+  tolerations:
+    - key: nvidia.com/gpu
+      operator: Exists
+      effect: NoSchedule
+    - key: vpc.amazonaws.com/efa
+      operator: Exists
+      effect: NoSchedule
+    - key: sagemaker.amazonaws.com/node-health-status
```
KeitaW: ### `sagemaker.amazonaws.com/node-health-status` blanket toleration

`operator: Exists` with no `value:` tolerates *every* value of `sagemaker.amazonaws.com/node-health-status`, including SchedulingDisabled, NotReady, and the various failure-injection taints HyperPod applies when it has detected a hardware issue. The intent is presumably to tolerate the "healthy" placeholder taint that HyperPod attaches; I'd scope to the specific value(s) you actually want to allow (e.g., `value: "Schedulable"` or whatever your HyperPod fleet attaches).

Same toleration block in `prefill.yaml`, `unified.yaml`, and `proxy.yaml`.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
```diff
@@ -0,0 +1,157 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: dsv3-decode-${DECODE_INDEX}
+  namespace: ${NAMESPACE}
+  labels:
+    app: dsv3-disagg
+    role: decode
+spec:
+  restartPolicy: Never
```
KeitaW: ### Bare `Pod` + `restartPolicy: Never` means a single OOM permanently kills the deployment

For an inference workload this is unusual — typically you'd use a `Deployment` (or LeaderWorkerSet for multi-pod) so the control plane re-creates the pod after a crash. With `Pod` + `Never`, any vLLM crash leaves the pod in a terminated state and `recipe/deploy.sh` has to be re-run manually. Is the choice deliberate (e.g., to capture the exact failure state for debugging during the benchmark)? If so, a one-line comment in the YAML explaining "Pod + Never on purpose so failures are inspectable" would save the next reader from guessing. If not, a `Deployment` with `replicas: 1` would behave more like operators expect.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
```diff
@@ -0,0 +1,157 @@
...
+      emptyDir:
+        medium: Memory
+        sizeLimit: 120Gi
+    - name: model-cache
+      hostPath:
+        path: /mnt/k8s-disks/0
+        type: DirectoryOrCreate
+  containers:
+    - name: vllm
+      image: ${REGISTRY}${IMAGE}:${TAG}
```
KeitaW: ### `imagePullPolicy` and `terminationGracePeriodSeconds` not set

Two small operational nits:

1. `imagePullPolicy` defaults to `Always` only when the tag is `:latest`; with your pinned tag K8s will fall back to `IfNotPresent` which is what you want — but a sibling reviewer might miss this implicit behavior. Worth setting `imagePullPolicy: IfNotPresent` explicitly.
2. With ~680 GB of weight cache in memory and active EFA RDMA, the default 30s grace period almost certainly isn't enough for clean shutdown. I'd bump `terminationGracePeriodSeconds: 120` so NIXL has time to drain.

KeitaW (commented):
## Review Batch 2/5 — Deployment Pipeline & K8s Operational Correctness

The biggest one in this batch is the `NCCL_SOCKET_IFNAME` runtime override — the pod command undoes the Dockerfile's exclusion pattern with a positive selection that excludes EFA interfaces. That matches the troubleshooting note about "28× slower with TCP fallback" so I think it's a real production-impact bug, not a style nit.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/proxy.yaml
```diff
@@ -0,0 +1,76 @@
...
+            echo "Waiting for $endpoint..."
+            until curl -sf "http://${endpoint}/health" >/dev/null 2>&1; do
+              sleep 10
+              echo "  still waiting for $endpoint..."
+            done
+            echo "  $endpoint is ready!"
+          done
+          echo "All instances ready!"
+
+          exec python3 /opt/vllm/tests/v1/kv_connector/nixl_integration/toy_proxy_server.py \
```
KeitaW: ### Proxy is path-fragile and has no liveness signal of its own

You already acknowledge this in the README ("For production traffic shaping you will want a smarter router — this proxy exists to demonstrate the wire protocol"), and I confirmed that `tests/v1/kv_connector/nixl_integration/toy_proxy_server.py` is in fact the *only* proxy v0.21.0 ships — the official [`docs/features/nixl_connector_usage.md`](https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html) directs users to invoke it from that path. So this isn't a "test fixture being misused" situation. Two operational consequences still worth a short inline comment, though:

1. Even though it's the documented entry point, it lives under `tests/` and is named `toy_*` — upstream is free to move or delete it on any release. A `${VLLM_COMMIT}` bump could silently break the sample. A pinned-by-tag reference in the README (and a tracked follow-up issue for "swap in a permanent proxy once vLLM ships one") would make the dependency explicit.
2. There's no `livenessProbe` on the proxy pod, so a wedged proxy stays `Ready` and the prefill/decode pods sit idle.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+- **NIXL** transfers KV blocks from prefill → decode over EFA via
+  `LIBFABRIC` backend
+- All pods run with `hostNetwork: true` so EFA / NIXL / NCCL can use the
+  full 16-EFA, 3,200 Gbps fabric
+
+## Prerequisites
+
+### Cluster
+
+- Amazon EKS or SageMaker HyperPod EKS, version 1.28+
```
KeitaW: ### EKS version floor of 1.28 is six months past end-of-extended-support

I checked the current [EKS version calendar](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html). As of today, K8s 1.28 has been entirely out of EKS support since **November 26, 2025** — any cluster on 1.28 was force-upgraded automatically. The floor in the README is effectively unreachable.

Current EKS state for context:
- **Standard support (free):** 1.33, 1.34, 1.35
- **Extended support ($0.60/cluster/hr):** 1.30, 1.31, 1.32
- **Out of support entirely:** 1.29 and earlier

I'd bump the floor to **1.34 or 1.35** rather than the oldest extended-support version: it stays in standard support through July 2026, hits extended support through July 2027 (covers the typical 6–12 month sample shelf life), and 1.33 specifically adds native EFA traffic in the default EKS security group — directly relevant to a p5en sample. (Same drift in the Software table at line 92.)

```suggestion
- Amazon EKS or SageMaker HyperPod EKS, version 1.33+
```

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
```diff
@@ -0,0 +1,186 @@
...
+
+# --- NIXL (cu13) -------------------------------------------------------------
+# Install nixl-cu13 first, then nixl meta package, then force-reinstall
+# nixl-cu13 to keep cu13 ownership of nixl_ep namespace.
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --system "nixl-cu13==$(echo ${NIXL_VERSION} | sed 's/^v//')" \
+    && uv pip install --system "nixl==$(echo ${NIXL_VERSION} | sed 's/^v//')" \
+    && uv pip install --system --reinstall --no-deps "nixl-cu13==$(echo ${NIXL_VERSION} | sed 's/^v//')"
```
KeitaW: ### NIXL triple-install dance is unnecessary in v1.1.0+ — recommend bumping

I dug into why the triple install is needed in v1.0.1 — it's a real bug, not a code smell. The `nixl==1.0.1` meta-package has an **unconditional** `nixl-cu12>=1.0.1` dep, even when you request `nixl[cu13]`. So step 2 (the meta install) drags cu12 in alongside your cu13 backend; step 3 (`--reinstall --no-deps nixl-cu13`) re-pins cu13 as the winning backend in the same install slot.

The good news: **[NIXL v1.1.0](https://github.com/ai-dynamo/nixl/releases/tag/v1.1.0) (released 2026-05-12) fixes this**. The release notes state directly: *"Downstream consumers should also switch their `requirements.txt` from `nixl[cu12]` / `nixl[cu13]` to plain `nixl` — the meta wheel now bundles both CUDA backends and auto-selects at runtime based on `torch.version.cuda`."* The PyPI metadata confirms the shape change: in [`nixl==1.0.1`](https://pypi.org/pypi/nixl/1.0.1/json), `requires_dist` is `nixl-cu12>=1.0.1` (unconditional — the bug), while [`nixl==1.1.0`](https://pypi.org/pypi/nixl/1.1.0/json) declares `nixl-cu12==1.1.0` and `nixl-cu13==1.1.0` both unconditional so the meta wheel pulls in both backends and runtime selects. Plain `pip install nixl==1.1.0` is sufficient and the install block here collapses to one line.

```suggestion
    uv pip install --system "nixl==$(echo ${NIXL_VERSION} | sed 's/^v//')"
```

(And bump `ARG NIXL_VERSION=v1.0.1` to `ARG NIXL_VERSION=v1.1.0` near the top of the Dockerfile.)

If you want to stay on v1.0.1 for some other reason, please leave a comment in the Dockerfile explaining what each of the three steps is defending against (the `# keep cu13 ownership of nixl_ep namespace` is a start but the unconditional cu12 dep is the actual root cause). Also worth adding a build-time assert like `python3 -c "import nixl; assert 'cu13' in nixl.__file__"` — the existing `import nixl` smoke test won't catch the wrong CUDA variant getting selected.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
```diff
@@ -0,0 +1,186 @@
...
+WORKDIR /opt/uccl
+RUN uv pip install --system .
+
+# UCCL-EP uses SM90+ features (TMA, mbarrier); avoid SM80 in this build.
+WORKDIR /opt/uccl/ep
+ENV PER_EXPERT_BATCHING=1
+RUN TORCH_CUDA_ARCH_LIST="9.0a+PTX;10.0a+PTX;10.3a+PTX" python3 setup.py install
+
+WORKDIR /opt/uccl/ep/deep_ep_wrapper
+RUN pip install --no-deps .
```
KeitaW: ### `pip install --no-deps .` in `deep_ep_wrapper` step

`--no-deps` skips dependency resolution entirely. If the wrapper grows a runtime dependency in a future UCCL commit, the build will succeed but `import deep_ep` will fail at pod start (caught only by the smoke-test block further down — good that it exists). I'd either drop `--no-deps` (the wrapper presumably depends on `uccl.ep` which is already installed) or leave a comment stating which dependency is intentionally being skipped and why.

KeitaW (commented):
## Review Batch 3/5 — Infrastructure, NCCL & Container Security

The two highest-impact items here are the EKS 1.28 floor (six months out of support) and the NIXL triple-install dance (real packaging bug in v1.0.1; v1.1.0 fixes it with a one-line install).

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+### NIXL KV transfer
+
+```yaml
+--kv-transfer-config '{"kv_connector":"NixlConnector",
+                       "kv_role":"kv_both",
+                       "kv_load_failure_policy":"fail",
+                       "kv_connector_extra_config":{"backends":["LIBFABRIC"]}}'
+```
+
+`kv_role: kv_both` lets vLLM act as either sender or receiver based on the
```
KeitaW: ### README's `kv_role` explanation contradicts vLLM 0.21.0 docs

The README says: *"`kv_role: kv_both` lets vLLM act as either sender or receiver based on the incoming `do_remote_decode` / `do_remote_prefill` request flag."* The second half (about the request flags) is right, but the first half is misleading — vLLM 0.21.0's [`docs/features/nixl_connector_usage.md`](https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html) explicitly states: *"NixlConnector currently does not distinguish kv_role; the actual prefiller/decoder roles are determined by the upper-level proxy. Therefore, kv_role in `--kv-transfer-config` is effectively a placeholder and does not affect NixlConnector's behavior."*

So `kv_role` is a required-but-ignored field — the config validator raises ValueError if it's unset, but the connector itself ignores the value. What actually drives P vs. D behaviour is the proxy injecting `do_remote_prefill` / `do_remote_decode` into the per-request `kv_transfer_params`. I'd tighten the explanation to attribute the behaviour to the proxy, not to the kv_role field, so a reader doesn't try to flip kv_role expecting it to matter.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+- **vLLM 0.21.0** for OpenAI-compatible serving
+- **UCCL-EP** (`deepep_high_throughput`, `deepep_low_latency`) for MoE
+  expert-parallel all-to-all over EFA
+- **NIXL** (NVIDIA Inference Xfer Library) over `LIBFABRIC` for KV-cache
+  transfer between prefill and decode pods
+- A simple Python proxy (`toy_proxy_server.py` shipped in vLLM) for routing
+
+It is the deployable version of the experiment
+[`experiments/2026-05-06-dsv3-disagg-mtp`](https://github.com/awslabs/awsome-distributed-ai)
```
KeitaW: ### Link to "upstream experiment" points at the repo root, not the experiment

The link text `experiments/2026-05-06-dsv3-disagg-mtp` suggests a path, but the URL is just the repo root (`https://github.com/awslabs/awsome-distributed-ai`) — and the `experiments/` directory doesn't exist in the repo at all (I checked with the contents API). A few lines later, "See `conclusion.md` in the upstream experiment for the full analysis" — also dangling. Could you either drop the references or replace them with whatever the canonical write-up is (a blog post, a Quip / SharePoint doc, an issue thread)?

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+CUDA-graph capture can take 2–5 minutes after `/health` returns 200. Be
+patient on the first benchmark.
+
+## Cost estimate (us-west-2 on-demand, May 2026)
+
+| Config | Nodes | Cost / hr | Cost per 1h benchmark |
+| --- | --- | --- | --- |
+| 1N | 1 | $98.32 | $98.32 |
```
KeitaW: ### Cost-estimate table will rot — recommend dropping it or pointing at the live source

Two issues with this table that I'd like to flag together:

1. **The $98.32/hr figure doesn't match `p5en.48xlarge` in any pricing tier I can find.** It is the on-demand rate for `p5.48xlarge` (H100) per the official [AWS EC2 On-Demand pricing page](https://aws.amazon.com/ec2/pricing/on-demand/) (filter: US West (Oregon) → GPU instances), which suggests a copy-paste from older p5 docs. Actual `p5en.48xlarge` is ~$63.30/hr on-demand in us-west-2 per the same AWS page (also surfaced via Vantage's parse of the AWS bulk-pricing API); [Capacity Block pricing](https://aws.amazon.com/ec2/capacityblocks/pricing/) for `p5en.48xlarge` lists rates that vary by region and reservation length, and EC2 spot is ~$25/hr. None of those round to $98.32.
2. **Even if we fix today's number, it rots.** [Capacity Block list prices](https://aws.amazon.com/ec2/capacityblocks/pricing/) reset roughly every quarter, on-demand p5en has been re-priced more than once since launch (track via the [What's New feed](https://aws.amazon.com/about-aws/whats-new/compute/?nc1=h_ls)), and the table has no "as-of" mechanism to age out.

Cleanest options, in order of preference:

- **Drop the table entirely** and replace it with a one-liner pointing at the live source: *"For current pricing, see [On-Demand](https://aws.amazon.com/ec2/pricing/on-demand/) or run `aws pricing get-products --service-code AmazonEC2 --filters Type=TERM_MATCH,Field=instanceType,Value=p5en.48xlarge ...`. As a rough sense of scale, the largest topology in this sample uses 6 nodes."*
- **Keep the table but pin it to a tier explicitly** (e.g., "Capacity Block list price, us-west-2, snapshotted YYYY-MM-DD; check the live page before budgeting"). That makes the staleness expectation honest rather than implicit.

(The hardware claims in the README — 8×H200 141GB, 16×EFA, 3,200 Gbps, ~30 TB NVMe — all check out against the [p5en instance page](https://aws.amazon.com/ec2/instance-types/p5/), so it's purely the cost rows.)

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+`setup/env_vars` controls every replaceable value. Required keys:
+
+| Variable | Example | Purpose |
+| --- | --- | --- |
+| `AWS_REGION` | `us-west-2` | ECR + EC2 region |
+| `ACCOUNT` | `123456789012` | AWS account ID for ECR |
+| `REGISTRY` | `${ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/` | ECR registry |
+| `IMAGE` | `vllm-uccl-ep` | ECR repo name |
+| `TAG` | `vllm0.21.0-uccl-0dc87eb` | image tag |
```
KeitaW: ### CLI tag naming drift: `vllm0.21.0-uccl-0dc87eb` vs `vllm0.21.0-uccl-0dc87eb-cu13`

The README's "Variables you must set" table shows `TAG=vllm0.21.0-uccl-0dc87eb`, while `setup/env_vars.example` defines `TAG="vllm0.21.0-uccl-0dc87eb-cu13"` (note the `-cu13` suffix). I'd standardize on the one that includes the CUDA marker so future CUDA-12 / CUDA-14 builds get distinct tags.

```suggestion
| `TAG` | `vllm0.21.0-uccl-0dc87eb-cu13` | image tag |
```

KeitaW (commented):
## Review Batch 4/5 — Documentation Consistency

Two highlights worth flagging:

- The README's explanation of `kv_role: kv_both` contradicts vLLM 0.21.0's own NixlConnector docs — the field is a no-op, the proxy's `do_remote_*` flags drive behavior. Worth tightening so readers don't try to flip `kv_role` expecting it to matter.
- The cost table has the wrong number ($98.32/hr is `p5.48xlarge` H100, not `p5en.48xlarge` H200) AND will rot quarterly even if corrected. I suggest dropping the table or pinning it to a tier explicitly with an "as-of" date.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/setup/env_vars.example
```diff
@@ -0,0 +1,74 @@
...
+#   cp setup/env_vars.example setup/env_vars
+#   $EDITOR setup/env_vars
+#   source setup/env_vars
+#
+# Never commit setup/env_vars — it contains your HF token and account ID.
+#
+# ---------------------------------------------------------------------------
+# AWS / ECR
+# ---------------------------------------------------------------------------
+export AWS_REGION="$(aws ec2 describe-availability-zones --output text --query 'AvailabilityZones[0].[RegionName]')"
```
KeitaW: ### `setup/env_vars.example` runs AWS CLI commands at source-time

The `export AWS_REGION="$(aws ec2 describe-availability-zones ...)"` and `export ACCOUNT="$(aws sts get-caller-identity ...)"` lines mean `source setup/env_vars` makes two synchronous AWS API calls every time. That's surprising behaviour for an env file — and on an unconfigured shell (or in a CI runner without IAM) it'll fail loudly. I'd either:

- replace them with placeholder values (`AWS_REGION="us-west-2"`, `ACCOUNT="123456789012"`) and a comment telling the reader to fill them in, or
- move the lookups into a dedicated `setup/discover-env.sh` script the reader runs once and that writes results back to `setup/env_vars`.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/setup/env_vars.example
```diff
@@ -0,0 +1,74 @@
...
+export ACCOUNT="$(aws sts get-caller-identity --query Account --output text)"
+export REGISTRY="${ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com/"
+export IMAGE="vllm-uccl-ep"
+export TAG="vllm0.21.0-uccl-0dc87eb-cu13"
+
+# ---------------------------------------------------------------------------
+# Cluster
+# ---------------------------------------------------------------------------
+export EKS_CLUSTER_NAME="<your-eks-cluster-name>"
```
KeitaW: ### Placeholder `<your-eks-cluster-name>` etc. without a "replace this" marker outside a code block

Bare `<your-eks-cluster-name>` and `<hf_your_token_here>` inside `export` lines will get copied into a `setup/env_vars` file verbatim and break subsequent commands with confusing errors. The convention in this repo is to either leave the value empty (`export EKS_CLUSTER_NAME=""`) with a `# TODO: fill in` comment, or prefix the placeholder with a clear sentinel value like `REPLACE_ME` so a `grep -n REPLACE_ME setup/env_vars` shows everything still pending. Same for `HF_TOKEN` on line 34.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
```diff
@@ -0,0 +1,401 @@
...
+[`experiments/2026-05-06-dsv3-disagg-mtp`](https://github.com/awslabs/awsome-distributed-ai)
+that benchmarked configs from a single 8-GPU unified node up to a 6-node
+1-prefill / 5-decode disaggregated topology, plus a UCCL-EP vs. AGRS
+(`allgather_reducescatter`) backend comparison.
+
+## What you can reproduce
+
+| Config       | Nodes | GPUs | Burst RPS | Burst tok/s | P50 TPOT @1RPS | P50 TPOT @burst | tok/s/GPU |
+| ------------ | :---: | :--: | --------: | ----------: | -------------: | --------------: | --------: |
+| 1N unified   |   1   |   8  |     13.49 |       3,454 |        21.83ms |         44.00ms |    431.55 |
```
KeitaW, line 34: ### "Burst RPS" comparisons rest on N=50 single-seed runs

The results table presents specific numbers (1P+4D: 19.18 RPS, 1P+5D: 17.16 RPS) and the headline findings claim a definite "optimal P:D ratio is 1:4" plus a "UCCL-EP beats AGRS by ~62%" comparison. The benchmark methodology underneath is `--num-prompts 50 --seed 1234` with a single run per rate per topology — that's a smoke-test sample size, not a statistical comparison. An 11% gap between 1P+4D and 1P+5D could easily disappear under run-to-run variance at N=50.

Two things would tighten this up:

1. Re-frame the table as "indicative numbers from a smoke benchmark" rather than research findings, so a reader doesn't carry the conclusions into their own capacity planning, OR
2. Run 3+ seeds per cell and add a confidence interval column (the `vllm bench serve` output already gives you per-run percentiles, you'd just need a loop over seeds).

(Worth noting that the public UCCL-EP vLLM benchmark in the upstream repo shows ~26% TPOT improvement and ~4% req-throughput improvement vs AGRS, so the ~30% TPOT figure here is consistent, but the 62% burst-RPS number isn't in any public benchmark I could find — that one is the Amazon-internal measurement that most warrants the framing tightening.)

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/recipe/benchmark.sh
```diff
@@ -0,0 +1,57 @@
...
+    --model "${MODEL}" \
+    --dataset-name random \
+    --random-input-len 1024 \
+    --random-output-len 256 \
+    --num-prompts 10 \
+    --request-rate 1 \
+    --seed 1234
+
+echo "==> Rate sweep (results -> ${RESULT_DIR})"
+for RATE in 1 4 8 16 inf; do
```
KeitaW: ### `benchmark.sh` runs each rate exactly once

Related to the methodology point on the README results table: the loop at line 41 does one `vllm bench serve` per rate. Wrapping that in a `for SEED in 1234 5678 9012; do` outer loop (with `--seed ${SEED}` and a result-dir-per-seed) would let users reproduce a multi-run aggregate without rewriting the script. The added cost is small relative to the 30–60 min model download.

KeitaW (commented):
## Review Batch 5/5 — Evaluation Methodology Validity + Wrap-up

Final batch — two methodology notes on the benchmark, then sources I relied on and a list of things I think the PR does particularly well.

---

## Sources

- RFC #1056 (placement rule): https://github.com/awslabs/awsome-distributed-ai/issues/1056
- EFA cheatsheet (NCCL_SOCKET_IFNAME guidance): https://github.com/awslabs/awsome-distributed-ai/blob/main/architectures/efa-cheatsheet.md
- nccl-tests Dockerfile (canonical exclusion-pattern example): https://github.com/awslabs/awsome-distributed-ai/blob/main/micro-benchmarks/nccl-tests/nccl-tests.Dockerfile
- EKS version calendar: https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
- EC2 on-demand pricing (cost-table verification): https://aws.amazon.com/ec2/pricing/on-demand/
- EC2 Capacity Block pricing: https://aws.amazon.com/ec2/capacityblocks/pricing/
- AWS What's New feed (track p5en pricing changes): https://aws.amazon.com/about-aws/whats-new/compute/?nc1=h_ls
- p5 / p5en instance specs: https://aws.amazon.com/ec2/instance-types/p5/
- vLLM v0.21.0 NixlConnector docs (`kv_role` placeholder, toy_proxy_server.py): https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html
- vLLM v0.21.0 `--all2all-backend` source (`vllm/config/parallel.py:40` Literal): https://github.com/vllm-project/vllm/blob/v0.21.0/vllm/config/parallel.py
- vLLM v0.21.0 release notes: https://github.com/vllm-project/vllm/releases/tag/v0.21.0
- NIXL v1.1.0 release notes (verbatim "switch to plain `nixl`" guidance): https://github.com/ai-dynamo/nixl/releases/tag/v1.1.0
- NIXL PyPI metadata (comparison of `requires_dist` between 1.0.1 and 1.1.0): https://pypi.org/pypi/nixl/1.0.1/json and https://pypi.org/pypi/nixl/1.1.0/json
- NIXL libfabric plugin (EFA + GDR requirement): https://github.com/ai-dynamo/nixl/blob/main/src/plugins/libfabric/README.md

---

## Things That Look Great

- The `recipe/deploy.sh` scoped `envsubst` allowlist (`$NAMESPACE $INSTANCE_TYPE …`) is exactly the right pattern — it preserves embedded shell variables like `$MY_IP` and `$endpoint` inside the pod commands. Nice to see this done properly first time.
- The pod-side `export VLLM_NIXL_SIDE_CHANNEL_HOST=$MY_IP` line is *required*, not just nice-to-have: NIXL's vLLM side-channel defaults to `localhost`, which would silently make cross-node disaggregation impossible. The pattern of computing `MY_IP` once and exporting it (plus the matching `--data-parallel-address`) is the right move.
- Version-pinning across the stack (vLLM `v0.21.0`, UCCL commit SHA, NIXL `v1.0.1`, NCCL `v2.30.4-1`, EFA `1.48.0`, GDRCopy `v2.5.2`) is thorough and meets/exceeds the CI floor on every component.
- The Dockerfile smoke-test block at the end catches the most likely silent-failure modes — `import vllm`, `import nixl`, `import uccl.ep`, `import deep_ep`, and the `has_deep_ep()` assert specifically. That's a great pattern other test cases should copy.
- The architecture diagram in the README is unusually clear for a multi-pod K8s sample — the prefill / decode / proxy boxes plus the NIXL and UCCL-EP arrows make the wire topology immediately legible.
- The troubleshooting section anticipates the three real failure modes (NCCL timeout on model load, NVSHMEM OOM, NIXL peer connection) with concrete remediations — much better than the generic "check the logs" troubleshooting most test cases ship with.
- Calling out `hostNetwork: true` as mandatory with the "28× slower with TCP fallback" warning is the kind of unvarnished operator wisdom that saves people days of debugging.
- The `.gitignore` keeps `setup/env_vars` (HF token, account ID) and `results/` out of git — clean separation of secrets and ephemeral output.
- The headline findings section is refreshingly honest about disaggregation's tradeoffs ("pays off mostly in latency, not throughput") — easy to imagine a version of this README that overclaims and yours doesn't.

Thanks again for the contribution — this is a strong addition to the inference samples and most of my comments are tightening rather than blocking.

KeitaW (changes requested):
Left some comments

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
pbelevich: Fixed — flipped to `SPDX-License-Identifier: MIT-0`. Same header now on every YAML / shell / env_vars / .gitignore / README in the test case.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Fixed — repo-layout diagram now shows `dsv3-uccl-nixl/` and the PR description's directory-structure section was updated too.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
pbelevich: Fixed — pinned to `ray[default]==2.55.1`. Matches the Software table in the README and the PR description.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/setup/install-prereqs.sh
pbelevich: Fixed — EFA plugin URL now uses `${EFA_PLUGIN_VERSION}` (already declared on line 9), matching the NVIDIA plugin's pinning style.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/unified.yaml
pbelevich: Fixed — see the rollup reply on `decode.yaml`.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
pbelevich: Fixed — pod commands now `export NCCL_SOCKET_IFNAME=^lo,docker,veth` (exclusion pattern preserved, EFA stays in the hot path). `GLOO_SOCKET_IFNAME=$IFACE` is kept since it's control-plane only. Same fix applied to `prefill.yaml` and `unified.yaml`. Inline comment in each manifest explains the rationale so the next reader doesn't re-introduce the override.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/prefill.yaml
pbelevich: Fixed — see the rollup reply on `decode.yaml`.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/hf-token-secret.yaml
pbelevich: Fixed — `hf-token-secret.yaml` deleted; the README's imperative `kubectl create secret` is the supported path.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
pbelevich: Fixed — toleration scoped to `operator: Equal, value: Schedulable` in all 4 manifests so failure-injection taints aren't tolerated.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
pbelevich: Fixed — converted to `Deployment` (replicas: 1, Recreate strategy) so K8s recreates pods on crash, and added `livenessProbe` (initialDelaySeconds: 1800, periodSeconds: 60, failureThreshold: 5) plus a `readinessProbe` on `/health` for prefill/decode/unified. Proxy gets a tcpSocket livenessProbe on port 8000. Will revisit with a `/v1/completions` round-trip probe once we've characterised steady-state latency on B300.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
pbelevich: Fixed — converted prefill/decode/proxy/unified to `Deployment` with `replicas: 1` and `strategy: Recreate`. K8s now recreates the pod after a vLLM crash; `recipe/teardown.sh` switched to deleting Deployments instead of Pods.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
pbelevich: Fixed — bumped to `NIXL_VERSION=v1.1.0` and collapsed the install to a single `uv pip install --system "nixl==1.1.0"`. Added a smoke-test step that asserts `nixl_ep` links `libcudart.so.13` (and not 12) at build time, so the wrong CUDA variant getting selected is now a build failure rather than a silent runtime issue. Comment block above the install line summarizes the v1.0.1 → v1.1.0 fix for future archeology.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/proxy.yaml
pbelevich: Addressed — added a tcpSocket livenessProbe on port 8000 to the proxy Deployment, plus an inline comment in `proxy.yaml` and a README section explaining that `tests/v1/.../toy_proxy_server.py` is the documented entry point but lives under `tests/`, so a `VLLM_COMMIT` bump can break the path. The image pins `VLLM_COMMIT=v0.21.0` for that reason. I'll open a follow-up issue tracking the swap to a permanent proxy once vLLM ships one.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/manifests/decode.yaml
pbelevich: Fixed — `imagePullPolicy: IfNotPresent` and `terminationGracePeriodSeconds: 120` set explicitly in all 4 manifests.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Fixed — bumped to `1.33+` in the prerequisites and Software table, with a one-liner explaining why (1.32 in extended support; 1.33 added native EFA traffic in the default EKS security group).

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/Dockerfile
pbelevich: Fixed — dropped `--no-deps`. `pip install .` now resolves `uccl.ep` (already installed in the previous step) and any future runtime deps the wrapper grows, so a missing dep fails at build rather than at `import deep_ep`.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Fixed — both dangling references removed. The "deployable version of the experiment ..." line was rewritten to drop the link, and the "see `conclusion.md`" pointer is gone too. The detailed analysis lives in an Amazon-internal write-up that doesn't have a public URL yet; if/when that changes, I'll add the link.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Fixed — table dropped. Replaced with a short "Cost" section pointing at the live On-Demand and Capacity Block pricing pages, plus an `aws pricing get-products` snippet so users can query the current rate themselves. Sense-of-scale ("largest topology uses 6 nodes; full sweep is 4–6 h") preserved.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Fixed — README "NIXL KV transfer" section rewritten to attribute prefill/decode behavior to the proxy's `do_remote_prefill` / `do_remote_decode` flags, and explicitly notes that `kv_role` in vLLM 0.21.0 is a placeholder per the upstream NixlConnector docs (linked). We still set `kv_both` because the validator requires *some* value.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/setup/env_vars.example
pbelevich: Fixed — both `aws ec2 describe-availability-zones` and `aws sts get-caller-identity` calls removed. `AWS_REGION` and `ACCOUNT` are now `REPLACE_ME` placeholders with a comment block above giving the one-liner to discover them, so sourcing the file makes zero AWS calls.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/recipe/benchmark.sh
pbelevich: Fixed — `benchmark.sh` now wraps the rate sweep in `for SEED in ${SEEDS}; do ...; done` (default seeds: 1234 5678 9012). Per-seed result dirs at `results/<topology>/seed-<seed>/`. `SEEDS`, `RATES`, and `NUM_PROMPTS` are env-overridable so you can drop to a single seed for a quick run or bump prompts for tighter intervals. README step 7 documents the override pattern.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Reframed — results table now sits under an "Indicative results" heading with an explicit blockquote caveat that the numbers are single-seed N=50 smoke-benchmark runs, not statistically tight measurements, and asks readers to re-run with the multi-seed loop in `recipe/benchmark.sh` and aggregate before quoting downstream. The "headline findings" block was rewritten as "directional observations" and now explicitly says the burst-RPS delta vs AGRS comes from a single-seed run and warrants reproduction. Same caveat propagated to the PR description's Test Results section.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/setup/env_vars.example
pbelevich: Fixed — every placeholder is now `REPLACE_ME`, including `EKS_CLUSTER_NAME`, `HF_TOKEN`, and all node hostnames / IPs. README step 1 now says `grep REPLACE_ME setup/env_vars` should print nothing once you've filled it in, so users get a single grep-based discovery for missing values.

3.test_cases/pytorch/vllm/dsv3-uccl-nixl/README.md
pbelevich: Fixed — README's "Variables you must set" table now uses `vllm0.21.0-uccl-0dc87eb-cu13`, matching `setup/env_vars.example`. CUDA marker stays so future cu12/cu14 builds get distinct tags.

pbelevich:
@KeitaW thanks for the thorough five-batch review — really helpful, and the rate of "this is a real bug" findings was high. I've pushed a single revision that addresses all 23 inline comments plus the cross-cutting items from each batch body. Per-comment replies are inline; here's the rollup so you can scan resolutions in one place:

**Batch 1 — Structure & Hygiene.** MIT-0 SPDX headers added to every YAML / shell / env_vars / `.gitignore` / README. Dockerfile flipped from Apache-2.0 to MIT-0. On placement: I noted in the PR description that this lands at `3.test_cases/pytorch/vllm/dsv3-uccl-nixl/` as an **interim** location pending [RFC #1056](https://github.com/awslabs/awsome-distributed-ai/issues/1056). The sample is self-contained, so a `git mv` to `examples/inference/vllm/deepseek-v3-disagg/` during the RFC cutover will be straightforward — happy to do that move in this PR if you'd prefer.

**Batch 2 — Pipeline & K8s correctness.** All three `NCCL_SOCKET_IFNAME` overrides fixed (now exclusion pattern, with a comment in each manifest about why). Pods → Deployments (replicas: 1, Recreate). Liveness + readiness probes on all 4 deployments. Scoped sagemaker toleration. Explicit imagePullPolicy + 120s grace. Unused `hf-token-secret.yaml` deleted.

**Batch 3 — Infra, NCCL & container.** EKS floor → 1.33+. NIXL → v1.1.0 (single-line install, plus a build-time `nixl_ep` cu13 link assertion). EFA plugin URL pinned to `${EFA_PLUGIN_VERSION}`. Ray pinned to `==2.55.1`. `--no-deps` dropped in `deep_ep_wrapper`.

**Batch 4 — Documentation.** `kv_role` rewritten per vLLM 0.21.0 docs (placeholder, proxy drives behavior). Cost table dropped, replaced with live-source links + `aws pricing get-products` snippet. Dangling experiment / `conclusion.md` references removed. README directory diagram → `dsv3-uccl-nixl/`. TAG drift fixed (`-cu13` everywhere). Proxy section now flags the `tests/v1/...` path-fragility.

**Batch 5 — Methodology.** Results table reframed as "indicative" with an explicit single-seed N=50 smoke-benchmark caveat. `recipe/benchmark.sh` now loops over 3 seeds (default 1234/5678/9012) with per-seed result dirs; `SEEDS` / `RATES` / `NUM_PROMPTS` env-overridable. Same caveat in the PR description's Test Results section.

Re your sources list — the [vLLM v0.21.0 NixlConnector docs](https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html), the [NIXL v1.1.0 release notes](https://github.com/ai-dynamo/nixl/releases/tag/v1.1.0), and the [EFA cheat sheet](https://github.com/awslabs/awsome-distributed-ai/blob/main/architectures/efa-cheatsheet.md) all turned out to be load-bearing. Linked them in the README's References section so future readers don't have to re-derive.

Ready for a second look whenever you have time.

KeitaW (approved):
# Code Review (round 2): PR #1110 — DeepSeek-V3 Disaggregated Inference with vLLM, UCCL-EP, and NIXL on EKS

**Author:** pbelevich
**Branch:** `dsv3-disagg-vllm-uccl-ep-nixl` → `main`
**Review date:** 2026-05-31
**Prior reviews:** [round 1 (CHANGES_REQUESTED)](https://github.com/awslabs/awsome-distributed-ai/pull/1110#pullrequestreview-4381181554) — 5 themed batches, 23 inline comments

---

## Summary

The revision (HEAD `0f835cf`) resolves **all 23** first-round inline findings plus the cross-cutting batch items — the rollup reply is accurate and I verified each fix against the changed files. **This is an approve-with-comments — I have no blocking concerns.** Beyond the desk review, I also **stood the sample up end-to-end on a live B300 EKS cluster** and exercised the actual NIXL/LIBFABRIC KV path — it works (details below). The only thing left is one optional documentation nicety, which shouldn't gate the merge.

One thing worth stating up front for other reviewers, since it came up on the thread: the `nixl#1609` "Bad address" failure (the "aws fork of libfabric" concern) is **GB200/arm64 + NVIDIA driver-595 specific** — it does **not** affect this `amd64`/`cu13` sample. I confirmed NIXL LIBFABRIC CUDA dma-buf registration succeeds cleanly on B300 (x86, driver 580.159.03), so #1609 is **not a merge blocker** for #1110.

---

|---|---|---|
| 1 | MIT-0 SPDX headers missing on YAML/shell/env/README | ✅ Resolved — header present on every file; `prefill.yaml:2`, `Dockerfile:2` |
| 2 | Dockerfile SPDX was Apache-2.0 | ✅ Resolved — flipped to MIT-0 |
| 3 | `NCCL_SOCKET_IFNAME=$IFACE` positive-selection override hid EFA | ✅ Resolved — all 3 GPU manifests `export NCCL_SOCKET_IFNAME=^lo,docker,veth` (`prefill.yaml:173`, `decode.yaml:171`, `unified.yaml:155`) |
| 4 | Bare `Pod` + `restartPolicy: Never` | ✅ Resolved — all 4 are `Deployment` (`replicas:1`, `strategy: Recreate`); teardown deletes Deployments |
| 5 | No `livenessProbe` on long-running pods | ✅ Resolved — liveness + readiness on prefill/decode/unified; tcpSocket liveness on proxy |
| 6 | Blanket `sagemaker…node-health-status` toleration | ✅ Resolved — scoped to `operator: Equal, value: Schedulable` (`prefill.yaml:69-72`) |
| 7 | `imagePullPolicy` / `terminationGracePeriodSeconds` unset | ✅ Resolved — `IfNotPresent` + `120s` explicit |
| 8 | Unused `hf-token-secret.yaml` | ✅ Resolved — file deleted; README uses imperative `kubectl create secret` |
| 9 | NIXL triple-install dance (v1.0.1 packaging bug) | ✅ Resolved — `nixl==1.1.0` single install + build-time `nixl_ep` cu13-link assertion (`Dockerfile:149,177-183`) |
| 10 | Ray pin unbounded vs documented 2.55.1 | ✅ Resolved — `ray[default]==2.55.1` (`Dockerfile:142`) |
| 11 | EFA plugin URL on unpinned `main` | ✅ Resolved — `${EFA_PLUGIN_VERSION}` wired into URL (`install-prereqs.sh:20`) |
| 12 | EKS floor 1.28 (out of support) | ✅ Resolved — bumped to 1.33+ with rationale |
| 13 | `--no-deps` in deep_ep wrapper | ✅ Resolved — dropped |
| 14 | `kv_role` README contradicts vLLM docs | ✅ Resolved — rewritten as placeholder; proxy `do_remote_*` drives behavior |
| 15 | Cost table wrong number + rot | ✅ Resolved — table dropped, replaced with live-pricing links + `aws pricing` snippet |
| 16 | Dangling `conclusion.md` / experiment refs | ✅ Resolved — removed |
| 17 | Dir-diagram name drift (`dsv3-disagg/`) | ✅ Resolved — `dsv3-uccl-nixl/` |
| 18 | TAG drift in README vs env_vars | ✅ Resolved — `-cu13` everywhere |
| 19 | `env_vars.example` made live AWS calls | ✅ Resolved — `REPLACE_ME` placeholders, zero AWS calls on source |
| 20 | Single-seed N=50 methodology framing | ✅ Resolved — "Indicative results" caveat + 3-seed `benchmark.sh` loop |
| 21 | Proxy path-fragility / no liveness | ✅ Resolved — inline comment + README note + tcp liveness; follow-up issue promised |
| 22 | Interim placement vs RFC #1056 | ✅ Resolved — called out as interim in PR description |
| 23 | Sources not linked in README | ✅ Resolved — References section added |

**Legend:** ✅ Resolved · 🟡 Partial · ❌ Open · ➖ N/A

---

## Reproduction & Validation (out-of-band, on B300)

I deployed the sample on a live `p6-b300.48xlarge` (Blackwell x86, driver 580.159.03) EKS cluster to validate the NIXL/EFA path the PR is built around. Three escalating tests, all green:

1. **NIXL LIBFABRIC CUDA dma-buf registration** — registers cleanly, no `Bad address` (the libfabric in the image even handles the `FORCE_PCIE → DEFAULT` dma-buf-mapping retry that GB200 trips on).
2. **Raw 2-node NIXL WRITE over EFA** — GPU→GPU RDMA across two nodes, data verified byte-for-byte.
3. **Full vLLM disaggregated run** using this PR's exact connector config (`NixlConnector`, `kv_role: kv_both`, `backends:[LIBFABRIC]`, `kv_load_failure_policy: fail`) with Llama-3.2-1B (TP=1) standing in for DeepSeek-V3: a request through `toy_proxy_server.py` returned 200 with **1 successful KV transfer prefill→decode @ 285 MB/s, 0 errors**. Because `kv_load_failure_policy: fail` is set, the successful completion proves KV actually moved over NIXL (a failed pull would have errored, not silently recomputed).

Net: the connector wiring, the `kv_both` placeholder decision, and the LIBFABRIC-over-EFA transport are all correct in practice on Blackwell x86. The one optional note below came out of running it.

### README is H200-only; worth noting the connector path is confirmed on Blackwell x86 too

**File:** `README.md` (lines 10, 95, 266–274, 360)

The prerequisites, node-selector examples, and tuning notes are all `p5en.48xlarge` (H200)-centric, and line 273-274 only says "for other hardware (H100, A100, B300) revisit `--gpu-memory-utilization`." Since the NIXL/LIBFABRIC/EFA KV path is the load-bearing part of this sample and I've now confirmed it works end-to-end on `p6-b300` (Blackwell x86), a one-line note in the prerequisites that the connector/EFA transport is hardware-portable (only the parallelism/memory knobs are H200-tuned) would broaden the sample's perceived applicability without overclaiming. Optional — purely a framing nicety.


---

## Sources

- vLLM v0.21.0 NixlConnector docs (`kv_role` placeholder, `toy_proxy_server.py` entry point): https://docs.vllm.ai/en/v0.21.0/features/nixl_connector_usage.html
- NIXL v1.1.0 release notes (single-`nixl` install): https://github.com/ai-dynamo/nixl/releases/tag/v1.1.0
- nixl#1609 (GB200/arm64 dma-buf "Bad address" — the non-blocker discussed in Summary): https://github.com/ai-dynamo/nixl/issues/1609
- `trust_remote_code` convention: [recurring-findings.md#trust_remote_codetrue](../../references/recurring-findings.md#trust_remote_codetrue)
- EFA cheatsheet (NCCL_SOCKET_IFNAME / EFA scheduling): https://github.com/awslabs/awsome-distributed-ai/blob/main/architectures/efa-cheatsheet.md

---

## Things That Look Great

- **Every round-1 finding closed in a single, well-documented revision** — the rollup reply maps cleanly to the diff, and several fixes went beyond the ask (the build-time `nixl_ep` cu13-link assertion in the Dockerfile turns a silent runtime CUDA-variant mismatch into a build failure — genuinely nice defensive engineering).
- **The connector config is correct in practice, not just on paper** — I exercised `NixlConnector` + `kv_both` + `LIBFABRIC` + `kv_load_failure_policy: fail` on real hardware and KV moved over EFA with zero errors. The `kv_role` placeholder reasoning from round 1's doc fix is borne out by the running system.
- **The `NCCL_SOCKET_IFNAME` exclusion-pattern fix landed with an explanatory comment in each manifest**, so the 28×-slowdown footgun won't get re-introduced by the next editor.
- **Methodology framing is now honest** — "Indicative results" + the multi-seed `benchmark.sh` loop is exactly the right way to ship a smoke-benchmark without overclaiming.

PR #1117, opened 2026-06-01
Updated version for Deepspeed

troy-ameigh:
## Purpose

Update the DeepSpeed test case to align with current library versions. The previous test relied on outdated dependencies that are no longer compatible, causing failures. This PR updates the test configuration to restore passing status.

## Changes
- Updated the TRANSFORMERS_VERSION to 4.52.0 from 4.44.0
- Updated an AWS CLI command to properly launch with new code versions
- Changed DeepSpeed Test Case README.md to properly use enroot to start the container.
- Removed Python-etcd as it is not needed with Slurm and python-etcd is abandoned (last release 2019) with no Python 3.12 support.
- Pinned to AWS CLI minimum version of 2.0 as v1 is in Maintenance Mode starting on July 15th, 2026.

## Test Plan

**Environment:**
- AWS Service: SageMaker Hyperpod w/ Slurm Orchestration
- Instance type: Used ml.g6.4xlarge for the compute instance type
- Number of nodes: 1 compute node. 

**Test commands:**
- Followed the README.md for the test case. 

## Directory Structure

```
3.test_cases/
└── pytorch/                # e.g. pytorch, megatron, jax
    └── deepspeed/              # e.g. picotron, FSDP, megatron-lm
        └── example_megatron_deepspeed/
         |_____ gpt
         |_____ qlora
        ├── 0.deepspeed.dockerfile      # Container / environment setup
        ├── 1.buildimage.sbatch
        ├── MakeFile




## Checklist

- [X} New test cases follow the [expected directory structure](#directory-structure).

### DIFF
--- a/3.test_cases/pytorch/deepspeed/0.deepspeed.dockerfile
+++ b/3.test_cases/pytorch/deepspeed/0.deepspeed.dockerfile
@@ -7,7 +7,11 @@
 # ============================================================
 FROM nvcr.io/nvidia/pytorch:25.04-py3
 
-ARG TRANSFORMERS_VERSION=4.44.2
+ARG TRANSFORMERS_VERSION=4.52.0
+ARG ACCELERATE_VERSION=1.6.0
+ARG DEEPSPEED_VERSION=0.16.4
+ARG PYNVML_VERSION=12.0.0
+ARG SENTENCEPIECE_VERSION=0.2.0
 ARG OPEN_MPI_PATH=/opt/amazon/openmpi
 
 ENV DEBIAN_FRONTEND=noninteractive
@@ -144,10 +148,12 @@ RUN mv ${OPEN_MPI_PATH}/bin/mpirun ${OPEN_MPI_PATH}/bin/mpirun.real \
 # 9. Python packages for DeepSpeed training
 # ============================================================
 RUN pip3 install --no-cache-dir \
-    awscli pynvml \
+    awscli \
+    pynvml==${PYNVML_VERSION} \
     transformers==${TRANSFORMERS_VERSION} \
-    sentencepiece python-etcd \
-    deepspeed>=0.16,<1.0 accelerate>=1.0,<2.0
+    sentencepiece==${SENTENCEPIECE_VERSION} \
+    deepspeed==${DEEPSPEED_VERSION} \
+    accelerate==${ACCELERATE_VERSION}
 
 RUN rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


KeitaW:
Thank you @troy-ameigh , could please add pre/post PR logs?

aravneelaws:
Thanks for the version pinning work @troy-ameigh - this is a welcome improvement. A few comments before I approve:

### Positives:
- Pinning all dependency versions to exact tags is the right approach and aligns with repo conventions. The move from ranges (>=0.16,<1.0) to exact pins (0.16.4) prevents silent breakage.
- The enroot start addition to the README is a real usability fix — without it, users would try to run the preprocessing script on the host where dependencies may not be available.

### Questions before approval:
1. python-etcd removal — This was silently dropped from the pip install without mention in the PR description. Was this intentional? DeepSpeed can use etcd as a rendezvous backend for multi-node launcher coordination. If the test case or any downstream sbatch relies on etcd-based rendezvous, removing it could break multi-node setups. Could you confirm this isn't needed?
2. Single-node testing only — The test plan mentions 1 compute node with ml.g6.4xlarge. Since this is a DeepSpeed test case (whose primary value is multi-node distributed training), was this previously tested on multi-node before the version bump? The python-etcd removal concern (above) is especially relevant here — if the old version relied on etcd for multi-node coordination, the new version needs an alternative or confirmation that DeepSpeed's default rendezvous (e.g., pdsh or torch.distributed) doesn't need it.
3. awscli still unpinned — Minor: everything else is now pinned but awscli is still installed without a version constraint. Would be good to pin it for consistency (e.g., awscli==1.x.x), or at least note that it's intentionally left floating.

Once I hear back on #1 and #2, happy to approve.
cc: @KeitaW

troy-ameigh:
Replying to questions:

1. Remove python-etcd dependency.
> python-etcd is abandoned (last release 2019) with no Python 3.12 support.
> It fails to build against the Python version in the pytorch:25.04-py3 base image, which was causing container build failures. This dependency is not needed for our use case — it was only required for DeepSpeed's legacy elastic training launcher, which uses an etcd server for worker discovery. Standard multi-node training with Slurm (which is what HyperPod uses) handles node coordination natively. DeepSpeed itself does not list python-etcd as an install dependency; it was optional and only relevant when explicitly using `deepspeed --elastic_training` with an etcd backed.

2. Everything should be good to go with multinode, however. I will retest with multinode, and remove the non-Slurm testing from the README.md

3. awscli still unpinned: I am pinning this and will update the PR.

troy-ameigh:
Retested with multi-node and multimode works fine. I also pinned AWS CLI to a minimum of v2.

aravneelaws (approved):
Thanks for the clarifications and the multi-node retest. LGTM.

PR #48, opened 2023-11-16
Neuronx nemo megatron

KeitaW:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #44, opened 2023-11-13
Update README.md

KeitaW:
add reviewer

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1109, opened 2026-05-24
Add reference cluster with AWS PCS

DaisukeMiyamoto:
## Purpose

Add reference cluster design with AWS Parallel Computing Service

## Changes

- update README for link to the PCS architecture
- add PCS architecture directory
- add PCS assets on assets/ directory including prerequisites, cluster, add compute node group templates.

## Test Plan

**Environment:**
Those assets are tested in the [AI/ML on PCS Workshop](https://catalog.workshops.aws/ml-on-pcs/) with CPU nodes. 
And also tested with `p5.48xlarge` instances on independent AWS account.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/awslabs/awsome-distributed-training/blob/main/CONTRIBUTING.md).
- [x] I am working against the latest `main` branch.
- [x] I have searched existing open and recently merged PRs to confirm this is not a duplicate.
- [x] The contribution is self-contained with documentation and scripts.
- [x] External dependencies are pinned to a specific version or tag (no `latest`).
- [x] A README is included or updated with prerequisites, instructions, and known issues.
- [x] New test cases follow the [expected directory structure](#directory-structure).

PR #50, opened 2023-11-20
Quick Create Link AWS Batch

sean-smith:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1120, opened 2026-06-03
feat(aws-pcs): add integrated monitoring, P6 GPU node groups, and container runtime to the PCS reference cluster

DaisukeMiyamoto:
## Purpose

The AWS PCS reference architecture already delivers a one-click (`deploy-all`),
ready-to-train ML cluster. This PR extends that existing experience with the pieces ML training
on PCS most commonly needs:

- an **integrated Prometheus/Grafana monitoring stack** (login node + compute exporters),
- **P6 (B200/B300) GPU node groups** with EFA, alongside the existing P5,
- a **container runtime (Enroot/Pyxis)** you can pre-bake into an AMI or install at first boot,
- **region-aware FSx performance tiers** (so the one-click flow works across more regions), and
- a **verified end-to-end test guide** with results from real hardware.

The base also adopts the **PCS-ready DLAMI**, so the one-click flow reaches a usable cluster
without a required custom AMI build (`BuildAMI=true` stays available for fastest node boot). All
validated on real hardware (p5 / p6-b200 / p6-b300) before submission.

Relates to: aws-parallelcluster-monitoring PRs #44 (Ubuntu/PCS support), #48 (node-local `/opt`
install), #49 (DCGM Docker-29.x tag) — all merged upstream; this PR pins to the resulting `v2.6.5`.

---

### Major updates

- **Integrated monitoring stack (`DeployMonitoring`).** Prometheus + Grafana + exporters deploy
  to the login node and DCGM/node exporters to compute nodes via the post-install hook. IAM
  (`MonitoringPolicy`) is built into `cluster.yaml`; Prometheus scopes discovery by the
  `aws:pcs:cluster-id` tag and tells login from compute by the `monitoring-role` tag (the
  dashboards key on a hardcoded `instance_name` label, not the EC2 `Name` tag). Pinned to
  **`v2.6.5`** (node-local `/opt` install fixes the shared-`/home` stale-file-handle race;
  DCGM tag works on Docker 29.x).
  Optional **`GrafanaPublicAccessCidr`** opens 443 on a login-only SG scoped to a CIDR, as an
  alternative to SSM port-forwarding. (Future direction: offer an AWS-managed backend — Amazon
  Managed Service for Prometheus + Amazon Managed Grafana — as an alternative to the self-hosted
  stack on the login node; tracked in `docs/ROADMAP.md`.)

- **P6 GPU node groups (`add-cng-p6-b200.yaml`, `add-cng-p6-b300.yaml`) + P5 multi-NIC EFA.**
  New B200/B300 templates with the correct EFA NIC layout; `deploy-all` auto-selects the P-series
  template by instance type. **EFA interface count is auto-derived from the instance type**
  (p5/p5e=32, p5en=16, p6-b200=8, p6-b300=16-of-17) — no manual `NetworkInterfaceCount`.
  `CapacityReservationId` documented as Capacity-Block-only.

- **Container runtime as a choice (Enroot 3.5.0 / Pyxis v0.20.0).** Pre-bake into a custom AMI
  (`BuildAMI=true`) **or** install at first boot (`PostInstallScriptUrl`). The installer is
  externalized to `scripts/install-enroot-pyxis.sh` (generic post-install hook, not Enroot-specific).

- **One-click / human-friendly `deploy-all`.** Console parameter groups + labels, sensible
  defaults, and **auto-resolving `AmiId`** that defaults to the **PCS-ready DLAMI** from SSM when
  left empty. Building on the PCS-ready DLAMI means the one-click flow needs no custom AMI build
  to get a working cluster — it shortens time-to-first-cluster, while `BuildAMI=true` remains
  available for those who want Enroot/Pyxis pre-baked for the fastest node boot.

- **README restructure + verified test guide.** README leads with one-click ML-training-ready,
  groups config to the console, and links a screenshot. New `tests/README.md` is a single guide
  with a **Verified configurations** matrix (real-HW results) plus runnable sbatch/scripts
  (nvidia-smi, NCCL all_reduce, FSDP Llama-2).

### Minor updates

- **Region-aware FSx storage tiers.** FSx for Lustre / OpenZFS support different deployment types
  and throughput tiers per region, so `LustreDeploymentType` (PERSISTENT_1/2 + free-form
  throughput) and `OpenZFSDeploymentType` are now parameters plumbed through from `deploy-all`.
  Defaults are unchanged; users in regions that lack a given tier can pick a supported one.
- `cluster.yaml`: dropped 5 unused params (now just `PrivateSubnetId` + `DefaultSecurityGroupId`);
  corrected inaccurate top-level `Description`s across templates.
- `install-enroot-pyxis.sh`: wait for apt/dpkg lock + retry `apt-get` — fixes intermittent
  first-boot `exit 100` (unattended-upgrades lock contention).
- Monitoring post-install retries up to 3× on transient failure (apt lock / NFS settle).
- Compute-node `RootVolumeSize` default 300 GiB (Enroot import needs local-disk overlay; can't
  overlay on Lustre); consistent across CNG templates.
- Docs: split `PARAMETERS.md` + `architecture-components.md` into `docs/`; add `docs/ROADMAP.md`
  (Multi-AZ prereq, managed monitoring, Trainium, test automation, LDAP/AD, IAM doc, P6e-GB200/300).
- `BuildAMI=true` documented to pair with `PostInstallScriptUrl=""` to avoid a boot-time
  double-install.

---

## Test Plan

**Environment:**
- AWS Service: AWS PCS (Slurm 25.11), `pcs-ml-cluster-deploy-all.yaml`
- Instance type: p5.48xlarge, p6-b200.48xlarge, p6-b300.48xlarge (Capacity Blocks); c6i (CPU/login)
- Number of nodes: 2× GPU per family; login + cpu1

**Test commands:**
```bash
# CPU/login + monitoring (BuildAMI=false, Enroot/Pyxis via PostInstallScriptUrl)
aws cloudformation create-stack --stack-name pcs-test \
  --template-url .../pcs-ml-cluster-deploy-all.yaml \
  --parameters ParameterKey=PrimarySubnetAZ,ParameterValue=us-west-2a \
               ParameterKey=DeployMonitoring,ParameterValue=true ...

# Pyxis container + NCCL + FSDP (see architectures/aws-pcs/tests/README.md)
srun --partition=cpu1 --container-image=ubuntu:22.04 bash -c 'echo OK'
sbatch -p gpu-p6-b200 tests/02-nccl-tests.sbatch
sbatch -p gpu-p6-b200 tests/03-fsdp-llama2.sbatch

# Pre-baked AMI path (BuildAMI=true, PostInstallScriptUrl="") — no double-install
aws cloudformation create-stack --stack-name pcs-amibuild-test \
  --parameters ParameterKey=BuildAMI,ParameterValue=true \
               ParameterKey=PostInstallScriptUrl,ParameterValue="" ...
```

## Test Results

| Config | NCCL all_reduce (2-node peak busbw) | FSDP Llama-2 7B |
|---|---|---|
| 2× p6-b200.48xlarge (16× B200) | ~654 GB/s (EFA, `found 8 nics`, `#wrong 0`) | ~223 TFLOPS/GPU, ~86k tok/s |
| 2× p6-b300.48xlarge (16× B300) | ~760 GB/s busbw † | ~195 TFLOPS/GPU, ~75k tok/s |
| 2× p5.48xlarge (16× H100) | ~480 GB/s (EFA, `found 32 nics`) | — |
| Login + CPU + monitoring | all Grafana targets up; dashboards populate | — |
| `BuildAMI=true` (custom AMI, `PostInstallScriptUrl=""`) | enroot 3.5.0 + Pyxis baked in, **no double-install**, cpu1 container job clean | — |

† B300 2-node/16 GiB likely doesn't saturate all 16 EFA cards — flagged in `tests/README.md`
for a larger-message / more-node re-test before treating ~760 GB/s as peak.

## Directory Structure

N/A — this PR updates the `architectures/aws-pcs` reference architecture (CloudFormation templates,
scripts, docs, validation guide), not a `3.test_cases/` training case.

## Checklist

- [x] External dependencies are pinned to a specific version or tag (no `latest`) — monitoring `v2.6.5`, Enroot 3.5.0, Pyxis v0.20.0.
- [ ] New test cases follow the expected directory structure — N/A (architecture update, not a `3.test_cases/` case).

## Notes for maintainers (deployment / release)

- **Before merge — templates and scripts must be hosted on S3 to deploy.** `deploy-all` uses
  nested stacks, so CloudFormation needs the child templates at an S3 `--template-url` (a GitHub
  raw URL is **not** a valid `TemplateURL`). The install script is fetched over GitHub raw at
  boot, but the templates themselves must be uploaded to a reachable S3 bucket
  (`S3BucketName`/`S3KeyPrefix`) to validate or deploy this PR.
- **After merge — please copy the templates to the public bucket.** Once merged, sync
  `architectures/aws-pcs/assets/*.yaml` to the public release bucket
  (`s3://awsome-distributed-ai/templates/`) so the README's one-click launch links and the
  workshop resolve against the released templates. The `PostInstallScriptUrl` default and the
  workshop URLs point at `aws-samples`/`awslabs` `main` (raw) and only resolve after this merge.
- **Workshop is already updated to match.** The companion workshop
  (`aiml-on-aws-parallel-computing-service-pcs`) has been updated for these changes (minimal
  `cluster.yaml` params, auto-resolved AMI, EFA auto-derive, public Grafana access, container
  runtime as a choice, refreshed monitoring/observability + NCCL pages). As soon as this PR is
  merged and the templates are published to the public bucket, we will proceed with publishing
  the workshop update.


KeitaW (commented):
## Review Batch 1/4 — Structure & Repository Hygiene

This is a substantial, well-engineered extension of the aws-pcs reference architecture (+3,051/−564 across 20 files): integrated Prometheus/Grafana monitoring, P6 B200/B300 node groups with auto-derived EFA layouts, an Enroot/Pyxis container runtime path, and a PCS-ready DLAMI default — all validated on real p5/p6-b200/p6-b300 hardware. I verified the riskiest claims independently — both statically (EFA topology against EC2 specs, live SSM resolution of the DLAMI parameter, tag containment of the upstream monitoring fixes) and **by deploying this PR's templates to a live PCS cluster** (p6-b200 Capacity Block, Slurm 25.11.2, us-east-1, 2026-06-04) — and the core engineering holds: one-click deploy reached a working cluster, Pyxis loads on 25.11, containers run for unprivileged users on stock Ubuntu 24.04, and the GPU dashboards populate. The findings below are: one near-certain boot-time defect in a default URL, a multi-user enroot cache bug found during live testing, a set of guided-path failures where defaults are invalid for the fallback options the PR itself recommends, security hardening on the opt-in public Grafana path (the unauthenticated-proxy exposure is empirically confirmed), and a structural ask about template and test-script duplication.

---

---

### Do we need a separate `add-cng-*.yaml` per GPU instance type?

**File:** `architectures/aws-pcs/assets/add-cng-p6-b200.yaml`, `add-cng-p6-b300.yaml`, `add-cng-p5.yaml`

After masking the `b200`/`b300` strings, the two new P6 templates differ in only ~80 of ~480 lines — they are ~85% identical, and the genuine deltas reduce to (a) the `NetworkInterfaces` block (8× `efa` vs ENA-primary + 16× `efa`) and (b) parameter defaults/descriptions. `add-cng-p5.yaml` already proves the consolidation pattern works: it serves three instance types (p5/p5e/p5en) from one template using a single condition (`Use32Interfaces`). Three near-identical templates triple the maintenance surface for every future fix — and we can already see sibling drift on day one: the p5 template gates `InstanceType` with `AllowedValues` but the two p6 templates don't.

Two concrete consolidation options, in order of preference:

**Option A — one `add-cng-gpu.yaml` with a per-type mapping + `Fn::ForEach`.** The `AWS::LanguageExtensions` transform lets the interface list be *generated* from a mapping instead of hardcoded 8/17/32 times:

```yaml
Transform: AWS::LanguageExtensions
Mappings:
  InstanceTypeConfig:
    p5.48xlarge:        { EfaCardIndices: ["0","1", ..., "31"], Card0IsEfa: "true"  }
    p5e.48xlarge:       { EfaCardIndices: ["0","1", ..., "31"], Card0IsEfa: "true"  }
    p5en.48xlarge:      { EfaCardIndices: ["0","1", ..., "15"], Card0IsEfa: "true"  }
    p6-b200.48xlarge:   { EfaCardIndices: ["0","1", ..., "7"],  Card0IsEfa: "true"  }
    p6-b300.48xlarge:   { EfaCardIndices: ["1","2", ..., "16"], Card0IsEfa: "false" }  # card 0 is ENA-only
Resources:
  LaunchTemplate:
    Properties:
      LaunchTemplateData:
        NetworkInterfaces:
          # plain ENA primary, emitted only for types whose card 0 is ENA-only (b300)
          - !If [Card0IsEnaOnly, { NetworkCardIndex: 0, DeviceIndex: 0, Groups: [...] }, !Ref 'AWS::NoValue']
          'Fn::ForEach::EfaCards':
            - CardIndex
            - !FindInMap [InstanceTypeConfig, !Ref InstanceType, EfaCardIndices]
            - InterfaceType: efa
              NetworkCardIndex: !Ref CardIndex
              DeviceIndex: 0   # unique per card is sufficient; AWS's own examples reuse 0 across cards
              Groups: [...]
```

Adding a future instance type (p6e, P7) becomes a one-line mapping entry instead of a fourth 450-line file. One caveat to weigh: templates using a transform need `CAPABILITY_AUTO_EXPAND` at create time, so the README's one-click launch links and CLI examples would need that capability added — worth checking it doesn't break the quick-create-link flow before committing to this path.

**Option B — plain CFN, no transform.** Keep one template with the full 32-entry interface list where each entry is wrapped in `!If [HasCardN, {...}, !Ref 'AWS::NoValue']`, with the handful of per-family conditions (`Is32Card`, `Is16Card`, `Is8Card`, `IsEnaOnlyCard0`) derived from an `AllowedValues`-locked `InstanceType` — exactly the `Use32Interfaces` pattern the p5 template already uses, extended one level. More verbose than Option A, but zero new capabilities/transform machinery and the same single-file maintenance story.

If you'd rather keep the three files split (e.g., because the b300's ENA-only card 0 makes a shared template harder to read, or to keep each template trivially diffable against the EC2 docs), that's a defensible call — could you note the rationale in the PR description or a template comment so the structure reads as a documented decision rather than copy-paste drift?

architectures/aws-pcs/tests/01-nvidia-smi.sbatch
```diff
@@ -0,0 +1,20 @@
...
+#SBATCH --ntasks=1
+#SBATCH --gres=gpu:8
+#SBATCH --exclusive
+#SBATCH --output=01-nvidia-smi_%j.out
+#SBATCH --error=01-nvidia-smi_%j.out
+
+set -euxo pipefail
+
+srun --container-image=docker://nvidia/cuda:12.4.1-base-ubuntu22.04 \
+     nvidia-smi
```
KeitaW: ### `tests/01-nvidia-smi.sbatch` — a full sbatch file for what is a one-liner

The stage-1 sanity check is `srun --container-image=... nvidia-smi` wrapped in 20 lines of sbatch boilerplate. Since `tests/README.md` already walks the user through each stage, I'd suggest dropping this file and showing the equivalent interactive one-liner in the README instead:

```bash
srun -p gpu-p5 --gres=gpu:8 --container-image=docker://nvidia/cuda:12.4.1-base-ubuntu22.04 nvidia-smi
```

That tests exactly the same thing (Pyxis/Enroot wiring + GPU visibility) with less to maintain, and it doubles as the user's first interactive command rather than a fire-and-forget batch job.

architectures/aws-pcs/tests/02-nccl-tests.sbatch
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
KeitaW: ### NCCL test duplicates `micro-benchmarks/nccl-tests` — reference it instead

The repo already maintains canonical NCCL test assets at [`micro-benchmarks/nccl-tests/`](https://github.com/awslabs/awsome-distributed-ai/tree/main/micro-benchmarks/nccl-tests), including Slurm launchers (`slurm/nccl-tests-ami.sbatch`, `slurm/nccl-tests-container.sbatch`) and the Dockerfile behind the same `public.ecr.aws/hpc-cloud/nccl-tests` image this script imports. Adding a parallel `02-nccl-tests.sbatch` + `02-import-nccl-image.sh` here creates a second copy that will drift (different EFA env defaults, different message-size sweeps) — this is the "reuse existing assets" rule from the repo checklist. I'd suggest the tests README link to `micro-benchmarks/nccl-tests/slurm/` and document only the PCS-specific deltas (partition names, the enroot-import step if the canonical launcher doesn't cover it). If the canonical launcher is missing something PCS needs, improving it there benefits every architecture at once.

For what it's worth, I ran this exact sbatch on the live verification cluster (2× p6-b200, 16× B200) and your test-results table reproduces: `NET/OFI Selected provider is efa, fabric is efa-direct (found 8 nics)`, peak busbw **642 GB/s** at 16 GiB vs your ~654 GB/s (within 2%), `Out of bounds values: 0 OK`. So this is purely a structure/reuse comment — the script itself works as documented.

architectures/aws-pcs/tests/02-import-nccl-image.sh
```diff
@@ -0,0 +1,35 @@
...
+#   ssh pcs-login            # or: aws ssm start-session --target <login-id>
+#   bash 02-import-nccl-image.sh
+#
+# Note: FSx Lustre cannot host the overlayfs that `enroot import` mounts, so the
+# import builds on the node-local root disk and writes only the resulting .sqsh to
+# /fsx — do not point ENROOT_* at /fsx.
+
+set -euxo pipefail
+
+: "${IMAGE_TAG:=latest}"
```
KeitaW: ### `02-import-nccl-image.sh` defaults to the `latest` image tag

`: "${IMAGE_TAG:=latest}"` — the PR checklist itself ticks "External dependencies are pinned to a specific version or tag (no `latest`)", and CONTRIBUTING calls this out explicitly, but the import script defaults to the mutable `latest` tag of `public.ecr.aws/hpc-cloud/nccl-tests`. Benchmark results aren't reproducible against a moving image (NCCL/EFA versions inside it change). I'd suggest defaulting `IMAGE_TAG` to the specific published tag the verified-configurations matrix was measured with, keeping the env-var override for people who want newer. (If this script is dropped in favor of `micro-benchmarks/nccl-tests` per the reuse comment above, this resolves itself.)

architectures/aws-pcs/tests/03-fsdp-llama2.sbatch
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
KeitaW: ### FSDP Llama test duplicates `3.test_cases/pytorch/FSDP` — reuse the existing test case

Same reuse concern as the NCCL script: the repo already has a maintained FSDP training test case at [`3.test_cases/pytorch/FSDP/`](https://github.com/awslabs/awsome-distributed-ai/tree/main/3.test_cases/pytorch/FSDP) (with `src/train.py`). I'd suggest the verification guide point at that test case's scripts (with a short PCS-specific launch note) rather than shipping a new `03-fsdp-llama2.sbatch`. That keeps one FSDP reference the team actually maintains, and the verified-configurations matrix can cite it just as well.


---

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -175,14 +282,29 @@ Parameters:
...
 
+  PostInstallScriptUrl:
+    Type: String
+    Description: >-
+      URL of a post-install script run on every node at first boot (PCS
+      equivalent of ParallelCluster OnNodeConfigured). Defaults to the
+      Enroot/Pyxis installer, which is the most common configuration. Set to an
+      empty string to skip, or override with another HTTP(S) URL to run a custom
+      script. S3 public hosting is not allowed.
+    Default: 'https://raw.githubusercontent.com/aws-samples/awsome-distributed-ai/main/architectures/aws-pcs/scripts/install-enroot-pyxis.sh'
```
KeitaW: ### `PostInstallScriptUrl` default points at the wrong GitHub org and will 404 at first boot

The default fetches the installer from `raw.githubusercontent.com/aws-samples/awsome-distributed-ai/...`, but this repo lives under **awslabs**. I verified live that GitHub raw does not alias the orgs (`awslabs/.../README.md` → 200, `aws-samples/.../README.md` → 404), so even after merge the default URL will 404 at node boot and the Enroot/Pyxis first-boot install fails for every default deployment. The same wrong-org URL also appears in the workshop-pointing docs, so it's worth a grep across the PR for `aws-samples/awsome-distributed-ai`.

```suggestion
    Default: 'https://raw.githubusercontent.com/awslabs/awsome-distributed-ai/main/architectures/aws-pcs/scripts/install-enroot-pyxis.sh'
```

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -412,10 +606,20 @@ Resources:
...
+          - ""
         IamProfileArn: !GetAtt ClusterStack.Outputs.InstanceProfileArn
         FSxLustreFilesystemId: !GetAtt PrerequisitesStack.Outputs.FSxLustreFilesystemId
         FSxLustreFilesystemMountName: !GetAtt PrerequisitesStack.Outputs.FSxLustreFilesystemMountname
         FSxOpenZFSFilesystemId: !GetAtt PrerequisitesStack.Outputs.FSxOFilesystemId
+        DeployMonitoring: !Ref DeployMonitoring
+        MonitoringVersion: !Ref MonitoringVersion
+        MonitoringRepo: !Ref MonitoringRepo
+        MonitoringRole: login
+        PostInstallScriptUrl: !Ref PostInstallScriptUrl
```
KeitaW, line 635: ### `BuildAMI=true` + default `PostInstallScriptUrl` double-installs — no template guard

`PostInstallScriptUrl` is passed unconditionally to every CNG stack, while the "set it to empty when `BuildAMI=true`" pairing rule lives only in the PR body — it isn't in the parameter descriptions a console user actually sees, and nothing enforces it. (Note the interaction with the previous finding: today the org-404 masks this — the script never fetches — so once the URL is fixed, this becomes the active foot-gun.) I'd suggest a conditional so the template enforces what the docs say:

```yaml
PostInstallScriptUrl: !If [BuildAMIEnabled, '', !Ref PostInstallScriptUrl]
```

or at minimum a warning in both parameter descriptions referencing each other.

KeitaW (commented):
## Review Batch 2/4 — Deployment Pipeline & Operational Correctness

architectures/aws-pcs/assets/add-cng-p6-b200.yaml
```diff
@@ -0,0 +1,426 @@
...
+
+  QueueName:
+    Type: String
+    Description: Name for Queue (leave empty to skip queue creation)
+    Default: ""
+
+  InstanceType:
+    Type: String
+    Description: Instance Type for Compute Node Group (p6-b200.48xlarge, 8 network cards)
+    Default: "p6-b200.48xlarge"
```
KeitaW, line 91: ### P6 templates accept any `InstanceType` string — failure surfaces late and cryptically

`add-cng-p5.yaml` gates `InstanceType` with `AllowedValues`, so a wrong value fails at stack validation. The p6 templates have no `AllowedValues`, so a mismatched type sails through CloudFormation and fails later at instance launch — the hardcoded 8-entry (b200) / 17-entry (b300) `NetworkInterfaces` blocks reference `NetworkCardIndex` values the wrong instance type doesn't have, which surfaces as an opaque compute-node-group launch error. Since each template's interface block is only correct for exactly one instance type anyway, locking the parameter makes the failure mode match p5's:

```suggestion
  InstanceType:
    Type: String
    Description: Instance Type for Compute Node Group (p6-b200.48xlarge, 8 network cards)
    Default: "p6-b200.48xlarge"
    AllowedValues:
      - "p6-b200.48xlarge"
```

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -109,15 +194,24 @@ Parameters:
...
+      - PERSISTENT_1
+
   PerUnitStorageThroughput:
-    Description: Provisioned Read/Write (MB/s/TiB)
+    Description: >-
+      Provisioned Read/Write throughput per unit of storage (MB/s/TiB). Valid
+      values depend on LustreDeploymentType: PERSISTENT_2 = 125/250/500/1000,
+      PERSISTENT_1 = 50/100/200.
     Type: Number
     Default: 250
```
KeitaW: ### FSx Lustre: default throughput (250) is invalid for the PERSISTENT_1 fallback the parameter recommends

The region-aware FSx parameters are a good idea and the regional restriction is real ([PERSISTENT_2 is limited-region per the FSx API reference](https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemLustreConfiguration.html)). But the fix is incomplete as shipped: `PerUnitStorageThroughput` defaults to 250, which is valid only for PERSISTENT_2 (P2 SSD: 125/250/500/1000; P1 SSD: 50/100/200). A user who follows the parameter's own guidance — switch `LustreDeploymentType` to PERSISTENT_1 because their region lacks P2 — and leaves the throughput default hits a `CreateFileSystem` failure deep in the nested prerequisites stack. Nothing couples the two parameters. I'd suggest either a Mappings/Conditions snap of throughput to the chosen type, or at minimum reciprocal warnings in both parameter descriptions ("if you change the deployment type, you must also change this value to …").

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -145,11 +239,24 @@ Parameters:
     Type: Number
     Default: 320
```
KeitaW: ### FSx OpenZFS: same coupling defect, and here the help text actively steers users into it

`HomeThroughput` defaults to 320, valid for the SINGLE_AZ_2/HA_2 grid but **not** for SINGLE_AZ_1 (64/128/256/512… per the [OpenZFS API reference](https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemOpenZFSConfiguration.html)) — and the `OpenZFSDeploymentType` description explicitly tells users to fall back to SINGLE_AZ_1 in regions without HA_2. Following that advice with the default throughput fails `CreateFileSystem`. Same suggested fix as the Lustre finding. (The MULTI_AZ_1 exclusion, by the way, is correct for this single-subnet topology — nice catch in the AllowedValues.)

architectures/aws-pcs/scripts/install-enroot-pyxis.sh
```diff
@@ -0,0 +1,150 @@
...
+
+# Configure Enroot
+ln -sf /usr/share/enroot/hooks.d/50-slurm-pmi.sh /etc/enroot/hooks.d/
+ln -sf /usr/share/enroot/hooks.d/50-slurm-pytorch.sh /etc/enroot/hooks.d/
+
+mkdir -p /tmp/enroot /tmp/enroot/data
+chmod 1777 /tmp/enroot /tmp/enroot/data
+
+wget -O /tmp/enroot.template.conf https://raw.githubusercontent.com/aws-samples/aws-parallelcluster-post-install-scripts/main/pyxis/enroot.template.conf
+ENROOT_CACHE_PATH=/tmp/enroot envsubst < /tmp/enroot.template.conf > /etc/enroot/enroot.conf
```
KeitaW: ### Shared `ENROOT_CACHE_PATH=/tmp/enroot` breaks the second user who imports an overlapping image — found in live testing

This one surfaced only when exercising the deployed cluster as two different users. The rendered `enroot.conf` makes the data and runtime paths per-user (`/tmp/enroot/data/user-$(id -u)` — good), but the **cache is shared**: `ENROOT_CACHE_PATH /tmp/enroot`. Enroot writes docker layer blobs there as `-rw-r----- <owner>` (0640), so when user A imports an image and user B later imports any image sharing a layer with it, B's import finds the cached blob but can't read it, and the job dies with the rather opaque `pyxis: tar: Error is not recoverable` → `couldn't start container` → `spank_pyxis.so: task_init() failed`. Reproduced on the live cluster: root imports `ubuntu:22.04`, then uid 1000 runs the same image → fails; same user with a fresh image (`alpine:3.20`) → works. On a multi-user Slurm cluster — the whole point of a login node — this will bite the second person who tries the README's container examples. The fix is one line in the `envsubst` call: make the cache per-user like the other paths.

```suggestion
ENROOT_CACHE_PATH=/tmp/enroot/cache-user-\$(id -u) envsubst < /tmp/enroot.template.conf > /etc/enroot/enroot.conf
```

(If a shared cache is desired to avoid duplicate downloads, the aws-hpc-recipes pattern of a group-scoped cache directory with group-readable blobs also works — but plain `/tmp/enroot` with 0640 blobs is the worst of both.)

architectures/aws-pcs/scripts/install-enroot-pyxis.sh
```diff
@@ -0,0 +1,150 @@
...
+  CPPFLAGS="-I ${SLURM_PATH}/include/" make install
+
+  mkdir -p "${SLURM_ETC_PATH}/plugstack.conf.d"
+  ln -sf /usr/local/share/pyxis/pyxis.conf "${SLURM_ETC_PATH}/plugstack.conf.d/pyxis.conf"
+
+  echo "Pyxis installed for Slurm ${SLURM_VERSION}"
+done
+
+# Update PATH for slurmd
+echo 'PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:/usr/lib/ccache/bin:/usr/local/bin:/usr/bin:/bin' >> /etc/default/slurmd
```
KeitaW: ### Duplicate `PATH=` append on re-run, and no slurmd restart — please confirm the execution context

Two small ones in the same neighborhood: (1) the `PATH=...` line is appended to `/etc/default/slurmd` with `>>` and no guard, so a re-run (e.g., the BuildAMI + post-install double-execution discussed above) appends a duplicate line — a `grep -q` guard or `sed -i` replace makes it idempotent; (2) unlike the aws-samples equivalent (`scontrol reconfigure` / `systemctl restart slurmd`), this script never restarts slurmd, so it only works if it always runs before slurmd starts. That holds for first-boot user data and AMI builds — could you confirm those are the only two execution contexts, and maybe note it in the script header? Also worth noting the hardcoded `slurm-25.11` in this PATH line silently mismatches clusters deployed with `SlurmVersion=25.05`, which the same script otherwise supports via `SLURM_VERSIONS="25.05 25.11"`.


---

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -205,20 +327,46 @@ Parameters:
...
+      (Optional) CIDR allowed to reach Grafana on the login node over HTTPS (port
+      443) directly from the internet. When SET, a login-only security group is
+      created that opens 443 to this CIDR, so you can browse https://<login-public-ip>/grafana/
+      without SSM port forwarding. Leave EMPTY (default) to keep Grafana private
+      and reachable only via SSM port forwarding. Use a tight CIDR (e.g. your
+      office IP /32); avoid 0.0.0.0/0 — Grafana would be exposed to the whole
+      internet (it still requires the admin password / Cognito, but open exposure
+      is risky). Only the login node is affected; compute nodes and FSx are not.
+    Default: ""
+    AllowedPattern: '^$|^(\d{1,3}\.){3}\d{1,3}/\d{1,2}$'
```
KeitaW: ### `GrafanaPublicAccessCidr` accepts `0.0.0.0/0` — the warning should be enforced, not prose

The overall posture here is genuinely good — empty-by-default (SSM port-forward only), 443-only SG scoped to the login node, random 32-char admin password in SSM SecureString. But the `AllowedPattern` matches any IPv4-shaped CIDR including `0.0.0.0/0` (and out-of-range octets like `999.1.1.1/32`), so the "avoid 0.0.0.0/0 — Grafana would be exposed to the whole internet" warning in the description is unenforced. Since the login node sits in the public subnet, this is the one place a single parameter value turns a reference deployment into an internet-exposed service. Tightening the regex to validate octets and reject `/0` makes the documented intent mechanical:

```suggestion
    AllowedPattern: '^$|^((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)\.){3}(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)/(3[0-2]|[12]\d|[1-9])$'
```

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -205,20 +327,46 @@ Parameters:
...
+    ConstraintDescription: Must be empty or a valid IPv4 CIDR (e.g. 203.0.113.4/32)
+
+  MonitoringVersion:
```
KeitaW, line 345: ### The monitoring stack's dcgm-exporter pin supports B200 but not B300 — the p6-b300 node group's GPU dashboards won't populate

I verified PRs #44/#48/#49 are all contained in the v2.6.5 tag — the pin's contents are what the PR body claims. One consequence worth surfacing: v2.6.5 pins `dcgm-exporter:4.2.0-4.1.0-ubuntu22.04` = DCGM 4.2.0. Per the [NVIDIA DCGM changelog](https://docs.nvidia.com/datacenter/dcgm/latest/release-notes/changelog.html), B200 support landed in 4.2.0 but **B300 requires 4.4.0** — and upstream #49 deliberately repinned *down* from 4.5.2 because DCGM tags ≥4.2.3 publish OCI image-index manifests Docker 29.x can't pull. Net effect: this PR adds a p6-b300 node group whose GPU dashboards won't populate.

I checked the failure mode on the live verification cluster (Docker 29.5.3): the #49 error reproduces (`docker pull ...:4.5.2-4.8.1-ubuntu22.04` → `error from registry: Incorrect Repository Format`), and the tag resolves to an OCI image index whose two `unknown/unknown` attestation manifests are what the pull trips on — the amd64/arm64 platform manifests themselves are ordinary. So the constraint is the index metadata served for NVCR tags ≥4.2.3, not the exporter or Docker versions (no `docker pull` argument helps either — `--platform` fails identically).

**Suggested update — a digest reference, verified working on the stock DLAMI Docker.** Pulling the same B300-capable image from NVCR **by digest** succeeds on the exact Docker (29.5.3) where the tag fails, because a digest pull skips the index negotiation entirely. So in upstream `compose/compute.gpu.yml`:

```yaml
# DCGM 4.5.2 (B200 + B300). Digest of the linux/amd64 manifest for
# 4.5.2-4.8.1-ubuntu22.04 — digest pull bypasses the NVCR OCI attestation
# index that breaks tag pulls on Docker 29.x (#47/#49); verified on the
# PCS DLAMI (Docker 29.5.3).
image: nvcr.io/nvidia/k8s/dcgm-exporter@sha256:a7ad6547d4546eaf4dd5d6b4c0b4db4101e63ef7dc3cdff7f42b767d2c60b706
```

Zero new infrastructure, still pulls from NVCR, and a stronger pin than a tag. One caveat: the digest is amd64-only (arm64 nodes would need the sibling digest `sha256:9b3969a9…` — only relevant if Grace-based nodes are added later). Since the change lives in aws-parallelcluster-monitoring (which you maintain), the path here would be an upstream release + a `MonitoringVersion` bump in this PR; if you'd rather not respin upstream now, the fallback for *this* PR is a README known-issues note (B300 GPU metrics pending) so b300 users don't debug a "broken" dashboard.


---

KeitaW (commented):
## Review Batch 3/4 — Infrastructure, Monitoring & Container Security

### Opening 443 exposes unauthenticated Prometheus and Pushgateway, not just password-gated Grafana

**File:** `architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml` (Grafana public access path, cross-references upstream v2.6.5 nginx config)

At monitoring v2.6.5, the login node runs the full head compose stack, and its [nginx 443 listener](https://github.com/aws-samples/aws-parallelcluster-monitoring/blob/v2.6.5/nginx/conf.d/nginx.conf) reverse-proxies four paths: `/grafana/` (auth-gated) plus `/prometheus/` → :9090, `/pushgateway/` → :9091 and `/slurmexporter/` → :8080, none of which have authentication. So whoever can reach the allowed CIDR can read all cluster metrics (and push to Pushgateway) without credentials — the parameter is named "Grafana public access" but actually publishes more. **Empirically confirmed on a live deployment of these templates** (2026-06-04): `curl -sk https://<login>/prometheus/api/v1/status/buildinfo` returns 200 with the full Prometheus 3.1.0 build info and zero credentials; `/pushgateway/` likewise 200. The raw ports themselves are *not* reachable (the SG opens only 443), so the exposure is solely via the proxy paths. Since the fix belongs upstream (restrict those locations to localhost/VPC or put them behind auth), I'd suggest: (a) raising it on aws-parallelcluster-monitoring, and (b) in this PR, renaming/clarifying the parameter description to say what is actually exposed, so operators make an informed choice. (Minor live observation for completeness: `/slurmexporter/` returned 502 on the test cluster — the slurm exporter wasn't serving; possibly the `MetricsType=metrics/openmetrics` prerequisite — worth a quick look since the Slurm dashboards depend on it.)

KeitaW (commented):
## Review Batch 4/4 — Documentation Consistency

### A few description/doc drifts worth a pass

**File:** multiple

- The PR body says monitoring identifies nodes "by `Name=HeadNode`/`Compute` tags" — at v2.6.5 that's no longer how it works (EC2 SD filters on `aws:pcs:cluster-id` + `monitoring-role` tags; `HeadNode`/`Compute` are hardcoded Prometheus *labels* kept for dashboard compatibility). Worth updating so future readers don't go tag-hunting.
- The DLAMI default resolves `/aws/service/pcs/ami/dlami-base-ubuntu2404/x86_64/latest/ami-id` — I confirmed via live SSM that **only** `/latest/` paths exist (no version-pinned alternative), and CloudFormation re-resolves it on stack updates, so node AMIs drift across scale-out events. That's a reasonable default for a reference architecture (and AWS's docs endorse it), but a one-line README note telling production users to capture the resolved AMI ID into the `AmiId` parameter would pre-empt the homogeneity surprise.
- `install-enroot-pyxis.sh` configures Slurm `25.05` and `25.11` only; PCS also still supports 24.11 (EOL 2027-05). Deliberate scoping is fine — a README sentence saying 24.11 clusters won't get Pyxis would make it explicit.


---

---

## Verification Environment

Findings marked "verified"/"empirically confirmed" were tested on a live deployment of this PR's templates at head `aadd9f4` (2026-06-04, us-east-1): `pcs-ml-cluster-deploy-all.yaml` with `SlurmVersion=25.11`, `DeployMonitoring=true`, `BuildAMI=false`, p6-b200.48xlarge on a Capacity Block, PCS-ready DLAMI (Ubuntu 24.04, Docker 29.5.3, driver 595.71.05), `PostInstallScriptUrl` pointed at the PR branch (the default URL 404s — see the first deployment finding). Stack reached CREATE_COMPLETE; verified: Pyxis SPANK load on Slurm 25.11.2, unprivileged container runs with `apparmor_restrict_unprivileged_userns=1`, 8× B200 enumerated by both `nvidia-smi` and dcgm-exporter 4.2.0 metrics, unauthenticated `/prometheus/` + `/pushgateway/` behind 443, and the cross-user enroot cache failure. The PR's own 2-node NCCL test (`tests/02-nccl-tests.sbatch`, 16× B200 over a Capacity Block) was also run end-to-end: `found 8 nics`/`efa-direct`, peak busbw 642 GB/s at 16 GiB (PR claims ~654; within 2%), zero data errors — including a live exercise of the CNG elastic path (second node cold-started on job submission: instance launch → srun-able in ~10 min via the first-boot install, and an earlier idle node was correctly powered down by PCS).

---

---

## Things That Look Great

- **The one-click promise is real.** Deploying this PR's templates end-to-end (with only the `PostInstallScriptUrl` org corrected) produced a working Slurm 25.11 cluster with monitoring dashboards receiving B200 GPU metrics and `srun --container-image` working — on the first attempt, ~35 minutes, including Capacity Block integration that picked up the instance two minutes after the reservation window opened.
- **The EFA auto-derivation is correct — all five instance types.** I checked p5/p5e=32, p5en=16, p6-b200=8, and p6-b300=16-of-17 against the EC2 specs, including the subtle part: the b300 template correctly leaves card 0 as a plain ENA interface (it's ENA-only on that platform) while p5/b200 correctly put `efa` on card 0. Removing the manual `NetworkInterfaceCount` parameter is a real usability win.
- **Real-hardware validation with a verified-configurations matrix** (p5, p6-b200, p6-b300, Capacity Blocks) — including honestly flagging the b300 busbw number as possibly unsaturated rather than presenting it as peak. That's exactly the right scientific instinct for a reference architecture. **And the numbers reproduce**: my independent 2-node p6-b200 run of your sbatch landed within 2% of your published busbw with identical NIC discovery and zero data errors.
- **Grafana credential handling is strong**: random 32-char password in SSM SecureString with `--no-overwrite`, KMS via-service condition, cluster-scoped IAM — no admin/admin anywhere, no anonymous access. Private-by-default with SSM port-forwarding as the blessed path.
- **The MonitoringPolicy IAM is near-least-privilege** — read-only, region-conditioned where the APIs allow it, and the only write is the tightly-scoped Grafana-password parameter.
- **The PCS adaptation of the installer is careful**: correct PCS plugstack/header paths (`/etc/aws/pcs/scheduler/...`), compile-from-source Pyxis per Slurm version, the apt/dpkg lock wait + `DPkg::Lock::Timeout` retry combination, and the `curl -f`/`gpg --batch --no-tty` hardening over the aws-samples original.
- **The 300 GiB root-volume default is well-justified** — enroot's overlay upperdir genuinely can't live on Lustre, and the PR places the writable paths on local disk accordingly.
- **FSx parameter plumbing is exactly right** — all 8 parameters pass parent→child with matching names/defaults, and the P2-only `MetadataConfiguration` is correctly conditioned.
- Consistent MIT-0 headers, `set -exo pipefail`, and version pins on every external dependency — the contribution-guideline basics are all in place.

---

## Sources

- EFA topology verification (F-series card counts, b300 ENA-only card 0): https://docs.aws.amazon.com/ec2/latest/instancetypes/ac.html, https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html
- Wrong-org 404 (live probes): https://raw.githubusercontent.com/awslabs/awsome-distributed-ai/main/README.md (200) vs https://raw.githubusercontent.com/aws-samples/awsome-distributed-ai/main/README.md (404)
- FSx valid values: https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemLustreConfiguration.html, https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystemOpenZFSConfiguration.html
- aws-samples pyxis postinstall.sh (slurmd-restart comparison): https://github.com/aws-samples/aws-parallelcluster-post-install-scripts/blob/main/pyxis/postinstall.sh
- Enroot local-overlay requirement (supports the 300 GiB root default): https://github.com/NVIDIA/enroot/issues/30
- Monitoring stack at the pinned tag: https://github.com/aws-samples/aws-parallelcluster-monitoring/blob/v2.6.5/nginx/conf.d/nginx.conf, .../installer/install.sh, .../compose/compute.gpu.yml, .../CHANGELOG.md
- DCGM GPU-generation support: https://docs.nvidia.com/datacenter/dcgm/latest/release-notes/changelog.html
- PCS DLAMI + Slurm versions: https://docs.aws.amazon.com/pcs/latest/userguide/working-with_ami_pcs-ready-dlami.html, https://docs.aws.amazon.com/pcs/latest/userguide/slurm-versions.html; live `aws ssm get-parameter(-s-by-path)` in us-west-2/us-east-1/ap-northeast-1/eu-west-1 (2026-06-03)
- Reuse targets: https://github.com/awslabs/awsome-distributed-ai/tree/main/micro-benchmarks/nccl-tests, https://github.com/awslabs/awsome-distributed-ai/tree/main/3.test_cases/pytorch/FSDP
- Nested-stack S3 requirement: https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-stack.html
- Live verification cluster (see Verification Environment): PCS cluster `pcs_3duasoz99i`, us-east-1, 2026-06-04 — slurmd/sinfo 25.11.2 output, dmesg AppArmor audit, `/tmp/enroot` cache ownership listing, dcgm-exporter `:9400/metrics`, nginx proxy curl probes

---

KeitaW (commented):
## Follow-up — consider removing the custom AMI-build path now that the vendored PCS-ready DLAMI exists

First: thank you for the thorough response push — the org fix, per-user enroot cache, FSx coupling validation, `AllowedValues`, the `DcgmExporterImage` override + upstream v2.9.1, and adopting the canonical NCCL/FSDP assets all look great.

One structural follow-up, in the same spirit as the template-consolidation comment. The Image Builder path (`BuildAMI` / `pcs-ready-dlami-with-enroot-pyxis.yaml`) came from #1109, when no vendored PCS DLAMI existed and a custom build was the only way to get a PCS-compatible AMI. That rationale is gone — at this PR's head the template's own `Mappings` base it on `/aws/service/pcs/ami/dlami-base-ubuntu2404/.../latest/ami-id`, so its only remaining function is pre-baking Enroot/Pyxis (plus rebuild-schedule/lifecycle/SSM extras). I'd suggest evaluating removing it from this PR while the revision is still unpublished:

**The carrying cost is visible in this very revision.** The `EnrootPyxisComponent` is an inline re-implementation of `install-enroot-pyxis.sh`, so the recent Slurm-version/PATH hardening (`PCS_SLURM_VERSION`, single-version Pyxis) had to be implemented twice — once in the script, once in the AMI component — plus an OPERATIONS.md section to explain the AMI's single-version semantics. Every future installer fix pays that double-maintenance tax.

**What removal deletes:** the 481-line template (~17 resources incl. two IAM roles and the Image Builder pipeline), the `DLAMIStack` wiring + 4-5 parameters in `deploy-all`, and the associated README/OPERATIONS sections — all net deletion.

**What it costs, with mitigations:**
- *Node cold-boot:* pre-baking only skips the Enroot/Pyxis install slice. On a live deployment of this PR I measured ~13 min launch→job-running on the first-boot path, of which the install is plausibly ~2–4 min — PCS bootstrap and the monitoring post-install run at boot either way, so the realistic saving is much smaller than the README's "~3 min boot" framing suggests.
- *AMI pinning / customization:* already covered — every CNG accepts a literal `AmiId`, and OPERATIONS.md §4 now documents exactly that production-pinning path. Users who want a pre-baked AMI can build one by any external means and pass `AmiId` + `PostInstallScriptUrl=""`.
- *Double-install risk:* already mitigated by your warn + idempotent-installer work, so the in-stack build is no longer load-bearing for safety either.

**If you'd rather keep a pre-bake recipe available**, a soft landing is to decouple: drop `DLAMIStack` + the `BuildAMI`/`BaseAmiId`/`SemanticVersion`/`BuildSchedule` parameters from `deploy-all` and keep the template as a standalone optional asset ("advanced: bake, then pass `AmiId` + `PostInstallScriptUrl=\"\"`"). That alone takes the one-click path down to a single AMI story.

And if the scheduled-rebuild/lifecycle features are a deliberate offering for your users, keeping it is defensible too — in that case I'd just ask for a sentence in the README on why the in-stack build earns its complexity, so the next reviewer doesn't re-litigate this. Timing-wise, doing whichever you choose in this PR avoids publishing (bucket + workshop) a path that might be removed right after.

KeitaW (changes requested):
Few comments

DaisukeMiyamoto:
Thanks for the thorough, hardware-validated review — it caught real bugs. The structural and guided-path issues are all fixed on `deploy-monitoring` and re-validated end-to-end across the matrix at the top of [`tests/README.md`](architectures/aws-pcs/tests/README.md): clean first-boot deploys on Slurm 25.05 and 25.11 (CPU + Pyxis), the `BuildAMI=true` path, and a 2× p6-b300 GPU run (NCCL/FSDP, monitoring v2.9.1). Rundown of the threads not answered individually below:

**Repo hygiene / test duplication**
- **`01-nvidia-smi.sbatch` (one-liner):** removed the file; Test 5 now shows the interactive `srun … nvidia-smi` one-liner instead. (`91c0774`)
- **`02-nccl-tests.sbatch` duplicates `micro-benchmarks/nccl-tests`:** removed; Test 6 now points at the canonical `slurm/nccl-tests-container.sbatch` and documents only the PCS deltas (login-node import, partition name). (`91c0774`)
- **`02-import-nccl-image.sh` defaults to `latest`:** removed with the script; the README example now pins a specific image tag. (`91c0774`)
- **`03-fsdp-llama2.sbatch` duplicates `3.test_cases/pytorch/FSDP`:** removed; Test 7 now references the canonical FSDP case (`create_venv.sh` + `llama2_7b-training.sbatch`) with only the PCS deltas (venv/HF cache on `/fsx`, 2 nodes). (`91c0774`)

**Boot-time / guided-path correctness**
- **`PostInstallScriptUrl` wrong org (404):** fixed `aws-samples` → `awslabs` in the default and grepped the repo for the same wrong-org URL. (`d52c7ee`)
- **P6 templates accept any `InstanceType`:** added single-value `AllowedValues` to the p6-b200 and p6-b300 templates (p5 already had it), so a mismatched type fails at stack validation. (`83d84df`)
- **FSx Lustre throughput invalid for the PERSISTENT_1 fallback:** the throughput params are now `String` + `AllowedValues` (the full valid grid), and a CloudFormation `Rules` section asserts the (deployment-type, throughput) pair at stack-create time — so a mismatch fails immediately with a clear message instead of deep in the nested FSx stack. Defaults unchanged. (`bb32202`)
- **FSx OpenZFS same coupling defect:** same fix (`Rules` + `AllowedValues`) for `HomeThroughput` vs `OpenZFSDeploymentType`. (`bb32202`)
- **Shared `ENROOT_CACHE_PATH` breaks the 2nd user:** made the cache per-user (`/tmp/enroot/cache/user-$(id -u)`), matching the data/runtime paths. (`d52c7ee`)

**Beyond the original review** (surfaced while validating the above on real hardware):
- **`MonitoringVersion` default → `v2.9.1`:** carries the PCS `/opt` install fix, Docker-29.x DCGM tag, Grafana 13, and the upstream `DCGM_EXPORTER_IMAGE` support I added for B300 (#51). (`13b3781`)
- **`SlurmVersion` flows everywhere it's needed:** deploy-all → CNG UserData (as `PCS_SLURM_VERSION`) → `install-enroot-pyxis.sh`, and deploy-all → DLAMI build. The Pyxis SPANK plugin ABI is locked to its Slurm version, so the AMI is single-Slurm-version on purpose; both paths now build Pyxis only for the cluster's version. (`87ebf94`, `6e2d37c`, `7bdd26e`)
- **`MetricsType` is 25.11+ only:** `cluster.yaml` was emitting it unconditionally, which made `SlurmVersion=25.05` clusters fail to create. Gated on 25.11. (`0f96116`)
- **New `docs/OPERATIONS.md`:** consolidates the operational caveats above (plus AMI pinning, FSx deployment-type/throughput coupling) so the README doesn't grow each time we hit one. (`aef17d3`, `1e401f9`)

Each is linked to its commit above; happy to expand on any. The four threads with a design choice or a deeper finding I've answered in-thread.

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
DaisukeMiyamoto: Good catch that the pairing rule was unenforced. I went a different way than the suggested `!If [BuildAMIEnabled, '', !Ref PostInstallScriptUrl]` guard, because `PostInstallScriptUrl` is meant to be a **generic** first-boot hook (the PCS analogue of `OnNodeConfigured`), not Enroot/Pyxis-specific — forcing it empty under `BuildAMI=true` would silently drop any *other* first-boot setup wired through it.

Instead I made the installer **idempotent** (`a9572d1`): it skips the Enroot `.deb` when the version already matches, and skips Pyxis per Slurm version when its plugin and plugstack config are already in place. So `BuildAMI=true` + the default URL is a fast no-op on a pre-baked node — no double install. The `BuildAMI`/`PostInstallScriptUrl` descriptions cross-reference each other and recommend `PostInstallScriptUrl=""` for the cleanest boot, without forcing it (`2522c82`).

Related: the AMI build path now takes a `SlurmVersion` parameter (deploy-all passes the cluster's value through), so the baked Pyxis always matches the cluster (`7bdd26e`). Background in [`docs/OPERATIONS.md §2`](./architectures/aws-pcs/docs/OPERATIONS.md#2-container-runtime-postinstall-vs-ami-build).

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
DaisukeMiyamoto: Agreed the prose warning was unenforced. I kept `0.0.0.0/0` **accepted** rather than rejecting it, deliberately: the public path exists largely for short-lived PoC/workshop use where granting each attendee local SSM/IAM is impractical, and `0.0.0.0/0` is sometimes the pragmatic choice there. Hard-blocking it would push those users to hand-edit the SG (worse).

Instead I strengthened the docs to state the real exposure (`caff091`): the description + README now spell out that opening 443 also exposes the **unauthenticated** `/prometheus/`, `/pushgateway/`, `/slurmexporter/` proxy paths (not just password-gated Grafana), and tell users to narrow/clear the CIDR when done. Happy to also tighten the octet validation (reject malformed CIDRs like `999.x`) while still allowing `/0` — that part of your regex is a clear win.

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
DaisukeMiyamoto: Rather than pin the dcgm image here, the fix went upstream: the dcgm-exporter image is now configurable in aws-parallelcluster-monitoring (#51, **released in `v2.9.1`**). This PR defaults `MonitoringVersion` to `v2.9.1` and adds a **`DcgmExporterImage`** parameter (deploy-all → CNG → the monitoring UserData exports `DCGM_EXPORTER_IMAGE`). (`f3387f7`, `13b3781`)

Verified on 2× p6-b300 (Slurm 25.11, Docker 29.5.3) with the released v2.9.1 stack: the DCGM 4.5.2 image **by digest** (`sha256:a7ad6547…`, exactly your suggestion) pulls past the OCI-index issue and **all 16 B300 GPUs report in Grafana** (temp/util/power/NVLink/DCP profiling, 312 `DCGM_FI_*` series). Since the same image covers Hopper / B200 / B300 with no field changes vs 4.2.0 (per the DCGM changelog) and is what we ship-tested on B300 hardware, `DcgmExporterImage` defaults to that digest — every supported GPU family populates the dashboards out of the box. The monitoring stack's older 4.2.0 pin remains a one-line override for anyone who needs to match another fleet.

Documented for users: parameter row in [`docs/PARAMETERS.md` §7](./architectures/aws-pcs/docs/PARAMETERS.md#7-developer--advanced), rationale + override examples in [`docs/OPERATIONS.md §3.1`](./architectures/aws-pcs/docs/OPERATIONS.md#31-dcgmexporterimage-the-default-and-when-to-change-it).

architectures/aws-pcs/scripts/install-enroot-pyxis.sh
DaisukeMiyamoto: Confirmed the execution context: the script runs **before slurmd starts** (first-boot UserData or AMI build), so it intentionally doesn't restart slurmd — documented in the header now. The duplicate `PATH=` append on re-run is fixed (the line is de-duplicated before appending).

The hardcoded `slurm-25.11` matters because the PCS DLAMI ships three Slurm versions (24.11/25.05/25.11), and Slurm components are version-locked: an older `scontrol` on slurmd's PATH fatally fails to parse a 25.11 cluster's `slurm.conf` (`unrecognized key: MetricsType`, set for the monitoring OpenMetrics endpoint) and aborts the Pyxis PMI hook, and the Pyxis SPANK plugin itself is rejected with `Incompatible Slurm plugin version` if its `spank_pyxis.so` was built for a different Slurm. So both Pyxis and slurmd's PATH have to be pinned to the cluster's version.

The catch is that the script can't *discover* that version while it's running — cloud-init runs before slurmd starts and before `/etc/profile.d/slurm.sh` / the controller config that name it exist. So **the cluster's `SlurmVersion` is injected** instead: deploy-all threads it to every CNG, the CNG UserData exports it as `PCS_SLURM_VERSION` before running `install-enroot-pyxis.sh`, and the script uses it to build Pyxis only for that version (installing the `.so` to a per-version libdir) and pin slurmd's PATH to the same version's bin (`87ebf94`, `6e2d37c`). The AMI build path is single-Slurm-version for the same reason (`7bdd26e`). When PCS adds a native post-install hook, the cluster Slurm version is the kind of context it should expose the same way.

Single-version is fine across cluster upgrades because Slurm's [upgrade policy](https://slurm.schedmd.com/upgrades.html) supports running with components within "the previous two (three on 24.11+) major releases" — `scontrol`/`srun` from the prior version interop with a newer `slurmctld`. To advance, set the new `SlurmVersion` and redeploy; Pyxis is rebuilt for it.

Verified on clean first-boot deploys of 25.05 and 25.11 CPU clusters and on a 2× p6-b300 GPU cluster (NCCL + FSDP). Background in [`docs/OPERATIONS.md §2`](./architectures/aws-pcs/docs/OPERATIONS.md#2-container-runtime-postinstall-vs-ami-build).

DaisukeMiyamoto:
## Re: Do we need a separate `add-cng-*.yaml` per GPU type?

Thanks for laying out both options so concretely. I weighed them and **kept the three files split**, but addressed the two things you flagged (the drift, and documenting the decision).

**Why split:** Option A (`Fn::ForEach` + `AWS::LanguageExtensions`) needs `CAPABILITY_AUTO_EXPAND` at create time, which **breaks the README/workshop one-click quick-create links** — core to this PR's one-click goal. Option B (one file, no transform) means ~32 per-card `!If` wrappers plus card-0/DeviceIndex branching to cover b300's genuinely different topology (ENA-only card 0 + EFA on cards 1-16 at `DeviceIndex 0`, vs card 0 = EFA / `DeviceIndex 1` on p5/b200) — and editing the already-validated p5/b200 NIC blocks. Given b300 is the odd one out and each flat list is trivially diffable against the EC2 docs, I judged split as the safer, more readable call.

**What I changed (`83d84df`):** documented the rationale in a header comment on all three templates; added the `Fn::ForEach` consolidation to `docs/ROADMAP.md` (with your AUTO_EXPAND caveat); and fixed the day-1 drift — the p6 templates now gate `InstanceType` with `AllowedValues` like p5. Happy to revisit Option A as a follow-up if the quick-create-link concern is resolved.

DaisukeMiyamoto:
## Re: removing the custom AMI-build path

Thanks for the prompt-and-press on the structural cost — the double-maintenance during this round on `PCS_SLURM_VERSION` and the single-version Pyxis pin was real, and you're right that the AMI's origin reason (no vendored PCS DLAMI) is gone now that the templates resolve the official `dlami-base-ubuntu2404` parameter directly.

That said, I'd like to keep the path, for two reasons:

1. **Both pre-bake and first-boot install are independently useful.** ParallelCluster keeps the same pair (`CustomAmi` + `OnNodeConfigured`/`OnNodeStart`) in its mainline today, even with an official PC-vendored AMI available — because the trade-off is real: pre-bake gives faster boots and deterministic state for fleets that scale frequently, first-boot install gives easy iteration on the customization itself. PCS has the same use cases. Removing the pre-bake path now would close that door for users who'd choose it on the same grounds PC users do; if anything the precedent suggests we should keep it for the same lifecycle.
2. **The in-stack template earns its complexity vs. a one-shot build.** What it adds beyond an ad-hoc `aws imagebuilder start-image-pipeline-execution` is the *managed pipeline*: scheduled rebuilds (`BuildSchedule=Weekly`/`Monthly`), an AMI lifecycle policy that deprecates older AMIs on a configurable window, and SSM parameter publishing of the latest AMI ID. For users with a steady stream of clusters off the same PCS-tuned base, that's the difference between a CFN stack you forget about and a hand-rolled CI job with the same deprecate/notify story.

Concretely for the next reviewer: README §Container runtime now states why the in-stack build exists (managed pipeline + schedule + lifecycle + SSM publishing); OPERATIONS.md §2 already documents the single-Slurm-version semantics — that was the part of the recent hardening that doubled up with the script, so it's where the duplication tax landed.

If the pipeline parts turn out not to be used in practice, decoupling (or removal) in a follow-up PR is on the table — but I think removing it pre-merge here, before users have had a chance to opt in, would foreclose the option prematurely.

DaisukeMiyamoto:
## Re-considered the AMI-build path — decoupling done

Thanks for the prompt-and-press on this; revisiting it I came around to your view. The double-maintenance you flagged was the right reason to act, and walking through the full template, ParametersGroups, and the ParallelCluster precedent again, I no longer think keeping the in-stack build pulls its weight when the same outcome is just one extra step (build the AMI separately, pass `ami-xxx` as `AmiId`).

So I went ahead and pulled the AMI-build path out of `deploy-all`:

**Decoupled**
- Dropped 4 parameters from `pcs-ml-cluster-deploy-all.yaml`: `BuildAMI`, `BaseAmiId`, `SemanticVersion`, `BuildSchedule`. Also dropped the `BuildAMIEnabled` Condition and the nested `DLAMIStack` resource. Each CNG's `AmiId` is now a plain `!Ref AmiId` — the CNG templates already handle empty as SSM-resolve internally, so no second branch is needed at the deploy-all level.
- Added a top-level **`AmiId`** parameter under §2 PCS Cluster Configuration (rather than tucked under "AMI build"), since this is also the natural production knob — empty resolves the latest PCS-Ready DLAMI from SSM, explicit `ami-xxx` pins it. This is the same shape "pin the AMI for production" already wanted independent of any pre-bake.
- The standalone `pcs-ready-dlami-with-enroot-pyxis.yaml` template stays intact (single-Slurm-version Pyxis, optional schedule/lifecycle/SSM-publish features). It's now what you run separately to get a custom AMI; you read its `DLAMIforPCSAmiId` output and pass it back to the cluster as `AmiId`. (`88da755`)

**1-click parameter UX, while at it**
- ParameterGroups went from 8 to 7 sections, and the order/labels were tightened for the default 1-click case: `SlurmVersion` first in §2 (the substantive cluster choice), `LoginNodeInstanceType` after it, `RootVolumeSize` moved out of the container-runtime section into §2 PCS Cluster Configuration since it applies to every node group, not just the post-install hook. Long descriptions trimmed (`PostInstallScriptUrl`, `GrafanaPublicAccessCidr`, `BuildAMI`, `DcgmExporterImage`); FSx labels tagged with the mount path (`/fsx`, `/home`) so it's obvious which filesystem each row governs; `DcgmExporterImage` was orphaned (not in any ParameterGroup) and got grouped under §7 Developer/Advanced. (`5bc200a`, `2c295cc`, `709b85a`)
- `CapacityReservationId`'s description and console label both spell out **"Capacity Block for ML reservation ID"** explicitly, with the warning to *not* put an ODCR ID there, since the parameter name itself is the generic one.

**Docs and tests follow the new structure**
- New README §9 "Pre-baking Enroot/Pyxis into a custom AMI (optional)" walks the 3-step flow: build the AMI as a separate stack → read `DLAMIforPCSAmiId` → deploy the cluster with `AmiId=<ami-xxx>` + `PostInstallScriptUrl=""` for the cleanest boot. Each step has a copy-pasteable `aws cloudformation` block.
- `tests/README.md` Test 2 now covers only the default first-boot install path; the pre-baked-AMI flow is a new **Test 8** that walks the standalone DLAMI build → cluster pin → verify, run **per supported `SlurmVersion`** (Pyxis SPANK plugin ABI is version-locked, so testing 25.05 and 25.11 separately is the only way to catch wrong-version `spank_pyxis.so`). The pre-merge matrix gains a row 8 you skip when neither `pcs-ready-dlami-with-enroot-pyxis.yaml` nor `scripts/install-enroot-pyxis.sh` changed. (`c068374`)
- `OPERATIONS.md §2` rewritten — the section header is now "PostInstall vs. pre-baked AMI" and the opening line is "**they are decoupled**: the cluster stack does not run Image Builder." `PARAMETERS.md` ToC mirrors the new 7-section console layout. README most-used-parameters table swaps the old `BuildAMI` row for an `AmiId` row.
- One consistency sweep on top: the AMI is referred to as **"PCS-Ready DLAMI"** (short) / **"PCS-Ready Deep Learning AMI"** (long) consistently across every doc and template, after finding three different spellings in flight. (`39d02a2`)

**Quick links to the structural commits, if useful for review:** [`88da755`](https://github.com/awslabs/awsome-distributed-ai/pull/1120/commits/88da755) (decoupling), [`c068374`](https://github.com/awslabs/awsome-distributed-ai/pull/1120/commits/c068374) (Test 8 split), [`5bc200a`](https://github.com/awslabs/awsome-distributed-ai/pull/1120/commits/5bc200a) (1-click groupings).

The cluster stack is now genuinely just "Prerequisites + cluster + CNGs", which is what `deploy-all` should mean. Defaults still produce the same fast-path one-click that worked before — the difference is only visible if you previously set `BuildAMI=true`, in which case the README §9 walkthrough is the substitute (and shorter than expected, since the 3 steps are basically what the in-stack version was doing anyway).

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -205,20 +321,52 @@ Parameters:
...
-    Description: HPC Recipes for AWS release branch
-    AllowedPattern: '^(?!.*/\.git$)(?!.*/\.)(?!.*\\.\.)[a-zA-Z0-9-_\.]+$'
+    Description: >-
+      (Optional) CIDR allowed to reach the login node on HTTPS/443. Empty (default)
+      = SSM port-forward only (recommended). When set, also exposes the
+      UNauthenticated /prometheus/, /pushgateway/, /slurmexporter/ proxy paths
+      (not just password-gated Grafana). Use the tightest CIDR you can; clear it
+      when done.
+    Default: ""
+    AllowedPattern: '^$|^(\d{1,3}\.){3}\d{1,3}/\d{1,2}$'
```
KeitaW: ### Grafana CIDR: add octet validation now that `/0` is a documented choice

I'm fine with the resolution of the policy half: the description and OPERATIONS §3.2 now disclose exactly what 443 exposes, and accepting `0.0.0.0/0` for short-lived PoC/workshop use is a documented, deliberate call — your call to make. What remains is only the *typo guard*: the current pattern accepts non-CIDRs like `999.999.999.999/32` or `10.0.0.0/99`, which fail later at SG creation with a less helpful error. Since `/0` is intentionally allowed, the octet/prefix-validating version is:

```suggestion
    AllowedPattern: '^$|^((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)\.){3}(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)/(3[0-2]|[12]?\d)$'
```

(This permits `0.0.0.0/0` per your documented stance — it only rejects values that aren't valid IPv4 CIDRs at all.)

architectures/aws-pcs/assets/pcs-ml-cluster-deploy-all.yaml
```diff
@@ -205,20 +321,52 @@ Parameters:
...
+    Default: 'aws-samples/aws-parallelcluster-monitoring'
+
+  DcgmExporterImage:
+    Type: String
+    Description: >-
+      dcgm-exporter image for GPU nodes. Default = DCGM 4.5.2 by digest, covering
+      H100/B200/B300 (digest pull bypasses the Docker 29.x OCI-index failure on
+      newer NVCR tags). Override to pin a different build (also use a digest).
+      No effect on CPU nodes.
+    Default: 'nvcr.io/nvidia/k8s/dcgm-exporter@sha256:a7ad6547d4546eaf4dd5d6b4c0b4db4101e63ef7dc3cdff7f42b767d2c60b706'
```
KeitaW, line 370: ### `DcgmExporterImage` digest default: note it's the linux/amd64 manifest

Glad the digest route landed, and pairing it with upstream v2.9.1's override support is the right shape. One small completeness note for the description: `sha256:a7ad6547…` is the digest of the **linux/amd64** platform manifest specifically (the arm64 sibling is `sha256:9b3969a9…`). All instance types this template set deploys are x86_64 today, so nothing is broken — but since the parameter invites overrides "also use a digest", a parenthetical "(amd64 digest; arm64 nodes would need the matching arm64 digest)" would save the first Grace-based user a confusing pull failure if/when the ROADMAP's P6e-GB200/GB300 item lands.


---

architectures/aws-pcs/docs/OPERATIONS.md
```diff
@@ -0,0 +1,247 @@
...
+
+## 7. Known issues
+
+### 7.1 First-srun-on-a-fresh-cpu1-node can drain the node ("Prolog error")
+
+**Symptom.** The very first `srun` against a freshly-launched compute node fails with
+`Job prolog failed`, the node enters `drained` state with `Reason=Prolog error`, and the
+job is cancelled. The next `srun` (which lands on a sibling node) works.
+
+**Cause.** A systemd / cgroup-v2 race during PCS bootstrap, observed on the PCS-ready
```
KeitaW, line 196: ### cgroup-v2 prolog race: great known-issue entry — add a tracking pointer if one exists

The new known-issue entry is exactly the kind of operational honesty that makes a reference architecture trustworthy — observed sequence, misleading-error explanation, root-cause scoping ("nothing this repo touches is in the path"), and a concrete recovery command. Two small asks: (a) if this has been reported to the PCS service team (or there's an internal/public issue to follow), a one-line "tracked at <ref>" gives readers a way to know when it's fixed and gives future maintainers a signal for when this section can be deleted; (b) if it hasn't been reported yet, it probably should be — the analysis here is already most of a support case.


---

KeitaW (commented):
## Review (round 2)

This is an exemplary response round: 40 commits since round 1 resolving **16 of 17** prior findings, several of them more thoroughly than asked — FSx throughput/type coupling became CloudFormation `Rules` assertions (fail-at-create with clear messages), the AMI build was decoupled from `deploy-all` into a standalone path with `AmiId` promoted to a single top-level parameter, the B300 metrics gap became a `DcgmExporterImage` parameter defaulting to a DCGM 4.5.2 digest (plus upstream v2.9.1 support), and all four duplicated test scripts were deleted in favor of the canonical `micro-benchmarks/nccl-tests` and `3.test_cases/pytorch/FSDP` assets. The one partial carry-over is the Grafana CIDR pattern (now a documented deliberate accept for `0.0.0.0/0`, but still missing octet validation), and I have three minor new notes on content added this round. Nothing blocking remains from my side.

---

---

|---|---------|--------|
| 1 | Separate `add-cng-*.yaml` per GPU instance type | ✅ Resolved — took the documented-split option: each template's `InstanceType` now explains the per-family NIC lock-in, and OPERATIONS.md §6 records the rationale |
| 2 | `01-nvidia-smi.sbatch` → srun one-liner | ✅ Resolved — file deleted; tests/README shows the interactive one-liner |
| 3 | NCCL test duplicates `micro-benchmarks/nccl-tests` | ✅ Resolved — scripts deleted; tests/README now references the canonical sbatch (9 canonical-asset references) |
| 4 | `02-import-nccl-image.sh` defaults to `latest` | ➖ N/A — file deleted with the reuse change |
| 5 | FSDP test duplicates `3.test_cases/pytorch/FSDP` | ✅ Resolved — deleted; canonical test case referenced |
| 6 | `PostInstallScriptUrl` wrong org (404) | ✅ Resolved — default now `awslabs/...@main` (d52c7ee); fork refs stripped pre-merge (0af2621); resolves on merge |
| 7 | BuildAMI double-install unguarded | ✅ Resolved — installer is now idempotent (per-component version-skip), and the param description documents the no-op; the AMI build decoupling removes the interaction entirely |
| 8 | p6 templates missing `AllowedValues` | ✅ Resolved — both templates locked, with the NIC-topology rationale in the description (83d84df) |
| 9 | FSx Lustre throughput/type coupling | ✅ Resolved — `Rules`/`Assertions` validate P1 (50/100/200) and P2 (125/250/500/1000) at create time; stronger than the suggested description fix |
| 10 | FSx OpenZFS throughput/type coupling | ✅ Resolved — same `Rules` pattern for both the 64-grid (SINGLE_AZ_1/HA_1) and 160-grid (SINGLE_AZ_2/HA_2) families |
| 11 | Shared `ENROOT_CACHE_PATH` multi-user break | ✅ Resolved — cache is now per-user (`/tmp/enroot/cache/user-$(id -u)`, d52c7ee) |
| 12 | Duplicate `PATH=` append / slurmd restart context | ✅ Resolved — idempotent writes, `PCS_SLURM_VERSION`-derived single-version PATH, and the manual-run caveat documented in the script header |
| 13 | Grafana CIDR accepts `0.0.0.0/0` | 🟡 Partial — exposure now fully disclosed and `0.0.0.0/0` documented as a deliberate PoC/workshop accept (OPERATIONS §3.2); the pattern still passes invalid octets — see the typo-guard note below |
| 14 | 443 exposes unauthenticated Prometheus/Pushgateway | ✅ Resolved (as documentation) — parameter description now names the unauthenticated proxy paths explicitly; OPERATIONS §3.2 covers it; the auth fix itself remains an upstream matter |
| 15 | B300 GPU dashboards won't populate (DCGM 4.2.0 pin) | ✅ Resolved — `DcgmExporterImage` parameter, default DCGM 4.5.2 **by digest** (Docker-29.x-safe), `MonitoringVersion` bumped to v2.9.1 which honors the override (upstream #50) |
| 16 | Doc drifts (tag mechanism, AMI pinning, 24.11 scope) | ✅ Resolved — AMI-pinning tip (OPERATIONS §4), 24.11 out-of-scope note, naming consistency sweep |
| 17 | Follow-up: remove/decouple the custom AMI-build path | ✅ Resolved — Option B adopted: `DLAMIStack` + 4 parameters removed from `deploy-all`, the template is a documented standalone path with its own test flow (Test 8), and `AmiId` is one top-level parameter |

---

---

## Things That Look Great

- **The FSx `Rules` section** — assert-at-create with explicit valid-value lists and human-readable `AssertDescription`s is *better* than the coupling fix I suggested, and it establishes a validation pattern the template family can reuse.
- **The AMI-build decoupling** landed cleanly: `deploy-all` shrank by ~150 lines, `AmiId` is one comprehensible top-level parameter ("empty = latest PCS-Ready DLAMI; pin in production"), and the standalone path kept its own test flow (Test 8) rather than losing coverage.
- **The `DcgmExporterImage` solution is better than what I proposed** — a parameter with a digest default plus upstream override support (v2.9.1, #50) fixes B300 *and* gives users an escape hatch for future DCGM/Docker incompatibilities, instead of hard-pinning anything.
- **Deleting all four test scripts** in favor of canonical assets was decisive — tests/README now teaches with references instead of copies, which is exactly the "reuse existing assets" ideal.
- **The installer's idempotency design** (per-component version-skip, documented no-op on pre-baked AMIs) turned the double-install foot-gun into a non-issue without any template conditional.
- **The cgroup-v2 known-issue entry** in OPERATIONS.md — documenting a race you found in your own validation, with the journalctl evidence and recovery command, is the kind of operational content most reference architectures never ship.
- The personal-fork reference sweep (0af2621) before merge — small thing, often forgotten.

---

## Sources

- Resolution verification at head `39d02a20cbc735a4932342b8d6207b97503c5dd8`: `pcs-ml-cluster-deploy-all.yaml` (Rules section, parameter blocks), `install-enroot-pyxis.sh` (idempotency, per-user cache, `PCS_SLURM_VERSION`), `add-cng-p6-b200.yaml` (AllowedValues), OPERATIONS.md, tests/README.md
- Round-1 evidence base (EFA topology, live-cluster verification, DCGM/NVCR analysis): see round-1 reviews linked above
- arm64 digest for the dcgm-exporter note: NVCR manifest index for `4.5.2-4.8.1-ubuntu22.04` (inspected 2026-06-04)

---

KeitaW (approved):
LGTM! Thank you @DaisukeMiyamoto

PR #1121, opened 2026-06-03
Fix lcs u24 hotfix

hossamazon:
## Summary

Make `mock-gpu-driver-deb.sh` resilient when the running NVIDIA kmod version isn't published in apt. Today the script hard-fails if `apt-cache show libnvidia-compute-${MAJOR}=${DRV_VERSION}` misses, which blocks HyperPod cluster lifecycle scripts on
AMIs whose kmod patch lags or leads what apt currently ships.

## Motivation

Observed on a Slurm cluster (Ubuntu 24.04 AMI) where `modinfo -F version nvidia` reported `595.58.03`, but apt only had `libnvidia-compute-595` candidate `595.71.05-0ubuntu0.24.04.1`. All three exact-match attempts (`-1ubuntu1`, `-0ubuntu1`, ``) missed
and the lifecycle script exited 1 on every node:

Error: Could not find package libnvidia-compute-595 with version 595.58.03 in apt-cache
[SageMaker] The lifecycle scripts failed.

## Change

- Keep the existing exact-patch lookup as the preferred path (no behavior change on AMIs that already work).
- Add a fallback: if no exact match, fetch the latest `libnvidia-compute-${MAJOR}` candidate via `apt-cache policy`, build the equivs control file from it, and rewrite `Version:` to the running kmod's version so any downstream pins on the kmod patch
still resolve.
- Keep the original error message path for the case where apt has no candidate at all.

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/hotfix/enabling-slurm-25.md
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/hotfix/enabling-slurm-25.md
@@ -1,0 +1,1 @@
+# Enabling Slurm 25
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/hotfix/enabling-slurm-25.md
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/hotfix/enabling-slurm-25.md
@@ -1,1 +1,0 @@
-# Enabling Slurm 25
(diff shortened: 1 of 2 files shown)


PremiumSpider (approved):
## Test Results - APPROVED

Tested on HyperPod cluster (ml.g5.xlarge, NVIDIA driver 580.126.09, us-west-2).

**Test 1 - Exact version match (no regression):** PASS
- Found `580.126.09-1ubuntu1` in apt, built and installed mock package successfully

**Test 2 - Fallback path (version mismatch):** PASS
- Simulated kmod version `580.999.99` (not in apt)
- Correctly fell back to latest candidate (`580.159.04-1ubuntu1`)
- Version field rewritten to match kmod version
- Package metadata (Provides) preserved from real candidate


The fix correctly handles the real-world scenario where kmod reports a different patch version than what's available in apt. LGTM.

PR #1124, opened 2026-06-06
add-cng: EFA support for HPC-series CPU CNGs (hpc6a / hpc7a / hpc8a, …)

DaisukeMiyamoto:
## Summary

Purely additive opt-in EFA support for HPC-series CPU instances
(hpc6a / hpc7a / hpc7g / hpc6id / hpc8a, c7gn, …) on the on-demand CPU CNG,
plus a server-side `EfaEnabled` toggle on FSx for Lustre. **Every new
parameter defaults to off**, and every code path falls back to the exact
pre-existing behavior when the flag is `false` — verified by a vanilla
regression run.

### Added (all default off)
- **`add-cng.yaml`**: `EnableEfa`, `EfaInterfaceCount` (1 or 2),
  `PlacementGroupName` (empty + `EnableEfa=true` auto-creates a per-CNG
  cluster PG; named reuses an existing one).
- **`pcs-ml-cluster-deploy-all.yaml`**: surfaces the same as
  `OnDemandEnableEfa` / `OnDemandEfaInterfaceCount` /
  `OnDemandPlacementGroupName`. GPU CNGs (P5 / P6-B200 / P6-B300) untouched.
- **`ml-cluster-prerequisites.yaml`**: `FSxLustreEnableEfa` — sets
  `EfaEnabled=true` on the Lustre filesystem (PERSISTENT_2 SSD only). This
  is **server-side only**; the client-side install (Lustre client EFA
  modules / LNet, `nvidia-fs.ko` / `cufile.json` for GDS) is **out of
  scope** and tracked as a follow-up in
  [`docs/ROADMAP.md`](architectures/aws-pcs/docs/ROADMAP.md).

## Validation

Real hardware (us-east-2 / us-west-2):

| Instance | NICs | Spec | osu_mbw_mr | % of spec |
|---|---:|---:|---:|---:|
| hpc6a.48xlarge | 1 | 100 Gbps | 97.6 Gbps | 97.6% |
| hpc7a.96xlarge | 2 | 300 Gbps | 263 Gbps | 87.6% |
| hpc8a.96xlarge | 2 | 300 Gbps | 341 Gbps | 113% (Nitro v6) |

- FSx Lustre `EfaEnabled=true` with `Capacity=19200` — `aws fsx
  describe-file-systems` reports `EfaEnabled=true`; cluster + monitoring +
  Pyxis container test all pass.
- **Regression**: vanilla `deploy-all` with every new parameter at its
  default — no behavior change.

## Docs
- `tests/README.md` Test 9 (HPC EFA) + Test 10 (FSx storage health)
- `README.md` §EFA on CPU HPC instances
- `docs/PARAMETERS.md`: 4 new params
- `docs/ROADMAP.md`: client-side Lustre-on-EFA + GDS follow-up

## Test plan
- [x] `add-cng.yaml` with `EnableEfa=true` on hpc6a / hpc7a / hpc8a
- [x] `deploy-all` with `OnDemandEnableEfa=true` (hpc7a)
- [x] **Regression**: `deploy-all` vanilla, all new params off
- [x] `prerequisites` with `FSxLustreEnableEfa=true` on PERSISTENT_2
- [x] OSU MPI 7.4 multi-pair throughput per type
- [x] Monitoring + Pyxis container test on each configuration


architectures/aws-pcs/README.md
```diff
@@ -159,6 +143,36 @@ automatically.
...
+`OnDemandEnableEfa=true` (deploy-all) or `EnableEfa=true` (modular `add-cng.yaml`).
+The CPU compute node group then launches with EFA `NetworkInterfaces` and a cluster
+placement group is auto-created. GPU CNGs (P5/P6) are unaffected — they use their
+own dedicated multi-NIC EFA wiring per-family.
+
+| Instance type | EFA interfaces | Aggregate spec | Set `OnDemandEfaInterfaceCount` to |
+|---|---:|---:|---:|
+| `hpc6a.48xlarge` | 1 | 100 Gbps | **1** |
+| `hpc7a.96xlarge` (and 12/24/48) | 2 | 300 Gbps | **2** |
+| `hpc7g.16xlarge` (Graviton) | 1 | 200 Gbps | **1** |
```
KeitaW: ### Graviton types (hpc7g / c7gn) are listed as ready-to-use, but the default AMI is x86_64-only — they won't launch

The EFA instance-type table presents `hpc7g.16xlarge (Graviton)` and `c7gn.16xlarge` as supported EFA CPU types alongside the x86 ones, with no AMI caveat. But `hpc7g` and `c7gn` are **arm64** (I confirmed via `aws ec2 describe-instance-types`: both report `ProcessorInfo.SupportedArchitectures = [arm64]`), and the cluster's default `AmiId` auto-resolves the **x86_64** PCS-Ready DLAMI (`/aws/service/pcs/ami/dlami-base-ubuntu2404/x86_64/latest/ami-id`). An x86_64 AMI can't boot arm64 hardware, so a user who follows this table — picks `hpc7g`/`c7gn` and leaves `AmiId` at its default — gets a node group that fails to launch instances, with no hint from the docs as to why.

Interestingly, an **arm64 PCS DLAMI does exist** as a public SSM parameter — I confirmed `/aws/service/pcs/ami/dlami-base-ubuntu2404/arm64/latest/ami-id` resolves today — so this is fixable rather than fundamental. Your own `docs/ROADMAP.md` (line 18-20) already acknowledges arm64 isn't wired in (framed there for future GB200/Grace), which makes the EFA table advertising Graviton *now* an internal inconsistency. Worth noting too that the validation matrix only covers hpc6a/hpc7a/hpc8a (all x86_64) — so hpc7g, c7gn (and x86 hpc6id) are documented-but-untested.

My recommendation is to **scope this to match what the PR actually built and validated**: drop `hpc7g` and `c7gn` from the "supported now" EFA table (and the `OnDemandEnableEfa` / asset-table prose), and add them to `docs/ROADMAP.md` as a Graviton follow-up next to the client-side item already deferred there. The reasoning is that arm64 sits outside this PR's own declared scope on *both* counts: the validation matrix is entirely x86 (hpc6a/hpc7a/hpc8a), and making Graviton boot needs arm64 AMI resolution **plus** arm64 Enroot/Pyxis first-boot software — and client-side software setup is the exact thing this PR explicitly defers to the roadmap. Wiring arm64 here would expand a clean, regression-verified, server-side opt-in into an unvalidated arm64-support workstream that deserves its own PR and its own hardware validation. The arm64 DLAMI does exist (`/aws/service/pcs/ami/dlami-base-ubuntu2404/arm64/latest/ami-id`), so the follow-up is well-defined — it just shouldn't ride along on this one. Net effect: the table only advertises what deploys with defaults, and nobody copies an `hpc7g` example into an opaque launch failure.

KeitaW (commented):
## Review 1/2 — Overview, Live Validation & Documentation

This is a clean, purely-additive PR that brings opt-in EFA to the CPU/HPC compute node groups (`add-cng.yaml` + the deploy-all forwarding) and a server-side `EfaEnabled` toggle to FSx for Lustre — every new parameter defaults off and the `false` path is byte-identical to the prior template, which the included vanilla regression run confirms. I verified the load-bearing technical claims independently against primary AWS sources and the live EC2/SSM APIs, and the core engineering holds up well: the LaunchTemplate `NetworkInterfaces`/EFA shape is the AWS-documented PCS mechanism (not a workaround), all six EFA interface counts are correct, the `NetworkCardIndex`/`DeviceIndex` assignment matches the PCS sample pattern, the param forwarding into the nested CPU CNG stack is wired correctly, and the FSx `19200 GiB` capacity claim is exact. The findings below are not bugs in the EFA wiring — they're guided-path / documentation-accuracy issues: Graviton instance types advertised as usable today that the default x86_64 AMI can't actually boot, and a few spots where the FSx EFA story oversells the benefit for CPU-only fleets or fails quietly.

---

---

## Live Validation (2026-06-09 · us-east-1)

I deployed the PR-branch templates end-to-end on real hardware. Because HPC capacity wasn't
available, I substituted **`c6in.32xlarge`** — x86, **2 network cards / 2 EFA interfaces** (verified
via `describe-instance-types`) — which exercises the identical multi-NIC wiring path as hpc7a/hpc8a
(`EfaInterfaceCount=2`), at a ~200 Gbps aggregate spec instead of 300.

**Core feature — deployed and verified.** `pcs-ml-cluster-deploy-all.yaml` with
`OnDemandEnableEfa=true`, `OnDemandEfaInterfaceCount=2`, 2× `c6in.32xlarge` reached
`CREATE_COMPLETE` (all nested stacks) and:
- **Placement group auto-created** — `…-OnDemandCNGStack-…-PlacementGroup-…`, strategy `cluster`,
  both nodes members. Confirms the `CreatePlacementGroup` condition + `Placement` wiring.
- **2-NIC EFA topology exactly as designed** — each node's interfaces: `NetworkCardIndex 0 /
  DeviceIndex 0 / efa` **and** `NetworkCardIndex 1 / DeviceIndex 1 / efa` (`describe-instances`).
  This is the precise wiring the review verified statically — now confirmed live.
- **OS-level EFA** — on a compute node, `fi_info -p efa` shows two devices (`rdmap0s30`,
  `rdmap0s31`) and `lspci` shows two `Elastic Fabric Adapter (EFA)` controllers. Slurm `efa`
  partition up, `efa-[1-2]` idle.
- **Multi-pair bandwidth confirms the PR's central guidance** — single-pair `osu_bw` peaks **~80
  Gbps** (one libfabric endpoint = one NIC); 16-pair `osu_mbw_mr -np 32 -N 16` peaks **24,007 MB/s
  ≈ 192 Gbps** (~96% of the c6in 200 Gbps spec). The 2.4× jump empirically proves "a single pair
  uses one NIC; use `osu_mbw_mr` multi-pair to exercise both."

**FSx `EfaEnabled` — confirmed.** Standalone `ml-cluster-prerequisites.yaml`,
`FSxLustreEnableEfa=true`, PERSISTENT_2, `Capacity=19200` → `CREATE_COMPLETE`;
`describe-file-systems` reports `EfaEnabled=true` (PUST 250, AUTOMATIC metadata).

**Regression — confirmed.** Vanilla `deploy-all` (all defaults, every new parameter off) reached
`CREATE_COMPLETE` and produced the standard `cpu1` partition (`sinfo`: 4× `idle~` nodes,
single-NIC path) — no behavior change, confirming the additive claim.

**All three findings reproduced empirically:**
- **Finding 1 (Graviton) — confirmed.** `run-instances --dry-run` of the default x86 PCS DLAMI on
  arm64 `c7gn.16xlarge` → `InvalidParameterValue: The architecture 'arm64' … does not match the
  architecture 'x86_64' of the specified AMI`. The identical AMI on x86 `c6in.32xlarge` →
  *"Request would have succeeded"* — isolating the failure to the arch mismatch.
- **Finding 2 (FSx capacity trap) — confirmed.** Prerequisites with `FSxLustreEnableEfa=true`,
  PERSISTENT_2, `Capacity=1200` (the default) → `CREATE_FAILED`:
  *"Invalid storage capacity provided: 1200 GiB. Minimum storage capacity for an EFA enabled LUSTRE
  file systems with deployment type PERSISTENT_2, per unit storage throughput 250 and storage type
  SSD is 19200."* The trap is real, and FSx rejects it with a clear error (as the PR states).
- **Finding 3 (PERSISTENT_1 silent-ignore) — confirmed.** Prerequisites with
  `FSxLustreEnableEfa=true` on PERSISTENT_1 → `CREATE_COMPLETE`, `describe-file-systems` reports
  `EfaEnabled=None`. The opt-in is silently dropped with no warning.

*(Validation note: the first PERSISTENT_1 attempt failed on an unrelated account NAT-gateway-per-AZ
limit, not the PR; re-run in a less-congested AZ succeeded.)*

---

The inline comment below carries the one documentation finding (Graviton instance types).

architectures/aws-pcs/assets/ml-cluster-prerequisites.yaml
```diff
@@ -109,6 +112,24 @@ Parameters:
       - "2.12"
     Default: "2.15"
 
+  FSxLustreEnableEfa:
+    Description: >-
+      (Optional) Enable EFA + GPUDirect Storage (GDS) support on the FSx for
```
KeitaW: ### FSx `EfaEnabled` is primarily a GPUDirect Storage (GPU) feature — the "CPU CNGs benefit" framing oversells it, and the 16× capacity floor is a guided-path trap

The parameter description says enabling this lets "EFA-capable clients (CPU CNGs with `OnDemandEnableEfa=true`, or P5/P6 GPU CNGs)" use "a libfabric/GDS path to access Lustre." Per the FSx for Lustre User Guide, the headline feature here — GPUDirect Storage (GDS) — is a **GPU** capability: it DMAs file data straight into GPU memory and requires GPU instances plus the NVIDIA `nvidia-fs`/cuFile stack. A CPU-only HPC node gets nothing from the GDS half; it would benefit only from the EFA *transport* to storage, and only if it's itself an EFA-capable instance pushing past ~10 GBps. As written, the description reads as if a CPU HPC fleet is a primary beneficiary, which I think will set the wrong expectation.

The bigger operational point is the capacity coupling, which the docs do call out but the template doesn't guard: enabling `EfaEnabled` forces a **16× larger minimum `Capacity`** (19200 GiB vs 1200 at `PerUnitStorageThroughput=250` — I confirmed this is exact against the User Guide). A user who flips `FSxLustreEnableEfa=true` in deploy-all but leaves `Capacity` at its default will have the FSx stack fail at create time. Failing fast is acceptable, but for a one-click deploy-all it means a rolled-back stack after other resources have begun.

I'd suggest (a) reframing the benefit around GPU/GDS clients (P5/P6) as the primary case and EFA-transport for CPU as secondary/conditional, and (b) considering whether deploy-all should nudge `Capacity` to the EFA minimum (or pre-validate it) when the toggle is on, rather than relying on the downstream FSx error. At minimum the GPU-feature framing keeps the docs accurate.

architectures/aws-pcs/assets/ml-cluster-prerequisites.yaml
```diff
@@ -390,6 +416,13 @@ Resources:
           - LustreIsPersistent2
           - Mode: AUTOMATIC
           - !Ref AWS::NoValue
+        # EfaEnabled enables EFA + GPUDirect Storage (GDS) on the FSx-side
+        # endpoints. PERSISTENT_2 only; the FSxLustreEfaEnabled condition gates
+        # both the user toggle and the deployment-type prerequisite.
+        EfaEnabled: !If
```
KeitaW, line 438: ### `FSxLustreEnableEfa=true` on PERSISTENT_1 is silently ignored — prefer surfacing it

The `FSxLustreEfaEnabled` condition gates on `LustreIsPersistent2`, so on a PERSISTENT_1 filesystem the `EfaEnabled` property is simply never emitted and the user gets a fully successful stack with **no EFA and no warning** (the README accurately documents this as "silently ignored on PERSISTENT_1"). The PERSISTENT_2-only gate itself is correct — EFA on Lustre is a P2-only feature — but silently dropping an explicitly-requested capability is the kind of fail-quiet that surprises people later when GDS/EFA isn't there. There's no CloudFormation-native "warn" mechanism, so I don't think there's a clean in-template fix, but it's worth weighing whether the friendlier behavior is to *fail* the stack when `FSxLustreEnableEfa=true` is combined with a non-P2 deployment type (so the misconfiguration is caught at deploy time) rather than honoring the opt-in by ignoring it. If you keep the silent-ignore, the current doc note is the right mitigation — just flagging the design choice.

KeitaW (commented):
## Review 2/2 — Infrastructure Correctness, Sources & What Looks Great

The two inline comments below cover the FSx `EfaEnabled` findings.

---

## Sources

Graviton / AMI architecture:
- EC2 instance architectures — `aws ec2 describe-instance-types --instance-types hpc7g.16xlarge c7gn.16xlarge` → `arm64` (verified live, 2026-06-09)
- PCS DLAMI SSM parameters — `aws ssm get-parameters-by-path --path /aws/service/pcs/ami --recursive` → both `.../x86_64/latest/ami-id` and `.../arm64/latest/ami-id` exist (verified live, 2026-06-09)

EFA interface counts (all six PR claims confirmed):
- `aws ec2 describe-instance-types` → MaxNetworkCards / EfaInfo.MaximumEfaInterfaces: hpc6a=1, hpc7a=2, hpc7g=1, hpc6id=2, hpc8a=2, c7gn=1 (verified live, 2026-06-09)
- EC2 EFA-supported instance types: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa.html

PCS LaunchTemplate / EFA networking contract:
- AWS PCS User Guide — LaunchTemplate network interfaces & security groups: https://docs.aws.amazon.com/pcs/latest/userguide/
- aws-samples/aws-hpc-recipes (PCS EFA launch-template precedent): https://github.com/aws-samples/aws-hpc-recipes/tree/main/recipes/pcs

FSx for Lustre EFA / GDS + capacity:
- FSx for Lustre User Guide — EFA & GPUDirect Storage, minimum capacity ("multiples of 19200 GiB for 250 MBps/TiB"): https://docs.aws.amazon.com/fsx/latest/LustreGuide/efa.html
- FSx CreateFileSystem API (non-EFA SSD minimum 1200 GiB): https://docs.aws.amazon.com/fsx/latest/APIReference/API_CreateFileSystem.html

Multi-NIC MPI:
- OSU Micro-Benchmarks (`osu_mbw_mr` multi-pair semantics): https://mvapich.cse.ohio-state.edu/benchmarks/

---

---

## Things That Look Great

- **Genuinely additive, with a regression run to prove it.** Every new parameter defaults off and the `EfaEnabled=false` branch preserves the exact prior `SecurityGroupIds` form — the `!If [EfaEnabled, …]` structure keeps existing deployments untouched, and the vanilla `deploy-all` regression in the test plan backs the claim. This is the right way to land an opt-in feature.
- **The EFA wiring is idiomatic PCS, not a hack.** Hand-authoring the LaunchTemplate `NetworkInterfaces` array with `InterfaceType=efa` is exactly the AWS-documented PCS mechanism (PCS has no `EfaInterfaceCount` shortcut), and dropping top-level `SecurityGroupIds` in favor of per-interface `Groups` is *required* by PCS when an LT defines network interfaces — the PR does this correctly via the condition.
- **Every EFA interface count is correct** — I checked all six against `describe-instance-types` and hpc6a=1 / hpc7a=2 / hpc7g=1 / hpc6id=2 / hpc8a=2 / c7gn=1 all match, including the brand-new Nitro-v6 hpc8a. The `NetworkCardIndex 1 / DeviceIndex 1` assignment for the second NIC also matches the PCS sample pattern.
- **The param forwarding is fully wired.** `deploy-all` correctly threads `OnDemandEnableEfa/EfaInterfaceCount/PlacementGroupName` into the **CPU** `OnDemandCNGStack` (not login, not the P-series GPU stack) and `FSxLustreEnableEfa` into the prerequisites stack — no declared-but-dead parameters.
- **Asset reuse done right.** Test 6/9 `wget` the canonical `micro-benchmarks/nccl-tests/slurm/nccl-tests-container.sbatch` rather than re-vendoring it, and the OSU build pins a fixed version (`osu-micro-benchmarks-7.4.tar.gz`) — no `latest`. The repo has no existing OSU asset, so building it inline isn't duplication.
- **The multi-NIC benchmarking guidance is correct and genuinely useful.** The "a single MPI pair uses one libfabric endpoint → one NIC, so use `osu_mbw_mr -np 32 -N 16` to exercise both NICs" explanation is accurate, and Test 9 uses Open MPI `mpirun -np 32 -N 16` (16 procs/node × 2 nodes = 16 inter-node pairs) — exactly the right pattern to saturate a 2-NIC hpc7a/hpc8a. The honest 113%-of-spec hpc8a footnote (forward-direction vs per-direction sustained) is a nice touch.
- **Strong docs overall** — the PARAMETERS additions, the per-type EFA-count table, the placement-group sharing story (with the `PlacementGroupName` stack output for cross-CNG sharing), and the explicit out-of-scope/ROADMAP note for client-side Lustre-on-EFA all make this easy to consume and review.

architectures/aws-pcs/README.md
DaisukeMiyamoto: Thanks — agree, scoping to what was actually validated is the right call. Pushed `cfb5a345`:

- Removed `hpc7g.16xlarge` and `c7gn.16xlarge` from the EFA table in `README.md` (and from the surrounding prose / table-of-supported-instances in `docs/PARAMETERS.md`, `add-cng.yaml`, `pcs-ml-cluster-deploy-all.yaml`).
- Added a **Graviton (arm64) CPU CNG support** follow-up to `docs/ROADMAP.md` next to the existing Grace-Blackwell (arm64) entry — it captures that the arm64 PCS DLAMI exists at `/aws/service/pcs/ami/dlami-base-ubuntu2404/arm64/latest/ami-id`, that arch-aware `AmiId` resolution + an arm64 Enroot/Pyxis first-boot path are still missing, and that hpc7g + c7gn need end-to-end validation on real hardware.

So the docs now only advertise what deploys with defaults, with the arm64 work scoped as a separate, well-defined follow-up PR.

architectures/aws-pcs/assets/ml-cluster-prerequisites.yaml
DaisukeMiyamoto: Reframed in `cfb5a345`. The new `FSxLustreEnableEfa` description (consistent across `README.md`, `docs/PARAMETERS.md`, `ml-cluster-prerequisites.yaml`, `pcs-ml-cluster-deploy-all.yaml`) reads:

> Enable EFA on the FSx for Lustre filesystem. **The headline feature is GPUDirect Storage (GDS) for P5/P5e/P5en/P6-B200 GPU clients**, which DMAs file data straight into GPU memory (requires the NVIDIA `nvidia-fs` / cuFile stack on the client — tracked as a follow-up in `docs/ROADMAP.md`). EFA-capable CPU CNGs (`OnDemandEnableEfa=true`) get the EFA *transport* path to storage as a secondary benefit, useful when a single client is pushing past ~10 GBps.

So GPU/GDS is the primary case, CPU EFA-transport is explicitly secondary/conditional, and the missing client-side software is named.

For the capacity coupling, I addressed it with the fail-fast Rule below (Finding 3) but kept Capacity itself as a user choice rather than auto-nudging. The reasoning is that the EFA-minimum varies with `PerUnitStorageThroughput` (19200 at 250, 9600 at 500, 4800 at 1000), so any nudge would either pin to a single tier or grow into a derived-default expression that's hard to read at the param-table level. A user who flips the toggle hits a clear FSx-side error pointing at the exact required minimum within ~2 minutes of stack create — which is fail-fast enough that the Capacity-prevalidation feels like over-engineering. Happy to revisit if the deploy-all UX feels rough in practice.

architectures/aws-pcs/assets/ml-cluster-prerequisites.yaml
DaisukeMiyamoto: Good catch — agree that fail-quiet is the wrong default here. Switched to fail-fast in `cfb5a345`.

Added a CFN Rule (`FSxLustreEfaRequiresPersistent2`) on **both** `ml-cluster-prerequisites.yaml` and `pcs-ml-cluster-deploy-all.yaml`. Putting it on the parent (deploy-all) too means the validation fires at the parent stack's create time — before any nested resources are launched — instead of after rollback from the prerequisites stack.

The defense-in-depth Condition is kept (`FSxLustreEfaEnabled = And(EnableEfa=true, IsPersistent2)`) so the prerequisites template still synthesizes correct resources even if the Rule is bypassed (e.g. someone copies the YAML into a different harness without the Rules block).

Verified live on us-east-2 — `create-stack` with `FSxLustreEnableEfa=true` + `LustreDeploymentType=PERSISTENT_1` now fails immediately with:

> Parameter validation failed: assertion error: FSxLustreEnableEfa=true requires LustreDeploymentType=PERSISTENT_2 (EFA on FSx for Lustre is a PERSISTENT_2-only feature).

Documentation updated to match (the silent-ignore wording is gone from `README.md`, `docs/PARAMETERS.md`, and `tests/README.md`).

DaisukeMiyamoto:
Thank you for the thorough review and the live us-east-1 reproduction — independent verification of the core wiring (placement group, NIC topology, Slurm partition, OS-level EFA, multi-pair bandwidth) and of all three findings was hugely useful.

Pushed `cfb5a345` addressing all three:

- **Finding 1 (Graviton)** — `hpc7g` / `c7gn` removed from the README EFA table, prose, and template parameter descriptions; new "Graviton (arm64) CPU CNG support" entry in `docs/ROADMAP.md` next to the existing Grace-Blackwell entry. Docs now only advertise the x86 types the PR actually validated. Inline reply: https://github.com/awslabs/awsome-distributed-ai/pull/1124#discussion_r3380129178
- **Finding 2 (FSx framing)** — `FSxLustreEnableEfa` description reframed across all four files: GPU/GDS as primary (with the missing nvidia-fs / cuFile client stack named and pointed at the ROADMAP follow-up), CPU EFA-transport explicitly secondary. Inline reply: https://github.com/awslabs/awsome-distributed-ai/pull/1124#discussion_r3380129274
- **Finding 3 (PERSISTENT_1 silent-ignore)** — replaced with a fail-fast CFN Rule (`FSxLustreEfaRequiresPersistent2`) on **both** the prerequisites and deploy-all templates so the parent fails at create time rather than after a nested rollback. Verified live on us-east-2 — `create-stack` with `FSxLustreEnableEfa=true` + `LustreDeploymentType=PERSISTENT_1` now fails immediately with a clear assertion message. Inline reply: https://github.com/awslabs/awsome-distributed-ai/pull/1124#discussion_r3380129383

Diff stat for the response: 7 files, +82 / -43. All template changes were template-validate-clean and the new Rule was confirmed to fire end-to-end.

KeitaW (approved):
LGTM!

PR #1128, opened 2026-06-11
Add Sglang Inference Examples 

ybalbert001:
## Purpose

Provide more references for LLM inference on AWS — add a set of self-contained SGLang serving examples for Amazon EKS / SageMaker HyperPod EKS, covering both prefill/decode (PD) disaggregation and unified
serving across H200 and B300 hardware.

## Changes

1. Kimi2.6 on SGLang (node-level 1P1D) — examples/inference/sglang/kimi2.6-h200-1p1d/. Prefill + decode StatefulSets across 2× ml.p5en.48xlarge (16× H200), KV cache transferred over NIXL (LIBFABRIC/EFA),
fronted by the SGLang router. Image pinned to SGLang v0.5.12.post1. README documents the known NIXL 1.2.0 KV-transfer issue.
2. DeepSeek V4 Pro on B300 — examples/inference/sglang/dsv4pro-b300-single-node/. Unified (non-PD) baseline on a single B300 (8 GPU) node.
3. Qwen3.5-27B on B300 (intra-node PD) — examples/inference/sglang/qwen3.5-27b-b300-intra-pd/. 6 prefill + 2 decode in one pod on a single B300 node, NIXL, with an SGLang router sidecar.
4. Shared SGLang helpers — model pre-staging to local NVMe (download-model.sh + daemonset).

## Test Plan

Environment:
- AWS Service: SageMaker HyperPod EKS (1.33+)
- Instance type: ml.p5en.48xlarge (H200), ml.p6-b300.48xlarge (B300)
- Number of nodes: 2 (Kimi 1P1D); 1 (DeepSeek V4 Pro, Qwen3.5-27B)

## Test Results

▎ ⚠️  Not yet measured. The per-model READMEs currently carry placeholder (TBD) latency/throughput tables. Benchmark numbers (sglang.bench_serving sweep — burst RPS, P50 TTFT/TPOT, tok/s/GPU) need to be filled
▎ in before these are quoted downstream. The Kimi 1P1D example is also marked [draft-1] pending a working NIXL 1.2.0 KV-transfer path.

## Directory Structure

examples/inference/
└── sglang/
    ├── README.md                        # engine overview + shared helpers
    ├── download-model.sh / *-daemonset.yaml
    ├── dcgm-exporter-daemonset.yaml
    ├── setup-amp-monitoring.sh / prometheus-agent-amp.yaml
    ├── kimi2.6-h200-1p1d/               # Dockerfile, build-image.sh, README, kimi-pd-deploy.yaml
    ├── dsv4pro-b300-single-node/        # README, dsv4pro-deploy.yaml
    └── qwen3.5-27b-b300-intra-pd/       # README, qwen-pd-deploy.yaml

## Checklist

- [x] I have read the contributing guidelines (https://github.com/awslabs/awsome-distributed-training/blob/main/CONTRIBUTING.md).
- [x] I am working against the latest main branch.
- [x] External dependencies are pinned to a specific version or tag (SGLang v0.5.12.post1; no latest).
- [x] New test cases follow the expected directory structure (#directory-structure).


examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml
```diff
@@ -0,0 +1,237 @@
...
+            --disaggregation-bootstrap-port 9005 \
+            --mem-fraction-static 0.85 \
+            --reasoning-parser qwen3 --tool-call-parser qwen3_coder \
+            --context-length 32768 \
+            --disable-flashinfer-autotune \
+            --skip-server-warmup \
+            --enable-metrics
+      # ---- PD router (shares pod network → reaches engines on 127.0.0.1) ---
+      - name: router
+        image: lmsysorg/sgl-model-gateway:latest
```
KeitaW: ### `:latest` image tag on the router sidecar

`lmsysorg/sgl-model-gateway:latest` (and `prom/prometheus:latest`, and the DeepSeek router) pulls a moving tag. The repo convention is fixed tags everywhere so deployments are reproducible and air-gapped clusters don't break — I'd pin to a released tag (ideally a digest). The SGLang *engine* image is already pinned correctly; only the sidecars float.

examples/inference/sglang/prometheus-agent-amp.yaml
```diff
@@ -0,0 +1,199 @@
...
+      app: prometheus-agent
+  template:
+    metadata:
+      labels:
+        app: prometheus-agent
+    spec:
+      serviceAccountName: amp-iamproxy-ingest-service-account
+      containers:
+      - name: prometheus
+        image: prom/prometheus:latest
```
KeitaW: ### `prom/prometheus:latest`

Same `:latest` concern — pin Prometheus to a released tag so the agent is reproducible. (This pin is also the root cause of the agent-runs-in-server-mode bug flagged in the Deployment batch: `:latest` is now 3.x, where the `--enable-feature=agent` flag was removed.)

KeitaW (commented):

Themed batch covering structure, reuse, and image hygiene. Two inline comments accompany this batch (the `:latest` tags); the rest are cross-cutting and live here in the body.

### Observability duplicates an existing repo asset — reuse it instead of shipping a parallel stack
**Files:** `prometheus-agent-amp.yaml`, `setup-amp-monitoring.sh`, `dcgm-exporter-daemonset.yaml`, READMEs

This PR adds ~400 lines of bespoke monitoring (in-cluster Prometheus agent → AMP via SigV4, an IAM/workspace setup script, a DCGM DaemonSet). The repo already solves this at [`4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/`](https://github.com/awslabs/awsome-distributed-ai/tree/main/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability) via an **ADOT collector** (`adot-collector-prometheus.yaml`, `deploy-obs.sh`/`cleanup-obs.sh`, DCGM configs, dashboards). Per the checklist's "Reuse existing assets," I'd drop the bespoke files and have the READMEs link to that as a prerequisite.

Two even-lower-maintenance managed paths, which I'd recommend over both:
- **Plain EKS:** the [AWS managed collector / agentless scraper](https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-collector.html) (`aws amp create-scraper`) — AWS runs the scraper outside the cluster (no in-cluster agent to deploy/patch/HA). It's the only option in the EKS console's "Turn on Prometheus metrics" wizard, and covers everything this agent does for SGLang serving.
- **HyperPod EKS:** the [SageMaker HyperPod observability add-on](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-hyperpod-observability-addon.html) — managed DCGM/metrics + AMP + Grafana dashboards (add one scrape job for SGLang's `:30000/metrics`).

Adopting any managed path also retires four findings on `prometheus-agent-amp.yaml` (the server-mode bug, the `:latest` pin, the DCGM root context, and the monitoring share of the privileged concern).

### Use one shared SGLang image across all three models instead of three divergent ones
**Files:** `dsv4pro-b300-single-node/dsv4pro-deploy.yaml`, `qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml`, `kimi2.6-h200-1p1d/{Dockerfile,build-image.sh,kimi-pd-deploy.yaml}`

The three examples ship three different engine images for the same SGLang runtime: Qwen → `lmsysorg/sglang:v0.5.12.post1-cu130` (pinned, no EFA layer); Kimi → a custom ECR build (pinned base **+ EFA installer**); DeepSeek → `lmsysorg/sglang:deepseek-v4-b300` (opaque, unversioned tag). The model is a **runtime arg** (`--model-path`), so one image serves all three.

`deepseek-v4-b300` is worth flagging on its own: it's not a version, it was pushed **2026-04-29 — a month before** the `v0.5.12.post1-cu130` release the PR pins to, and it's larger (18.1 GB vs 13.0 GB). So the DeepSeek example runs an **older, non-pinned build with an unknown NIXL version**, quietly contradicting the PR's "pinned for NIXL 1.1.0" rationale.

**Suggestion:** promote the Kimi `Dockerfile` (pinned base + EFA installer) to the shared `examples/inference/sglang/` level, build once, and have all three manifests reference it — selecting the model via `--model-path`. Bonus: gives Qwen the EFA/LIBFABRIC path it currently lacks. If `deepseek-v4-b300` carries a needed patch, please state what.

### `build-image.sh` is missing its shebang and the MIT-0 license header
**File:** `kimi2.6-h200-1p1d/build-image.sh`

It starts straight at `algorithm_name=sgl-dev-cu13` — no `#!/usr/bin/env bash`, no copyright header (and no `set -euo pipefail`, see the Deployment batch). The sibling `download-model.sh` / `setup-amp-monitoring.sh` get this right. Suggested top-of-file:
```bash
#!/usr/bin/env bash
set -euo pipefail
```

examples/inference/sglang/prometheus-agent-amp.yaml
```diff
@@ -0,0 +1,199 @@
...
+        args:
+          - '--config.file=/etc/prometheus/prometheus.yml'
+          - '--storage.tsdb.path=/prometheus'
+          - '--storage.tsdb.retention.time=2h'
+          - '--enable-feature=agent'
```
KeitaW: ### Runs in **server mode, not agent mode** as written

`prom/prometheus:latest` is now **3.12.0**, and Prometheus removed `--enable-feature=agent` at 3.0 (it became `--agent`). I ran it to confirm — on 3.12 the manifest logs `WARN "Unknown option for --enable-feature" option=agent` then `Starting Prometheus Server mode=server`. So the flag is silently ignored and Prometheus starts as a **full server with a local TSDB** (written to the `emptyDir`), not the lightweight agent the file's header describes. It doesn't crash (remote_write still works in server mode) so it looks healthy while doing the wrong thing. Pin a real version and fix the flag (also pin the image per the Structure batch):

```suggestion
          - '--agent'
```

The ConfigMap (scrape_configs + remote_write, no alerting/rules) is agent-mode-compatible. Adjacent notes: `replicas: 1` is a SPOF (an HA pair needs `cluster`+`__replica__` labels or AMP bills 2×), and the WAL on `emptyDir` loses buffered samples on eviction — both moot if you adopt a managed path.

examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml
```diff
@@ -0,0 +1,237 @@
...
+      # ---- 6 prefill + 2 decode engines, all on one node ------------------
+      - name: sglang-pd
+        image: lmsysorg/sglang:v0.5.12.post1-cu130
+        imagePullPolicy: IfNotPresent
+        securityContext:
+          privileged: true
+          capabilities:
+            add: ["SYS_NICE"]
+        env:
+        - name: NIXL_LOG_LEVEL
```
KeitaW, line 82: ### Benign "use LIBFABRIC" warning — worth one README line (nit)

Not a correctness issue. No `SGLANG_DISAGGREGATION_NIXL_BACKEND` is set, so NIXL uses **UCX** — which is correct here (single-node intra-node PD; the KV hop never crosses EFA). When I ran this on a B300 node it served correctly, but the engine logs `16 Amazon EFA(s) were detected, but the UCX backend was configured ... recommended to use the LIBFABRIC backend instead`. A one-line README note ("intra-node PD uses UCX; the EFA-detected warning is expected/benign") saves the next person from chasing it. No manifest change needed.

examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
```diff
@@ -0,0 +1,381 @@
...
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "30000"
+        prometheus.io/path: "/metrics"
+    spec:
+      nodeSelector:
+        node.kubernetes.io/instance-type: ml.p5en.48xlarge
+      containers:
+      - name: sglang
+        image: <YOUR_ECR_IMAGE>   # the URI build-image.sh printed, e.g. <account>.dkr.ecr.<region>.amazonaws.com/sgl-dev-cu13:<tag>
+        imagePullPolicy: Always
```
KeitaW: ### `imagePullPolicy: Always` forces a pull on every start

Both the prefill and decode StatefulSets use `Always` (also line 211), which re-pulls a multi-GB CUDA image on every pod start and breaks air-gapped clusters. The image is a pinned ECR tag, so `IfNotPresent` is right.

```suggestion
        imagePullPolicy: IfNotPresent
```

KeitaW (commented):
## Review Batch 2/5 — Deployment Pipeline & K8s Operational Correctness

Operational-correctness batch. Three inline comments accompany it (the Prometheus server-mode bug, the Qwen UCX note, the Kimi `imagePullPolicy`); the cross-cutting items are below.

### `nodeSelector` pins the `ml.`-prefixed instance type — only matches HyperPod, not plain EKS
**Files:** `kimi-pd-deploy.yaml` (91, 200), `dsv4pro-deploy.yaml` (48), `qwen-pd-deploy.yaml` (51), `download-model-daemonset.yaml` (31, via `${INSTANCE_TYPE}`)

Every serving manifest pins `node.kubernetes.io/instance-type: ml.p5en.48xlarge` / `ml.p6-b300.48xlarge`. That `ml.` prefix is the **HyperPod** instance-group form; a **plain EKS** managed nodegroup labels the same key with the bare EC2 type (`p6-b300.48xlarge`). I verified the bare form on a real B300 EKS node and a HyperPod EKS system node — the pod sits `Pending` on plain EKS until changed. Since the READMEs advertise "EKS / HyperPod EKS," it should run on both. A `nodeSelector` can't express OR; `nodeAffinity` with `In` can (same key, two values):
```yaml
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node.kubernetes.io/instance-type
                operator: In
                values:
                - p6-b300.48xlarge      # plain EKS managed nodegroup
                - ml.p6-b300.48xlarge   # SageMaker HyperPod instance group
```
(Use the `p5en` pair for Kimi.) I confirmed this scheduled+served on a plain-EKS B300 node. No single cross-environment label is reliable (`sagemaker.amazonaws.com/instance-type` is HyperPod-only; `nvidia.com/gpu.product` needs the GPU Operator, which HyperPod doesn't run by default).

### A DaemonSet is the wrong primitive for model-weight download
**Files:** `download-model-daemonset.yaml`, `download-model.sh`

A DaemonSet can't model a one-shot "download once and stop": its pods must use `restartPolicy: Always`, so a download that exits 0 is **restarted in a loop**, and on node scale-out/replacement it **re-downloads** — with no completion signal beyond a manual `kubectl delete daemonset`. It's also novel here: on this base branch every model/dataset download is a **Job or script** (`examples/training/{verl,nemo-rl,optimum-neuron}`); the only DaemonSets are the EFA/health exporters. The sibling [`examples/inference/vllm/dsv3-uccl-nixl`](https://github.com/awslabs/awsome-distributed-ai/tree/main/examples/inference/vllm/dsv3-uccl-nixl) (which the Kimi README modeled on) uses **no downloader at all** — the serving pod downloads on first start to `HF_HOME` on local NVMe, with **FSx for Lustre** documented as the alternative (~680 GB loads in ~5 min local; +1–2 min on FSx first node). Recommend: single-node → download-on-startup or an `initContainer`; 2-node Kimi → `initContainer`/run-once Job, or FSx (download once, mount RO on both). Local-NVMe staging itself is fine — it's the DaemonSet wrapper that's wrong; "use a Job" is *not* the fix (a Job can't run one-pod-per-node).

### `build-image.sh` lacks `set -euo pipefail` and quotes on some expansions
**File:** `kimi2.6-h200-1p1d/build-image.sh`

Without `set -euo pipefail`, a failed `docker build`/`aws ecr` doesn't stop the script — it continues to `docker push` a stale/missing image. A couple of expansions are unquoted (`--region $region`, `-t ${algorithm_name}`). Add strict mode (see the header suggestion in the Structure batch) and quote the expansions.

### No `livenessProbe` on the serving deployments
**Files:** all four serving manifests

They define a `readinessProbe` but no `livenessProbe`. A wedged engine (CUDA hang, NIXL stuck in `KVPoll.WaitingForInput`) stops passing readiness and drops out of rotation, but nothing restarts the pod — it sits `Running` forever. Add a `livenessProbe` hitting `/health` on the foreground engine port.

examples/inference/sglang/dcgm-exporter-daemonset.yaml
```diff
@@ -0,0 +1,96 @@
...
+      containers:
+      - name: dcgm-exporter
+        image: nvidia/dcgm-exporter:3.3.5-3.4.0-ubuntu22.04
+        ports:
+        - name: metrics
+          containerPort: 9400
+          hostPort: 9400
+        securityContext:
+          runAsNonRoot: false
+          runAsUser: 0
```
KeitaW: ### DCGM runs as root — add a one-line rationale

`runAsNonRoot: false` + `runAsUser: 0` is the standard DCGM requirement (host GPU/driver access), so this is likely fine — but the checklist asks for a rationale comment so a future reader doesn't "fix" it. A short `# DCGM requires root for host GPU access` above the `securityContext` does it.

KeitaW (commented):
## Review Batch 3/5 — Infrastructure, NCCL & Container Security

Container-security and NCCL batch. One inline comment (DCGM root rationale); the rest below.

### `privileged: true` on all SGLang serving containers
**Files:** `dsv4pro-deploy.yaml` (54), `qwen-pd-deploy.yaml` (58), both Kimi StatefulSets

Every engine container runs fully privileged — all capabilities, host device access, isolation off. The workload needs *some* elevation (EFA/RDMA pinned memory, GPUDirect via `gdrdrv`, `SYS_NICE`), but `privileged: true` is a much bigger hammer. I'd scope to the specific capabilities (`IPC_LOCK`, `SYS_NICE`) plus the device mounts already declared, and drop `privileged`. If full privilege is genuinely required for the NIXL/EFA path, a one-line comment stating why makes it a deliberate, reviewable choice.

### `NCCL_SOCKET_IFNAME` is unset (minor, multi-node Kimi only)
**File:** `kimi2.6-h200-1p1d/kimi-pd-deploy.yaml`

Moot for single-node (TP on NVLink) and the cross-node KV hop uses NIXL not NCCL — low priority. But for the 2-node Kimi setup, if NCCL ever initializes a cross-node control channel, the default interface pick can land on the wrong NIC. Cheap guard: set `NCCL_SOCKET_IFNAME=^lo` (exclusion form, never positive selection like `eth0`). See the [EFA cheatsheet](https://github.com/awslabs/awsome-distributed-ai/blob/main/architectures/efa-cheatsheet.md).

KeitaW (commented):
## Review Batch 4/5 — Documentation Consistency

Documentation batch (no inline comments — both items are PR-level).

### PR description lists a CloudFormation change that isn't in the diff
The PR body's change #5 says "repoint 7 CloudFormation 1-click deploy templates at the renamed awsome-distributed-ai S3 bucket," but the diff touches only files under `examples/inference/sglang/` (15 files) — no CloudFormation templates. Either that work was dropped from this branch or belongs to a different PR; I'd reconcile the description with the diff so reviewers and the changelog aren't misled.

### Base branch is `worktree-repo-reorg`, not `main`
This PR targets `worktree-repo-reorg`. If that's intentional stacking onto an in-flight reorg, all good — just worth confirming the merge target is what you want.

KeitaW (commented):
## Review Batch 5/5 — Evaluation, Positives & Sources

Final batch — evaluation framing, what's great, and the source list.

### Placeholder benchmark tables — framing is correct, just don't quote them yet
The per-model READMEs ship `TBD` latency/throughput tables but explicitly label them "Not yet measured" and the Kimi example "[draft-1]." That honest framing is exactly right — this is "validates the deployment runs," not "reproduces a published result," so no methodology obligation is triggered. Just make sure the TBD numbers aren't quoted downstream until filled. My e2e run confirms the Qwen example serves correctly end-to-end.

## Things That Look Great
- **Directory placement is correct** — `examples/inference/sglang/<model>/` follows the RFC #1056 "library is the demo subject" rule, with shared helpers one level up.
- **Excellent shared-helper reuse** — `download-model.sh`, the DCGM daemonset, and the AMP monitoring are factored out and reused across all three examples; extracting the inlined Prometheus block into shared manifests was the right call.
- **Exemplary engine-image pinning with rationale** — `lmsysorg/sglang:v0.5.12.post1-cu130` is pinned *and* the README explains why (NIXL 1.1.0 vs the nightly's 1.2.0).
- **`download-model.sh` / `setup-amp-monitoring.sh`** have the shebang, MIT-0 header, `set -euo pipefail`, and a properly whitelisted `envsubst`.
- **Honest results framing** — "Not yet measured" / "[draft-1]" instead of fabricated numbers.
- **Thorough READMEs** — prerequisites, quick start, monitoring, and a genuinely useful "known issues" section (the DCGM `nvidia.com/gpu.present` node-label caveat is a real gotcha).
- **It works** — I deployed `qwen3.5-27b-b300-intra-pd` (adapted for plain EKS) on a real B300 node; it came up `2/2 Ready` in ~6 min and served a correct completion through the PD router. The intra-node prefill/decode + NIXL topology is sound.

## Sources
**Kubernetes:** [node affinity](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity), [probes](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes), [imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy), [SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/), [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/), [Job](https://kubernetes.io/docs/concepts/workloads/controllers/job/), [initContainers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/).
**AWS:** [AMP ingest](https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-onboard-ingest-metrics.html), [managed scraper](https://docs.aws.amazon.com/prometheus/latest/userguide/AMP-collector.html), [HyperPod observability add-on](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-hyperpod-observability-addon.html), [ADOT on EKS](https://docs.aws.amazon.com/eks/latest/userguide/cw-adot-collector.html), [FSx for Lustre CSI](https://docs.aws.amazon.com/eks/latest/userguide/fsx-csi.html).
**Prometheus:** [agent mode feature flag](https://prometheus.io/docs/prometheus/latest/feature_flags/#prometheus-agent), [3.0 migration](https://prometheus.io/docs/prometheus/latest/migration/). Verified live: `prom/prometheus:latest` == 3.12.0 → `mode=server` on `--enable-feature=agent` (2026-06-11).
**Repo precedent:** [eks-managed-observability](https://github.com/awslabs/awsome-distributed-ai/tree/main/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability), [dsv3-uccl-nixl](https://github.com/awslabs/awsome-distributed-ai/tree/main/examples/inference/vllm/dsv3-uccl-nixl), download-Job precedents under `examples/training/{verl,nemo-rl,optimum-neuron}`.

*Several findings here are evidence-backed by a live B300 deploy and a `prom/prometheus:latest` flag test, plus dedicated research on weight-staging and observability patterns.*

KeitaW (changes requested):
Few comments

ybalbert001:
> ### `build-image.sh` is missing its shebang and the MIT-0 license header
> **File:** `kimi2.6-h200-1p1d/build-image.sh`
> 
> It starts straight at `algorithm_name=sgl-dev-cu13` — no `#!/usr/bin/env bash`, no copyright header (and no `set -euo pipefail`, see the Deployment batch). The sibling `download-model.sh` / `setup-amp-monitoring.sh` get this right. Suggested top-of-file:
> 
> ```shell
> #!/usr/bin/env bash
> # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
> # SPDX-License-Identifier: MIT-0
> set -euo pipefail
> ```
Keita, Thanks for your careful review.

1. At first, I will remove all duplicated files for Observability.  

2. DeepSeek-V4-Pro only works on lmsysorg/sglang:deepseek-v4-b300, while 'deepseek-v4-b300' is a special and fixed tag, so it cannot share an image with Qwen and Kimi.

lmsysorg/sglang:v0.5.12.post1-cu130 is built on source code of sglang , but lmsysorg/sglang:deepseek-v4-b300 involve lots of dedicated modifications (https://hub.docker.com/layers/lmsysorg/sglang/deepseek-v4-b300/images/sha256-049680cf295483bb5f9bb3d53c005f5261f6563ad59bbff785dcc8cf4fc15038#L78).

And, I don't think Qwen and Kimi should share the same image, as Qwen is a single-node deployment and therefore does not require EFA. Since these recipes are organized under the example/ folder, keeping them separate would better illustrate the distinction between intra-node deployment and inter-node deployment. Some descriptions were added as below. 
Screenshot 2026-06-17 at 17 55 33


3. shebang and the MIT-0 license header will be added.

examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml
ybalbert001: OK

examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml
ybalbert001: OK

examples/inference/sglang/prometheus-agent-amp.yaml
ybalbert001: no longer existed.

examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
ybalbert001: OK

ybalbert001:
> A DaemonSet is the wrong primitive for model-weight download

I've investigated this. Both Job and Init Container cannot guarantee downloading models on each node. When deploying across multiple nodes, DaemonSet allows each node to download a copy of model weights to nvme. If multiple pods are scheduled on a single node, it may cause the model to be downloaded repeatedly.

In download-model-daemonset.yaml, the last line uses `sleep infinity`, which prevents the download from being executed repeatedly.

examples/inference/sglang/prometheus-agent-amp.yaml
ybalbert001: deleted

examples/inference/sglang/dcgm-exporter-daemonset.yaml
ybalbert001: removed

ybalbert001: I also saw 'privileged: true' in vllm example. I didn't have the related knowledge about that. Regarding this sglang deployment, I don't think it's risky, maybe narrow it down later.

ybalbert001:
> ## Review Batch 4/5 — Documentation Consistency
> Documentation batch (no inline comments — both items are PR-level).
> 
> ### PR description lists a CloudFormation change that isn't in the diff
> The PR body's change #5 says "repoint 7 CloudFormation 1-click deploy templates at the renamed awsome-distributed-ai S3 bucket," but the diff touches only files under `examples/inference/sglang/` (15 files) — no CloudFormation templates. Either that work was dropped from this branch or belongs to a different PR; I'd reconcile the description with the diff so reviewers and the changelog aren't misled.
> 
> ### Base branch is `worktree-repo-reorg`, not `main`
> This PR targets `worktree-repo-reorg`. If that's intentional stacking onto an in-flight reorg, all good — just worth confirming the merge target is what you want.

Documentation has been updated. And `worktree-repo-reorg` is my target branch.

examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml
ybalbert001: pined to lmsysorg/sgl-model-gateway:v0.3.2

KeitaW (commented):
## Review (round 2) — Batch 1/5: Image Consolidation (must-fix)

_Round-2 re-review of head `b1172bfc`. Thanks for the thorough round-1 pass — most of it is resolved (see table). One finding I had conceded in round 1 is reopened as a **must-fix** after empirical testing on B300; details below._

## Summary

Thanks for the thorough pass on round 1 — most of it is addressed. The bespoke AMP/DCGM observability stack is gone (retiring four findings at once), the serving manifests now schedule on both plain EKS and HyperPod via `nodeAffinity` (with a genuinely clear explanatory comment), liveness probes and the `^lo` NCCL exclusion are in, `build-image.sh` is hardened, and the docs got the engine-image column. I concede the DaemonSet point — a DaemonSet really is the natural one-copy-per-node primitive for the multi-node case (R2-D).

One **must-fix** is back, however: image consolidation (**R2-G**). In round 1 I withdrew it on your rationale that DeepSeek-V4-Pro *needs* its dedicated `deepseek-v4-b300` image — I then verified that empirically and it does **not** hold. All three models serve on a single `stock + EFA` image; DeepSeek-V4-Pro runs **bit-identically** on the unified image vs the dedicated one on real B300 hardware (test details below). Carrying three divergent base images for three examples is a maintenance burden we shouldn't take on when one image demonstrably covers all of them. The rest is minor: three cleanup slips (stale README section, one un-pinned router tag, one download-path nodeSelector) plus `privileged` deferred.

---

|---|---|---|
| 1 | Observability duplicates an existing repo asset | ✅ Resolved — AMP/DCGM manifests deleted (but see R2-A: README still references them) |
| 2 | Use one shared SGLang image across all models | ❌ **Reopened (must-fix)** — round-1 concession empirically refuted: DeepSeek-V4-Pro serves bit-identically on the unified image on B300; one image covers all three (R2-G) |
| 3 | `build-image.sh` missing shebang + MIT-0 header | ✅ Resolved — shebang, MIT-0, `set -euo pipefail` all present |
| 4 | `:latest` image tags on sidecars | 🟡 Partial — dsv4pro router pinned `v0.3.2`; prom `:latest` gone with the file; **qwen router still `:latest`** (R2-B) |
| 5 | `nodeSelector` only matches HyperPod (portability) | 🟡 Partial — serving manifests use `nodeAffinity`/`In` ✅; **`download-model-daemonset.yaml` still bare `nodeSelector`** (R2-C) |
| 6 | DaemonSet is the wrong primitive for download | 🟡 Partial (conceded for multi-node) — residual: single-node + `sleep infinity` re-download |
| 7 | `prometheus-agent-amp.yaml` runs in server mode | ✅ Resolved — file deleted |
| 8 | `build-image.sh` missing `set -euo pipefail` | ✅ Resolved |
| 9 | Qwen UCX "use LIBFABRIC" warning needs a README note | ✅ Resolved — note added ("intra-node PD uses UCX; the EFA-detected warning is expected/benign") |
| 10 | `imagePullPolicy: Always` on Kimi engines | 🟡 Mostly — 4× `IfNotPresent` now; **1 `Always` remains** somewhere in the manifests |
| 11 | No `livenessProbe` on serving deployments | ✅ Resolved — `httpGet /health:30000` added across manifests |
| 12 | `privileged: true` on all serving containers | ❌ Deferred — author defers ("not risky, narrow later; vllm example also uses it") |
| 13 | DCGM exporter runs as root | ✅ Resolved — file removed |
| 14 | `NCCL_SOCKET_IFNAME` unset (multi-node) | ✅ Resolved — `value: ^lo` (exclusion form) added |
| 15 | PR description CloudFormation mismatch | ✅ Resolved — description updated |
| 16 | Base branch is `worktree-repo-reorg` | ➖ N/A — confirmed intentional target |
| 17 | Placeholder benchmark tables | ➖ N/A — was a positive (honest framing) |

**Legend:** ✅ Resolved · 🟡 Partial · ❌ Open/Deferred · ➖ N/A

**Correction to a round-1 concession:**
- **#2 (images):** I withdrew this in round 1 accepting that `lmsysorg/sglang:deepseek-v4-b300` "carries dedicated DeepSeek-V4-Pro modifications (the layer link settles it)." I then tested that claim and it does **not** hold — the model arch and kernel stack are both in mainline `0.5.12.post1`, and DeepSeek-V4-Pro serves bit-identically on the unified image on B300. Reopening as a must-fix (**R2-G** below), with the evidence.

**Conceded on your evidence (no action needed):**
- **#6 (DaemonSet):** fair — a DaemonSet does fan out exactly one copy per node, and neither a Job nor an initContainer pins one-per-node natively (my own follow-up research reached the same conclusion). I'm withdrawing the "wrong primitive" framing for the multi-node case. See R2-D for the narrow residual.

---

---

### R2-G — Consolidate the three examples onto one `stock + EFA` image; drop the dedicated `deepseek-v4-b300` dependency

**Files:** `dsv4pro-b300-single-node/manifests/dsv4pro-deploy.yaml`, `qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml`, `kimi2.6-h200-1p1d/{Dockerfile,build-image.sh,manifests/kimi-pd-deploy.yaml}`, `README.md` (engine-image column)

The PR ships three different base images for three examples — stock `lmsysorg/sglang:v0.5.12.post1-cu130` (Qwen), a `build-image.sh` EFA build of that same stock (Kimi), and `lmsysorg/sglang:deepseek-v4-b300` (DeepSeek). From a maintenance standpoint we don't want to carry three divergent base images across these examples, and I verified empirically that **we don't have to** — one image covers all three.

**Evidence (tested 2026-06-18):**
- **Qwen + Kimi already share a base** — Kimi's image is literally `FROM lmsysorg/sglang:v0.5.12.post1-cu130` + the AWS EFA installer (an *additive* layer). Mainline `0.5.12.post1` carries both `qwen3_5` and `kimi_k25` archs; the `stock+EFA` image serves the single-node UCX path with no regression and the multi-node LIBFABRIC/EFA path (the latter already confirmed earlier on B300). The EFA layer is harmless on the single-node example.
- **DeepSeek-V4-Pro does *not* need the dedicated image** — A/B on one B300 node (`p6-b300.48xlarge`), the PR's exact `dsv4pro` command, same cached weights, swapping only the image:
  - **dedicated `deepseek-v4-b300`** vs **unified `stock+EFA`** → **bit-identical greedy output** (same 24 `output_ids`), both `DeepseekV4ForCausalLM`, EAGLE spec-decode accept-rate 1.0, `flashinfer_mxfp4` MoE, fp8 KV.
  - The dedicated image's apparent hard dependency — a packaged `config_backup_large.json` that overrides the checkpoint config — is **not** load-bearing: the unified image used `deepseek-ai/DeepSeek-V4-Pro`'s own complete `config.json` (`model_type: deepseek_v4`, incl. `hc_sinkhorn_iters`/`index_head_dim`) and produced identical results. The two dev-branch-only `SGLANG_OPT_*` flags (`USE_FAST_MASK_EP`, `USE_JIT_NORM`) are silently ignored by mainline — no crash.
  - So `deepseek-v4-b300` is a **dev-branch convenience snapshot** (built from `git clone -b deepseek_v4 … + pip install -e`), not a permanent fork. (This corrects my round-1 "the layer link settles it" — the layers diverge because it's a branch build, not because the model code is exclusive to it.)

**Ask:** standardize all three examples on the single `build-image.sh` `stock+EFA` image (the EFA layer is a no-op for the single-node cases, so one image is the simplest correct choice), and point `dsv4pro` at it instead of `lmsysorg/sglang:deepseek-v4-b300`. That removes a third image, an opaque non-semver external tag, and the divergent build from the repo's maintenance surface.

**Performance checked too — no penalty.** I benchmarked both images on the same B300 node (`sglang.bench_serving`, 1024-in/512-out, identical flags, EAGLE accept-length identical at ~1.99 so the comparison is fair):

| | dedicated `deepseek-v4-b300` | unified `stock+EFA` |
|---|---|---|
| Output tok/s @ concurrency 1 | 79.4 | 83.8 (tie/edge) |
| Median TPOT @ c1 | 10.76 ms | 10.66 ms |
| **Output tok/s @ concurrency 32** | **1118** | **1609 (+44%)** |
| 131,072 gen tokens in | 117 s | 81 s |

Per-request latency is a tie; at concurrency 32 the **unified image was meaningfully *faster*** (+44% output throughput). Likely because the dedicated image runs the dev-branch `compressed` attention path with the `SGLANG_OPT_USE_FAST_MASK_EP`/`USE_JIT_NORM` knobs on, while mainline auto-selects `dsv4` and ignores them. So there is no performance reason to retain the dedicated image either — consolidation is strictly a simplification. (Caveat: one model/profile/two concurrency points, not an exhaustive sweep; the c32 edge is for this workload. Full data: `IMAGE_UNIFICATION_TEST_PR1128.md`.)



---

examples/inference/sglang/README.md
```diff
@@ -0,0 +1,99 @@
...
+# e.g. ./setup-amp-monitoring.sh eks-hypd-0512-b2ad us-west-2 sglang-kimi
+# then watch the agent leave CrashLoopBackOff:
+#   kubectl rollout status deployment/prometheus-agent
+```
+
+The agent scrapes every pod labeled `sglang-metrics=true` or `dcgm-metrics=true`
+and remote-writes via SigV4. (Requires `awscli`, `eksctl`, `kubectl`, `envsubst`
+and AWS creds with AMP + IAM permissions.)
+
+**2. GPU metrics** — [`dcgm-exporter-daemonset.yaml`](./dcgm-exporter-daemonset.yaml)
```
KeitaW: ### R2-A — `sglang/README.md` still documents the deleted AMP/DCGM monitoring files

**File:** `examples/inference/sglang/README.md` (line 65)

Removing `prometheus-agent-amp.yaml`, `setup-amp-monitoring.sh`, and `dcgm-exporter-daemonset.yaml` resolved the observability finding — but the README's entire **"Monitoring (Prometheus + Grafana)"** section (≈lines 59–90) still walks through them: `./setup-amp-monitoring.sh …`, ``[`prometheus-agent-amp.yaml`](./prometheus-agent-amp.yaml)``, and `kubectl apply -f dcgm-exporter-daemonset.yaml`. The Kimi README's "Files" table also still links `../dcgm-exporter-daemonset.yaml`. These now point at files that don't exist — a reader following the monitoring steps hits "file not found." I'd either drop that section and link to the repo's [`eks-managed-observability`](https://github.com/awslabs/awsome-distributed-ai/tree/main/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability) / the HyperPod observability add-on as the prerequisite, or restore a (managed) monitoring path — but the docs shouldn't reference deleted manifests. → see [review-checklist.md Category 5 — File path accuracy]



---

KeitaW (commented):
## Review (round 2) — Batch 2/5: Documentation Consistency

One should-fix: the README still documents monitoring manifests that were deleted in this PR (inline).

examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml
```diff
@@ -0,0 +1,284 @@
...
+            --disaggregation-bootstrap-port 9005 \
+            --mem-fraction-static 0.85 \
+            --reasoning-parser qwen3 --tool-call-parser qwen3_coder \
+            --context-length 32768 \
+            --disable-flashinfer-autotune \
+            --skip-server-warmup \
+            --enable-metrics
+      # ---- PD router (shares pod network → reaches engines on 127.0.0.1) ---
+      - name: router
+        image: lmsysorg/sgl-model-gateway:latest
```
KeitaW: ### R2-B — Qwen router image is still `:latest` (the `v0.3.2` pin landed in dsv4pro but missed this one)

**File:** `examples/inference/sglang/qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml` (line 206)

The DeepSeek manifest's router is pinned to `lmsysorg/sgl-model-gateway:v0.3.2` (thank you), but the Qwen router here is still `:latest`. Looks like the pin just didn't get copied to this file. Same fix:

```suggestion
        image: lmsysorg/sgl-model-gateway:v0.3.2
```



---

KeitaW (commented):
## Review (round 2) — Batch 3/5: Structure & Repository Hygiene

One should-fix: the Qwen router image is still `:latest` (inline, with a one-click pin suggestion).

examples/inference/sglang/download-model-daemonset.yaml
```diff
@@ -0,0 +1,69 @@
...
+  selector:
+    matchLabels:
+      app: model-downloader
+  template:
+    metadata:
+      labels:
+        app: model-downloader
+    spec:
+      nodeSelector:
+        node.kubernetes.io/instance-type: ${INSTANCE_TYPE}
```
KeitaW: ### R2-C — `download-model-daemonset.yaml` still uses a bare `nodeSelector` (won't schedule on plain EKS)

**File:** `examples/inference/sglang/download-model-daemonset.yaml` (line 31)

The serving manifests got the nice `nodeAffinity`/`In` portability fix, but the download DaemonSet still pins `node.kubernetes.io/instance-type: ${INSTANCE_TYPE}` — and the README invokes it with the `ml.`-prefixed value (`./download-model.sh deepseek-ai/DeepSeek-V4-Pro ml.p6-b300.48xlarge`). On a plain-EKS nodegroup (bare label `p6-b300.48xlarge`) the DaemonSet pods won't schedule, so the pre-stage step silently does nothing there — the same HyperPod-vs-EKS mismatch you just fixed for serving. I'd either give this manifest the same `nodeAffinity`/`In` (rendering both label forms) or document passing the bare type on plain EKS.

KeitaW (commented):
## Review (round 2) — Batch 4/5: Deployment Pipeline & K8s Operational Correctness

### R2-D — DaemonSet downloader: residual (narrowed, non-blocking)

**Files:** `download-model-daemonset.yaml`, `download-model.sh`

Conceding the multi-node point from round 1: a DaemonSet *is* the natural one-copy-per-node primitive, and neither a plain Job nor an initContainer pins one-completion-per-node natively (Kubernetes has no "run-once-per-node-to-completion" object — it's been requested upstream since 2017 and never shipped). I followed up with dedicated research to make sure I wasn't hand-waving; the verdict is that this is a **real but non-blocking** smell, and worth stating precisely rather than as a vague "consider something lighter."

**Why the `sleep infinity` is there (and why it signals the wrong primitive):** a DaemonSet Pod template must use `restartPolicy: Always`, so a downloader container that exits 0 is *restart-looped* into CrashLoopBackOff. `sleep infinity` is the only way to park it — which is exactly the tell that a DaemonSet is being bent to do one-shot work it wasn't designed for. The cost is then permanent: the parked pod holds a scheduling slot (and any resource requests) on every node for the cluster's lifetime, gets no automatic node-pressure-eviction protection, and — because the DaemonSet is decoupled from the serving workload — **re-downloads on every node replacement / scale-out** (the weights sit on instance-store NVMe, which is wiped on stop/terminate/retirement, so a new node always starts empty). (Minor correction to my round-1 wording: `kubectl delete` of the DaemonSet *does* GC its pods — the defect is the missing tie to the workload, not failed cleanup.)

**The lighter options, in order:**
1. **Match the repo's own house style.** The sibling multi-node example [`dsv3-uccl-nixl`](https://github.com/awslabs/awsome-distributed-ai/tree/main/3.test_cases/pytorch/vllm/dsv3-uccl-nixl) serves a 600+ GB model with **download-on-startup into a hostPath NVMe cache** (`HF_HOME` on `/mnt/k8s-disks/0`) + `imagePullPolicy: IfNotPresent`, one-pod-per-node enforced by a **pinned Deployment** (`nodeAffinity` on hostname + `podAntiAffinity`) — no DaemonSet, no initContainer. That's the established in-tree idiom this PR diverges from.
2. **For the single-node examples (dsv4pro, qwen)** a per-node DaemonSet is plainly more than needed — an `initContainer`-with-skip-if-present (init containers *do* run to completion, unlike DaemonSet containers) gets the same staging and ties lifecycle to the workload.
3. **For production shared storage**, FSx for Lustre linked to S3 (or EFS) mounted RWX stages weights *once* and eliminates per-node copies entirely — which is what AWS's EKS AI/ML best-practices, NVIDIA Dynamo ("shared PVC + one-time Job"), and KServe all converge on. None of them endorse an unmanaged `sleep infinity` hostPath DaemonSet; the one sanctioned node-local-cache mechanism (KServe `LocalModelCache`) splits it into a run-to-completion **Job** + a thin monitoring DaemonSet, never one long-lived pod doing both.

Still **non-blocking for an example** — but the answer to "will the DaemonSet run forever even after the download?" is *yes, by construction*, and the repo's own sibling example shows the lighter pattern. Recommend aligning with `dsv3-uccl-nixl` (download-on-startup) rather than keeping the standalone DaemonSet.



---

KeitaW (commented):
## Review (round 2) — Batch 5/5: Infrastructure, Security, Positives & Sources

### R2-E — `privileged: true` still present (deferred)

**Files:** `dsv4pro-deploy.yaml`, `qwen-pd-deploy.yaml`, Kimi manifests

Acknowledged that you've deferred this and that the sibling vllm example also runs privileged — so it's consistent with existing repo practice, and I won't push. Leaving it as a low-priority follow-up: when convenient, scoping to the specific capabilities (`IPC_LOCK`, `SYS_NICE`) + the device mounts already declared would tighten it, but it's not blocking for an example.

---

### R2-F — Add an explicit "stock image will crash" guard to the Kimi README (nit; the config itself is correct)

**File:** `kimi2.6-h200-1p1d/README.md`

To be clear up front: **the PR already does the right thing here** — I'm not asking for a config change. The Kimi multi-node manifest references `<YOUR_ECR_IMAGE>` (the `build-image.sh` output, i.e. the Kimi `Dockerfile` = base + `aws-efa-installer 1.47.0`) and sets `SGLANG_DISAGGREGATION_NIXL_BACKEND=LIBFABRIC`, and the README's Quick Start walks through building it. I verified that path works end-to-end on a 2-node B300 setup: NIXL 1.1.0 instantiates the LIBFABRIC backend, `fi_info -p efa` shows `efa-direct / FI_PROTO_EFA`, GPU memory is registered on EFA rails, and the prefill→decode KV transfer rides EFA (~9 GiB of RDMA traffic measured at the decode node for 30 completions vs 0 idle). The Qwen example correctly uses UCX intra-node on the stock image (no crash), and DeepSeek is non-PD — so no manifest in this PR pairs a stock image with `LIBFABRIC`.

The only residual is a documentation guard. The built image is *mandatory* for this example, but that reads as a normal build step — and the failure mode if someone substitutes the (tempting, "already pinned") stock `lmsysorg/sglang:v0.5.12.post1-cu130` tag is a cryptic hard crash-loop, not a clear error. I'd add one callout near the build/image step:

> ⚠️ **The built image is required.** This example's KV transfer uses NIXL's LIBFABRIC backend, which needs libfabric's `efa` provider — present only in the image `build-image.sh` builds (base + AWS EFA installer). Do **not** substitute the stock `lmsysorg/sglang:v0.5.12.post1-cu130` tag: it lacks the efa provider and the engine hard-crashes (CrashLoopBackOff) with `nixlNotFoundError: NIXL_ERR_NOT_FOUND` at `create_backend("LIBFABRIC")`.

That converts a confusing CrashLoopBackOff into a one-line prereq. Purely a doc nicety — the shipped configuration is correct.



---

---

## Things That Look Great (improvements since round 1)

- **The whole bespoke observability stack is gone** — removing `prometheus-agent-amp.yaml` / `setup-amp-monitoring.sh` / `dcgm-exporter-daemonset.yaml` retired four round-1 findings (the server-mode bug, the prom `:latest`, the DCGM-root context, and the reuse concern) in one move. Cleanest possible resolution.
- **Portability done right** — the `nodeAffinity` + `In`-over-values block in the serving manifests, with a comment explaining the HyperPod-vs-EKS instance-type label, is exactly the fix; and the new `manifests/nodegroup-b300-eks.yaml` + `NODEGROUP-EKS.md` make the B300 example actually runnable on a plain-EKS capacity-block nodegroup.
- **Liveness + NCCL** — `httpGet /health:30000` liveness probes and the `^lo` exclusion-form `NCCL_SOCKET_IFNAME` both landed correctly.
- **`build-image.sh` hardened** — shebang, MIT-0 header, and `set -euo pipefail` all in.
- **Taint tolerations are correctly key-scoped** (`key: nvidia.com/gpu`), not the dangerous blanket `{operator: Exists}` — nicely done.
- **Docs engaged with the feedback** — the engine-image column explaining the per-model image choice, the Qwen UCX/EFA-warning note, and the CloudFormation description fix all directly answer round-1 comments.

---

## Sources

- File-path accuracy / fixed-tag / security rules — repo `references/review-checklist.md`, `recurring-findings.md`.
- Repo observability precedent — [eks-managed-observability](https://github.com/awslabs/awsome-distributed-ai/tree/main/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability); weight-staging sibling — [dsv3-uccl-nixl](https://github.com/awslabs/awsome-distributed-ai/tree/main/3.test_cases/pytorch/vllm/dsv3-uccl-nixl).
- Weight-staging pattern verdict — `/lt:research` report `daemonset-vs-initcontainer-model-weight-staging-eks` (K8s DaemonSet restartPolicy=Always; no native run-once-per-node primitive; AWS EKS AI/ML storage best-practices; KServe LocalModelCache; EC2 instance-store ephemerality).
- Verified against the live diff at head `b1172bfc` (2026-06-17): `nodeAffinity/In` in dsv4pro-deploy.yaml, `livenessProbe httpGet /health:30000`, `NCCL_SOCKET_IFNAME=^lo`, `build-image.sh` header+`set -euo pipefail`, taint tolerations key-scoped (`key: nvidia.com/gpu`, not blanket `operator: Exists`).

---

ybalbert001:
> ## Review (round 2) — Batch 1/5: Image Consolidation (must-fix)
> _Round-2 re-review of head `b1172bfc`. Thanks for the thorough round-1 pass — most of it is resolved (see table). One finding I had conceded in round 1 is reopened as a **must-fix** after empirical testing on B300; details below._
> 
> ## Summary
> Thanks for the thorough pass on round 1 — most of it is addressed. The bespoke AMP/DCGM observability stack is gone (retiring four findings at once), the serving manifests now schedule on both plain EKS and HyperPod via `nodeAffinity` (with a genuinely clear explanatory comment), liveness probes and the `^lo` NCCL exclusion are in, `build-image.sh` is hardened, and the docs got the engine-image column. I concede the DaemonSet point — a DaemonSet really is the natural one-copy-per-node primitive for the multi-node case (R2-D).
> 
> One **must-fix** is back, however: image consolidation (**R2-G**). In round 1 I withdrew it on your rationale that DeepSeek-V4-Pro _needs_ its dedicated `deepseek-v4-b300` image — I then verified that empirically and it does **not** hold. All three models serve on a single `stock + EFA` image; DeepSeek-V4-Pro runs **bit-identically** on the unified image vs the dedicated one on real B300 hardware (test details below). Carrying three divergent base images for three examples is a maintenance burden we shouldn't take on when one image demonstrably covers all of them. The rest is minor: three cleanup slips (stale README section, one un-pinned router tag, one download-path nodeSelector) plus `privileged` deferred.
> 
> ## Resolution Status of Prior Findings
> #	Finding (round 1)	Status
> 1	Observability duplicates an existing repo asset	✅ Resolved — AMP/DCGM manifests deleted (but see R2-A: README still references them)
> 2	Use one shared SGLang image across all models	❌ **Reopened (must-fix)** — round-1 concession empirically refuted: DeepSeek-V4-Pro serves bit-identically on the unified image on B300; one image covers all three (R2-G)
> 3	`build-image.sh` missing shebang + MIT-0 header	✅ Resolved — shebang, MIT-0, `set -euo pipefail` all present
> 4	`:latest` image tags on sidecars	🟡 Partial — dsv4pro router pinned `v0.3.2`; prom `:latest` gone with the file; **qwen router still `:latest`** (R2-B)
> 5	`nodeSelector` only matches HyperPod (portability)	🟡 Partial — serving manifests use `nodeAffinity`/`In` ✅; **`download-model-daemonset.yaml` still bare `nodeSelector`** (R2-C)
> 6	DaemonSet is the wrong primitive for download	🟡 Partial (conceded for multi-node) — residual: single-node + `sleep infinity` re-download
> 7	`prometheus-agent-amp.yaml` runs in server mode	✅ Resolved — file deleted
> 8	`build-image.sh` missing `set -euo pipefail`	✅ Resolved
> 9	Qwen UCX "use LIBFABRIC" warning needs a README note	✅ Resolved — note added ("intra-node PD uses UCX; the EFA-detected warning is expected/benign")
> 10	`imagePullPolicy: Always` on Kimi engines	🟡 Mostly — 4× `IfNotPresent` now; **1 `Always` remains** somewhere in the manifests
> 11	No `livenessProbe` on serving deployments	✅ Resolved — `httpGet /health:30000` added across manifests
> 12	`privileged: true` on all serving containers	❌ Deferred — author defers ("not risky, narrow later; vllm example also uses it")
> 13	DCGM exporter runs as root	✅ Resolved — file removed
> 14	`NCCL_SOCKET_IFNAME` unset (multi-node)	✅ Resolved — `value: ^lo` (exclusion form) added
> 15	PR description CloudFormation mismatch	✅ Resolved — description updated
> 16	Base branch is `worktree-repo-reorg`	➖ N/A — confirmed intentional target
> 17	Placeholder benchmark tables	➖ N/A — was a positive (honest framing)
> **Legend:** ✅ Resolved · 🟡 Partial · ❌ Open/Deferred · ➖ N/A
> 
> **Correction to a round-1 concession:**
> 
> * **#2 (images):** I withdrew this in round 1 accepting that `lmsysorg/sglang:deepseek-v4-b300` "carries dedicated DeepSeek-V4-Pro modifications (the layer link settles it)." I then tested that claim and it does **not** hold — the model arch and kernel stack are both in mainline `0.5.12.post1`, and DeepSeek-V4-Pro serves bit-identically on the unified image on B300. Reopening as a must-fix (**R2-G** below), with the evidence.
> 
> **Conceded on your evidence (no action needed):**
> 
> * **#6 (DaemonSet):** fair — a DaemonSet does fan out exactly one copy per node, and neither a Job nor an initContainer pins one-per-node natively (my own follow-up research reached the same conclusion). I'm withdrawing the "wrong primitive" framing for the multi-node case. See R2-D for the narrow residual.
> 
> ### R2-G — Consolidate the three examples onto one `stock + EFA` image; drop the dedicated `deepseek-v4-b300` dependency
> **Files:** `dsv4pro-b300-single-node/manifests/dsv4pro-deploy.yaml`, `qwen3.5-27b-b300-intra-pd/qwen-pd-deploy.yaml`, `kimi2.6-h200-1p1d/{Dockerfile,build-image.sh,manifests/kimi-pd-deploy.yaml}`, `README.md` (engine-image column)
> 
> The PR ships three different base images for three examples — stock `lmsysorg/sglang:v0.5.12.post1-cu130` (Qwen), a `build-image.sh` EFA build of that same stock (Kimi), and `lmsysorg/sglang:deepseek-v4-b300` (DeepSeek). From a maintenance standpoint we don't want to carry three divergent base images across these examples, and I verified empirically that **we don't have to** — one image covers all three.
> 
> **Evidence (tested 2026-06-18):**
> 
> * **Qwen + Kimi already share a base** — Kimi's image is literally `FROM lmsysorg/sglang:v0.5.12.post1-cu130` + the AWS EFA installer (an _additive_ layer). Mainline `0.5.12.post1` carries both `qwen3_5` and `kimi_k25` archs; the `stock+EFA` image serves the single-node UCX path with no regression and the multi-node LIBFABRIC/EFA path (the latter already confirmed earlier on B300). The EFA layer is harmless on the single-node example.
> * **DeepSeek-V4-Pro does _not_ need the dedicated image** — A/B on one B300 node (`p6-b300.48xlarge`), the PR's exact `dsv4pro` command, same cached weights, swapping only the image:
>   
>   * **dedicated `deepseek-v4-b300`** vs **unified `stock+EFA`** → **bit-identical greedy output** (same 24 `output_ids`), both `DeepseekV4ForCausalLM`, EAGLE spec-decode accept-rate 1.0, `flashinfer_mxfp4` MoE, fp8 KV.
>   * The dedicated image's apparent hard dependency — a packaged `config_backup_large.json` that overrides the checkpoint config — is **not** load-bearing: the unified image used `deepseek-ai/DeepSeek-V4-Pro`'s own complete `config.json` (`model_type: deepseek_v4`, incl. `hc_sinkhorn_iters`/`index_head_dim`) and produced identical results. The two dev-branch-only `SGLANG_OPT_*` flags (`USE_FAST_MASK_EP`, `USE_JIT_NORM`) are silently ignored by mainline — no crash.
>   * So `deepseek-v4-b300` is a **dev-branch convenience snapshot** (built from `git clone -b deepseek_v4 … + pip install -e`), not a permanent fork. (This corrects my round-1 "the layer link settles it" — the layers diverge because it's a branch build, not because the model code is exclusive to it.)
> 
> **Ask:** standardize all three examples on the single `build-image.sh` `stock+EFA` image (the EFA layer is a no-op for the single-node cases, so one image is the simplest correct choice), and point `dsv4pro` at it instead of `lmsysorg/sglang:deepseek-v4-b300`. That removes a third image, an opaque non-semver external tag, and the divergent build from the repo's maintenance surface.
> 
> **Performance checked too — no penalty.** I benchmarked both images on the same B300 node (`sglang.bench_serving`, 1024-in/512-out, identical flags, EAGLE accept-length identical at ~1.99 so the comparison is fair):
> 
> dedicated `deepseek-v4-b300`	unified `stock+EFA`
> Output tok/s @ concurrency 1	79.4	83.8 (tie/edge)
> Median TPOT @ c1	10.76 ms	10.66 ms
> **Output tok/s @ concurrency 32**	**1118**	**1609 (+44%)**
> 131,072 gen tokens in	117 s	81 s
> Per-request latency is a tie; at concurrency 32 the **unified image was meaningfully _faster_** (+44% output throughput). Likely because the dedicated image runs the dev-branch `compressed` attention path with the `SGLANG_OPT_USE_FAST_MASK_EP`/`USE_JIT_NORM` knobs on, while mainline auto-selects `dsv4` and ignores them. So there is no performance reason to retain the dedicated image either — consolidation is strictly a simplification. (Caveat: one model/profile/two concurrency points, not an exhaustive sweep; the c32 edge is for this workload. Full data: `IMAGE_UNIFICATION_TEST_PR1128.md`.)

updated

KeitaW (approved):
LGTM!

PR #1129, opened 2026-06-11
 Add SLIME (SGLang + Megatron + Ray) GRPO test case on HyperPod EKS

allela-roy:
## Purpose

Add a new test case demonstrating reinforcement learning (RL) post-training with
[SLIME](https://github.com/THUDM/slime) (SGLang + Megatron-LM + Ray) on Amazon
SageMaker HyperPod with EKS orchestration. The sample runs GRPO training for
Qwen3-class models and covers two topologies:

- **Colocated** (baseline) — training (Megatron-LM) and rollout (SGLang) share
  the same GPUs, with an in-process built-in reward. Simplest setup.
- **Disaggregated** (advanced) — a cost-optimized topology that offloads reward
  scoring to a separate EC2 **Spot** CPU instance group, freeing the GPUs and
  letting the reward tier scale independently.

## Changes

- **New test case** `3.test_cases/pytorch/slime/` for SLIME-based GRPO RL training on HyperPod EKS.

-

## Test Plan

**Environment:**
- AWS Service: SageMaker HyperPod (EKS orchestration)
- Instance type: 2× `ml.p5.48xlarge` (GPU, On-Demand), 2× `ml.m5.4xlarge` (general), 4× `ml.c5.4xlarge` (reward pool, **EC2 Spot**)
- Number of nodes: 8 (16× H100 GPUs total)

**Test commands:**
```bash
# 0. KubeRay operator (one-time)
helm install kuberay-operator kuberay/kuberay-operator --version 1.4.2 \
  --namespace kuberay-operator --create-namespace

# 1. Build + push images
docker build -t ${REGISTRY}slime-hyperpod:${TAG} -f slime.Dockerfile .
docker build -t ${REGISTRY}slime-reward:cpu      -f reward_service.Dockerfile .
docker push ${REGISTRY}slime-hyperpod:${TAG} && docker push ${REGISTRY}slime-reward:cpu

# 2. (optional) provision the Spot CPU reward pool
CLUSTER_NAME=<cluster> GROUP_NAME=reward-spot-c5 \
  INSTANCE_TYPE=ml.c5.4xlarge INSTANCE_COUNT=4 CAPACITY_TYPE=spot \
  infra/add_cpu_instance_group.sh

# 3. Deploy reward service (Spot pool) + Ray cluster (GPU pool)
cp env_vars.disaggregated.example env_vars.disaggregated
source env_vars && source env_vars.disaggregated
envsubst < kubernetes/reward-service.yaml | kubectl apply -f -
envsubst < kubernetes/raycluster.yaml     | kubectl apply -f -

# 4. End-to-end reward path from a GPU pod
python3 -c "import asyncio; from types import SimpleNamespace; \
from slime.utils.types import Sample; from slime.rollout.rm_hub import batched_async_rm; \
args=SimpleNamespace(custom_rm_path=None, rm_type='remote_rm', \
rm_url='http://slime-reward.default.svc.cluster.local:8000/score'); \
print(asyncio.run(batched_async_rm(args, [Sample(prompt='Explain gravity.', \
response='Gravity is the curvature of spacetime caused by mass-energy.', label=None)])))"
```

## Test Results

- **Reward service (Spot pool)**: 4/4 replicas Running, one per c5 Spot node; reward model discriminates correctly (good response **+6.49**, bad response **−2.90**).
- **Spot resilience**: simulated pod eviction kept the Service serving with zero downtime (other replicas) and the deployment self-healed to 4/4; across sessions, HyperPod auto-replaced reclaimed Spot nodes back to 4/4.
- **End-to-end `remote_rm`**: SLIME's `batched_async_rm` from a GPU worker → Spot reward pool returned `[7.03, -3.17, 0.38]` (good > bad), out-of-the-box (no manual installs).
## Directory Structure

```
3.test_cases/
```

## Checklist


paragao:
## Review Findings — Addressed

All findings from the code review have been addressed in commit 0b55936. Summary:

| # | Severity | File | Finding | Fix |
|---|----------|------|---------|-----|
| 1 | CRITICAL | `reward_service/app.py` | Dead/unreachable duplicate try/except block (lines 133–141) | Removed dead code |
| 2 | CRITICAL | `env_vars.disaggregated.example` | Batch-size constraint violated (64×16≠512×1) | Added `NUM_STEPS_PER_ROLLOUT=2` with constraint formula comment |
| 3 | HIGH | `kubernetes/raycluster.yaml` | `rayVersion: "2.44.0"` mismatches Docker image (2.55.1) | Updated to `"2.55.1"` |
| 4 | HIGH | `raycluster.yaml` + `reward-service.yaml` | Plain-text `HF_TOKEN` injection (security) | Replaced with `secretKeyRef` to K8s Secret; added creation instructions |
| 5 | MEDIUM | `slime.Dockerfile` | Permissive SSH config without rationale | Added comment block explaining HPC multi-node use case and cluster-boundary scope |
| 6 | MEDIUM | `kubernetes/data-prep-pod.yaml` | Hardcoded `namespace: default` | Changed to `${NAMESPACE}` |
| 7 | MEDIUM | `recipe/run_grpo_qwen3_30b_a3b.sh` | Missing `--rm-url` handling for `remote_rm` | Added RM_ARGS conditional block (matches 4B recipe) |
| 8 | LOW | `env_vars.colocated.example` | Missing copyright header | Added standard Apache/MIT-0 header |
| 9 | LOW | `requirements.txt` | `aiohttp` used in evaluate.sh but not declared | Added as explicit dependency |
| 10 | LOW | `scripts/evaluate.sh` | Hardcoded 120s startup timeout too short for large models | Made configurable via `SGLANG_STARTUP_TIMEOUT` (default 300s) |
| 11 | LOW | `README.md` | Incorrect `cd slime_hyperpod` path in Quick Start | Fixed to `cd 3.test_cases/pytorch/slime` |

3.test_cases/pytorch/slime/.gitignore
```diff
@@ -0,0 +1,17 @@
...
+*.egg-info/
+.ipynb_checkpoints/
+
+# Eval / training artifacts
+eval_results/
+*.log
+
+# OS / editor
+.DS_Store
+*.swp
```
KeitaW: ### `.gitignore` duplicates the repo-root ignores and is missing the MIT-0 header

Most entries repeat the repo-root `.gitignore` (`*.log`, `__pycache__/`, `*.py[cod]`, `*.egg-info/`, `.DS_Store`, `*.swp` via root's `*.sw[po]`). The only test-case-specific entries are `env_vars`, `env_vars.disaggregated`, and `eval_results/`. Suggest trimming to those plus the MIT-0 header (committed files need it even when every line is a comment):

```suggestion
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT-0

# Local environment files with filled-in secrets/values -- never commit
env_vars
env_vars.disaggregated

# Eval / training artifacts
eval_results/
```

KeitaW (commented):
## Review Batch 1/6 — Structure & Repository Hygiene


### Fold the CPU reward pool into Prerequisites instead of bundling cluster-mutation scripts in `infra/`

**Files:** `infra/add_cpu_instance_group.sh`, `infra/remove_cpu_instance_group.sh`, `infra/README.md`

The `infra/` helpers call `aws sagemaker update-cluster` to add/remove a HyperPod instance group — cluster-level infrastructure provisioning, which the repo convention keeps out of test cases and documents as a prerequisite instead. The sibling RL test cases already do this: `pytorch/openrlhf/README.md` lists *"An EKS cluster with HyperPod managed node groups"* as **prerequisite #1**, and neither `openrlhf`, `verl`, nor `trl` ships instance-group scripts.

Two reasons beyond convention:

1. **The script mutates cluster topology against a narrow model of `UpdateCluster`.** The header asserts *"UpdateCluster replaces the FULL InstanceGroups list"* and resubmits every existing group as a union to avoid deleting them. Omitting a group does **not** delete it — deletion uses the dedicated `InstanceGroupsToDelete` parameter ([Scaling down a HyperPod cluster](https://docs.aws.amazon.com/sagemaker/latest/dg/smcluster-scale-down.html)) — so the full-union resubmit is unnecessary. It's also risky: `keep()` rebuilds each existing group from an allowlist (`InstanceGroupName`, `InstanceType`, `InstanceCount`, `LifeCycleConfig`, `ExecutionRole`, `ThreadsPerCore`, optionally `InstanceStorageConfigs`/`OverrideVpcConfig`/`CapacityRequirements`), but [`ClusterInstanceGroupSpecification`](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_ClusterInstanceGroupSpecification.html) also carries `OnStartDeepHealthChecks`, `KubernetesConfig` (labels/taints), `ScheduledUpdateConfig`, `TrainingPlanArn`, `ImageId`, `MinInstanceCount`, `InstanceRequirements`. The author's own code proves omitted fields aren't preserved (`CapacityRequirements` is explicitly carried "immutable; must be resubmitted"). So resubmitting an existing group **without** those fields risks resetting them — e.g. disabling deep health checks or dropping K8s labels/taints on the **GPU** pool, which the union includes. AWS doesn't document exact per-field merge semantics, so treat this as a documented-shaped risk, not a certainty — but it's exactly the kind of cluster-management subtlety that belongs in the provisioning layer, maintained against the current API, not copy-pasted into a test case.
2. The README already half-commits: *"This sample assumes the cluster already has a Spot-backed `reward-spot-c5` instance group"* — so this is finishing the move, not inventing it.

**Suggested direction:** add a Prerequisites line for the disaggregated path (e.g. *"A CPU instance group in the same AZ as the GPU pool to host the reward service; see [HyperPod instance-group management](https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-hyperpod-eks-operate-cli-command-update-cluster.html)"*) and drop the bundled scripts. (`architectures/` here only contains `aws-pcs`, a different orchestrator, so point at the HyperPod docs rather than an in-repo path.)

KeitaW (commented):
## Review Batch 2/6 — Infrastructure, NCCL & Container Security

One nit here on the worker `NCCL_SOCKET_IFNAME`. See the inline comment.

3.test_cases/pytorch/slime/kubernetes/raycluster.yaml
```diff
@@ -0,0 +1,167 @@
...
+                - name: FI_EFA_FORK_SAFE
+                  value: "1"
+                - name: NCCL_PROTO
+                  value: "Simple"
+                - name: NCCL_DEBUG
+                  value: "WARN"
+                - name: RDMAV_FORK_SAFE
+                  value: "1"
+                - name: NCCL_SOCKET_IFNAME
+                  value: "eth0"
```
KeitaW: ### `NCCL_SOCKET_IFNAME: "eth0"` — align with repo convention (nit; not a throughput killer)

The worker sets `NCCL_SOCKET_IFNAME` with positive selection (`eth0`); `slime.Dockerfile` sets the exclusion form (`^docker,lo,veth_def_agent`) and the manifest env wins. To be precise about impact: `NCCL_SOCKET_IFNAME` governs only NCCL's TCP **bootstrap/control** socket, not the EFA data path (collectives go over libfabric via `FI_PROVIDER=efa` + aws-ofi-nccl). In a pod `eth0` *is* the primary CNI interface, so bootstrap works and the 32 EFA devices still carry the collectives — p5.48xlarge throughput is very likely **unaffected**.

Still worth fixing: (1) repo convention is exclusion-based patterns (`^lo` for K8s), which is robust if the primary NIC is ever not named `eth0`; (2) it removes the silent Dockerfile-vs-manifest disagreement. Nit, not a blocker.

```suggestion
                  value: "^lo"
```

3.test_cases/pytorch/slime/kubernetes/raycluster.yaml
```diff
@@ -0,0 +1,167 @@
...
+      # Head runs the Ray GCS and dashboard only -- no training/inference on head.
+    template:
+      metadata:
+        labels:
+          ray.io/node-type: head
+      spec:
+        containers:
+          - name: ray-head
+            image: ${FULL_IMAGE}
+            imagePullPolicy: Always
```
KeitaW: ### `imagePullPolicy: Always` on pinned images

`Always` forces a registry pull on every pod start (slow cold starts, breaks air-gapped clusters). `${FULL_IMAGE}` is tag-pinned, so `IfNotPresent` is right here. Same `Always` on the worker (L103) and the reward Deployment (`reward-service.yaml` L72).

```suggestion
            imagePullPolicy: IfNotPresent
```

_Nuance for the reward service:_ its tag is the **mutable** `:cpu`, so `IfNotPresent` there alone could serve a stale build — version that tag too (e.g. `slime-reward:cpu-0.2.4`) and then use `IfNotPresent`.

3.test_cases/pytorch/slime/kubernetes/reward-service.yaml
```diff
@@ -0,0 +1,123 @@
...
+            limits:
+              cpu: "8"
+              memory: "16Gi"
+          readinessProbe:
+            httpGet:
+              path: /health
+              port: 8000
+            initialDelaySeconds: 20
+            periodSeconds: 10
+          livenessProbe:
```
KeitaW, line 114: ### Reward `livenessProbe` can crashloop on a cold Spot start

The backend downloads + loads `deberta-v3-large` on CPU at startup (`@app.on_event("startup")`), but `livenessProbe.initialDelaySeconds: 60` (L106). On a freshly reclaimed Spot c5 with a cold HF cache, download + load can exceed 60s → liveness fires before ready → kubelet restarts → crashloop precisely during the Spot churn this design is meant to ride out. Readiness gates traffic correctly; liveness shouldn't race a cold load. Add a `startupProbe` and let liveness take over after it passes:

```yaml
          startupProbe:
            httpGet:
              path: /health
              port: 8000
            periodSeconds: 10
            failureThreshold: 30   # up to ~5 min for a cold model download/load
```

Pre-baking the reward model into the image would remove the cold-download window entirely.

3.test_cases/pytorch/slime/env_vars.disaggregated.example
```diff
@@ -0,0 +1,53 @@
...
+export RM_TYPE="remote_rm"
+export REWARD_BACKEND="reward_model"
+export REWARD_MODEL_NAME="OpenAssistant/reward-model-deberta-v3-large-v2"
+# Name of the CPU instance group hosting the reward pods. This sample assumes it
+# is a Spot-capacity group (CapacityRequirements: Spot) -- the reward service is
+# stateless and SLIME's remote_rm client retries on transient failures, so Spot
+# interruptions are tolerated. Provision it with infra/add_cpu_instance_group.sh.
+export REWARD_NODE_GROUP="reward-spot-c5"
+export REWARD_REPLICAS=4               # one per c5.4xlarge node
+export REWARD_TORCH_THREADS=6          # c5.4xlarge ~7.9 allocatable vCPU on EKS
```
KeitaW: ### Reward pod under-provisions the c5 node, and the sizing comment is wrong

This comment says `~7.9 allocatable vCPU` and sets `REWARD_TORCH_THREADS=6`; the reward pod requests `cpu: "6"` (limit `"8"`), one replica/node. But `ml.c5.4xlarge` is **16 vCPU** — `~7.9` is a `c5.2xlarge`. On the tier the README measures as the bottleneck (~223s to score 1024 vs ~18s to generate), each node then runs reward inference on ~6–8 of its 16 vCPUs, leaving ~half idle — working against the "scale the busy reward tier independently" rationale. Correct the comment to 16 vCPU and raise `TORCH_NUM_THREADS` / the pod `cpu` request toward allocatable (e.g. threads ~12–14, request ~12), or note why it's capped.

KeitaW (commented):
## Review Batch 3/6 — Deployment Pipeline & K8s Operational Correctness

Three operational items on the Ray cluster and reward service — see inline comments (`imagePullPolicy`, reward `livenessProbe` cold-start race, reward pod sizing).

3.test_cases/pytorch/slime/reward_service/app.py
```diff
@@ -0,0 +1,162 @@
...
+    _scorer = _build_scorer()
+
+
+@app.get("/health")
+def health():
+    return {"status": "ok", "backend": REWARD_BACKEND}
+
+
+@app.post("/score")
+def score(req: ScoreRequest):
```
KeitaW, line 159: ### Default reward backend has no exception guard — any scoring error becomes an HTTP 500

`MathVerifyScorer.score` wraps its logic in `try/except` and returns `0.0` on failure, but the **default** backend `RewardModelScorer.score` (tokenize → forward → tensor index) and this `/score` handler have no guard. Since `REWARD_BACKEND` defaults to `reward_model`, the production path is the unprotected one: an odd truncation, an unexpected `logits` shape, or a CPU OOM on the small c5 surfaces as a 500.

SLIME's client source shows why that's costly: `remote_rm` POSTs to `rm_url`, calls `resp.raise_for_status()`, returns `await resp.json()` as the reward, retries `max_retries=10` with exponential backoff (`min(2**attempt, 30) + random()`) and then **re-raises** ([`rm_hub/__init__.py` L34–L52 @ v0.2.4](https://github.com/THUDM/slime/blob/v0.2.4/slime/rollout/rm_hub/__init__.py#L34-L52)); `batched_async_rm` gathers with a bare `asyncio.gather(*tasks)` (no `return_exceptions`, L94–L105), so that raise **fails the whole rollout batch**. One reproducibly-bad sample burns ~10 backed-off round-trips and can then take down the step. The rule-based fallback is hardened; the default isn't. Wrap the handler, log, and return a neutral float:

```python
@app.post("/score")
def score(req: ScoreRequest):
    try:
        return _scorer.score(req.prompt, req.response, req.label)
    except Exception as e:  # never 500 the rollout loop over one sample
        logger.warning("scoring error: %s", e)
        return 0.0
```

KeitaW (commented):
## Review Batch 4/6 — Reward Service Robustness

One should-fix on the reward server's error handling — see inline comment on `app.py`.

3.test_cases/pytorch/slime/env_vars.disaggregated.example
```diff
@@ -0,0 +1,53 @@
...
+#     provisioned on the HyperPod cluster (same AZ as the GPU pool, no EFA,
+#     CapacityRequirements: Spot). Provision it with infra/add_cpu_instance_group.sh
+#     and set the name below to match.
+
+# ----- Heavier GRPO config (GPU tier) -----
+export ROLLOUT_BATCH_SIZE=64          # was 16  -> 4x more prompts per rollout
+export N_SAMPLES_PER_PROMPT=16        # was 8   -> 2x more responses per prompt
+# GRPO batch-size constraint: ROLLOUT_BATCH_SIZE × N_SAMPLES_PER_PROMPT = GLOBAL_BATCH_SIZE × NUM_STEPS_PER_ROLLOUT
+export GLOBAL_BATCH_SIZE=512          # was 128 -> 4x bigger optimizer step
+export NUM_STEPS_PER_ROLLOUT=2        # adjusted to satisfy constraint: 64*16 = 512*2
```
KeitaW, line 35: ### GRPO batch-size constraint is violated at runtime — `NUM_STEPS_PER_ROLLOUT` isn't wired into the recipes

This sets `NUM_STEPS_PER_ROLLOUT=2` to satisfy `ROLLOUT_BATCH_SIZE × N_SAMPLES_PER_PROMPT = GLOBAL_BATCH_SIZE × NUM_STEPS_PER_ROLLOUT` (64×16 = 512×2 = 1024). But neither recipe reads it — both `recipe/run_grpo_qwen3_4b.sh` and `recipe/run_grpo_qwen3_30b_a3b.sh` hardcode `--num-steps-per-rollout 1`. So at runtime the profile evaluates to 64×16 = 1024 ≠ 512×1 = 512, violating the constraint the README says must "always hold"; the env var is currently dead. Plumb it through both recipes:

```bash
    --num-steps-per-rollout ${NUM_STEPS_PER_ROLLOUT:-1} \
```

KeitaW (commented):
## Review Batch 5/6 — Training Configuration Correctness

The GRPO batch-size invariant is violated at runtime in the disaggregated profile — see inline comment.

3.test_cases/pytorch/slime/README.md
```diff
@@ -0,0 +1,834 @@
...
+
+```bash
+# Authenticate to ECR
+aws ecr get-login-password --region ${AWS_REGION} | \
+    docker login --username AWS --password-stdin ${REGISTRY}
+
+# Create repository (first time only)
+aws ecr create-repository --repository-name ${IMAGE} --region ${AWS_REGION} || true
+
+# Build image (build context is the repo root)
```
KeitaW: ### "build context is the repo root" comment is incorrect

Both Dockerfiles `COPY` paths relative to the **test-case directory** — `slime.Dockerfile` does `COPY requirements.txt …` and `reward_service.Dockerfile` (L442) does `COPY reward_service/app.py …` — so the context is `3.test_cases/pytorch/slime/`, not the repo root. With step 1's `cd 3.test_cases/pytorch/slime`, the `.` now resolves correctly and the build works, but this comment still says the wrong thing and will mislead anyone reasoning about the context. Suggest updating it to say the build context is the test-case directory.

KeitaW (commented):
## Review Batch 6/6 — Documentation Consistency

One doc-accuracy note inline on the build-context comment.

---

### Things that look great

- **The disaggregated-reward write-up is genuinely excellent** — the two-offload-boundary table (Train↔Rollout on EFA vs. Rollout→Reward on HTTP), the "why no EFA / why Spot" call-outs, and the measured ~12× gen-vs-score asymmetry make the architecture crisp and evidence-backed.
- **Container version pinning is thorough** — EFA 1.48.0, NCCL v2.30.4-1, CUDA 13, aws-ofi-nccl v1.19.0, GDRCopy v2.5.2 all at/above the CI floors; SLIME/SGLang/Megatron pinned to tag or SHA. The comment explaining *why* the NCCL/OFI ARGs are declared is a nice touch.
- **Baking the Megatron `helpers.cpp` build into the image** to avoid the multi-node rank-0 lazy-build race is a sharp, well-commented fix.
- **The reward Deployment is thoughtfully Spot-aware** — rolling update with `maxUnavailable: 1`, topology spread, a scoped `NoExecute` toleration, and readiness/liveness probes on a real `/health` path (just needs the `startupProbe` tweak above).
- **The head pod is correctly kept off GPU nodes** via the portable `nvidia.com/gpu.present NotIn ["true"]` affinity, with the `NVIDIA_VISIBLE_DEVICES=void` rationale documented.
- **SLIME is a genuinely new framework** here (Ray + SGLang + Megatron), distinct from `openrlhf`/`verl`/`trl`, so a new test-case directory is the right call.

---

**Summary:** Substantial, well-reasoned contribution. The two items worth resolving before merge are the `infra/` cluster-mutation scripts (Batch 1 — better as a prerequisite) and the GRPO invariant (Batch 5 — the disaggregated example's headline path). The rest are should-fix/nits. Nice work overall.

KeitaW (changes requested):
Few comments

allela-roy:
Thanks for the review @KeitaW .  I have fixed the issues raised. 
On the c5 under-provisioning , we  provision the group with ThreadsPerCore=1 (hyperthreading off), so although ml.c5.4xlarge is 16 physical vCPU, EKS only exposes ~8 logical / 7.9 allocatable (I've this verified on the live nodes: capacity.cpu=8, allocatable.cpu=7910m). So TORCH_NUM_THREADS=6 alongside the cpu: 6 request is right-sized, not under-provisioned.

KeitaW (approved):
LGTM! Thanks @allela-roy

PR #1137, opened 2026-06-17
feat(aws-pcs): strengthen cluster operations & administration (multi-user, multi-AZ, IAM, Regions, SSH)

DaisukeMiyamoto:
## Summary

A major update to the AWS PCS reference architecture (`architectures/aws-pcs/`) that
strengthens **operations & administration**. It adds **five user-facing capabilities** on
top of the existing cluster:

1. **Multi-user support (OpenLDAP)**
2. **Multi-AZ VPC / subnets**
3. **cluster-admin / cluster-user IAM policy stacks**
4. **Expanded Region coverage**
5. **SSH access to the login node**

Validated end-to-end on real hardware (hpc6a/hpc8a, p6-b200 ×4) across 10 Regions.

> ⚠️ **Contains breaking changes** (parameter renames, VPC CIDR layout). See "Breaking
> changes & migration" below.

---

## A. New features (direct user benefit)

Each item: *what you can do / how it's built / constraints.*

### A-1. Multi-user support (OpenLDAP)
- **What**: share one cluster across multiple users; an LDAP user resolves to the same
  UID on every node and can run Slurm jobs directly.
- **How**: an OpenLDAP server runs on the login node; compute nodes auto-join via SSSD.
  The user database lives on FSx OpenZFS `/home`.
- **Constraint**: because the DB is on `/home`, it survives a login-node replacement, but
  the directory *service* is a single point of failure — enabling it implies a
  single-login-node cluster.

### A-2. Multi-AZ VPC / subnets
- **What**: spread the VPC's private subnets across up to 3 AZs for higher availability or
  to match capacity (Capacity Block / reservation) in a specific AZ.
- **How**: the prerequisites stack creates additional per-AZ private subnets; the NAT
  gateway stays in the primary AZ (lowest cost).
- **Constraints**: the VPC CIDR layout changes (see B-2 / breaking changes); both shared
  filesystems (FSx for Lustre `/fsx` and OpenZFS `/home`) are created **single-AZ in the
  primary AZ**, so compute nodes in the additional AZs reach storage cross-AZ.

### A-3. cluster-admin / cluster-user IAM policy stacks
- **What**: distribute least-privilege policies for cluster operators and users as
  ready-to-deploy CloudFormation stacks.
- **How**: separate policy + group stacks for admin and user. The admin policy alone can
  run `deploy-all` to completion (no `iam:CreatePolicy` needed — see B-3).
- **Constraint**: the cluster user gets SSM access only and cannot read the LDAP admin
  password.

### A-4. Expanded Region coverage
- **What**: know what works where (deploy / monitoring / storage / Pyxis / GPU) and avoid
  Region-specific pitfalls.
- **How**: deploy-verified in 10 Regions and recorded in a region-coverage table (with the
  working storage deployment type and the verification date).
- **Constraint**: some Regions don't support the default storage deployment type (e.g.
  OpenZFS in Osaka) — the fallback value is noted in the table.

### A-5. SSH access to the login node
- **What**: SSH directly to the login node from a trusted CIDR (no bastion / SSM
  port-forward required).
- **How**: a login-node-only security group opens just the requested port from the given
  CIDR; compute nodes stay private. Grafana (443) uses the same mechanism.
- **Constraint**: off by default (no extra security group is created unless a CIDR is set).

---

## B. Supporting changes (needed to make the features above work)

Each item notes *why it was needed*.

### B-1. Parameter cleanup (rename / consolidate / remove)
**Why**: keep deploy-all's parameter naming and granularity consistent as the new features
(SSH/Grafana access, monitoring) were added. **Callers must update** (see breaking changes).
- Rename `DeployMonitoring` (bool) → `MonitoringStack` (enum); behaviour is equivalent.
- Rename `GrafanaPublicAccessCidr` → `GrafanaAccessCidr` (pairs with `SSHAccessCidr`).
- Remove `OnDemandEnableEfa`, consolidating into `OnDemandEfaInterfaceCount` (0/1/2).
- Remove `VPCName` (the VPC is named from the stack name).
- Reorganize the console ParameterGroups (monitoring/directory grouped together).

### B-2. VPC private-subnet CIDR split (for A-2 multi-AZ)
**Why**: carving out subnets for the additional AZs required splitting the VPC CIDR into
more blocks.
- The primary subnet's size changes; the rest is reserved for the additional AZs.
- **The primary subnet's CIDR changes even without adding AZs** (breaking) — see the table.

### B-3. Inline split of the PCS instance-role permissions (for A-3 IAM)
**Why**: to let the admin policy run `deploy-all` without `iam:CreatePolicy`, the managed
policy that required it had to go.
- The managed policy is removed; the permissions become per-feature inline policies
  (always-on vs monitoring-only).

### B-4. PostInstallScriptUrl fetch method / location (for A-1, A-4)
**Why**: to distribute the directory setup script and to fetch boot scripts in accounts
that can't use public S3, the source moved to S3 (fetched via the instance role).
- Adds `s3://` scheme support (works with a private bucket, no public S3); `http(s)://`
  still works as before.
- Boot scripts moved under `assets/scripts/`, included in the distribution-bucket sync.
- The default source changes from GitHub raw → S3 (see the table).

### B-5. Docs / tests reorganization (for all features)
**Why**: the README grew as features were added; basic flow vs advanced features needed
separating, and parameter drift needed automated detection.
- README restructured (basic flow first, advanced features under `§8`; the `§4`
  Configuration parameter tables are grouped by — and named after — the deploy-all console
  ParameterGroups).
- Feature guides organized under `docs/`: `IAM.md` / `USER-MANAGEMENT.md` / `CUSTOM-AMI.md`
  (AMI pre-bake) / `DEPLOY-TESTING.md` (deploying unpublished templates from your own
  bucket) / `PARAMETERS.md` / `OPERATIONS.md`. Verification records moved into `tests/` as
  reproducible procedures.
- Test guide split into category files (infra / compute / training / hpc-efa / storage /
  multi-user / gpu-healthcheck / iam); the README is an index + matrix.
- Added a docs-consistency lint (`tests/lint-docs.sh`: stale/renamed params, undocumented
  params, broken anchors) as pre-merge Test 0.

---

## C. Bug fixes (found while adding the features)

- VPC `Name` tag was hard-coded; now derived from the stack name.
- first-boot apt/dpkg lock contention (unattended-upgrades) — added wait + retry.
- cloud-init directory setup moved from a MIME shellscript to a runcmd block (PCS does not
  execute the shellscript part).

---

## Breaking changes & migration

| Change | Old | New | Migration |
|---|---|---|---|
| Param rename | `DeployMonitoring=true/false` | `MonitoringStack=Prometheus-LoginNode/none` | Update deploy-all invocations |
| Param rename | `GrafanaPublicAccessCidr` | `GrafanaAccessCidr` | Same |
| Param consolidate | `OnDemandEnableEfa` + `OnDemandEfaInterfaceCount` | `OnDemandEfaInterfaceCount=0/1/2` | 0 = off, 1/2 = on |
| Param removal | `VPCName` | (removed; named from stack) | Stop passing it |
| **VPC CIDR layout** | primary private subnet `/17` | `/18` (+ reserved for additional AZs) | **Not an in-place update** (subnets are recreated); deploy fresh |
| IAM | managed policy | inline policy | A monitoring-enabled stack update replaces the IAM resource |

---

## Validation

- **Environments**: us-east-2 (CPU/EFA general), us-west-2 (Slurm-version differences),
  ap-south-1 (GPU: p6-b200 ×4, Capacity Block), plus 10 Regions total. Both Slurm 25.05
  and 25.11.
- **What was exercised**:
  - Major-update e2e (single deploy-all): multi-AZ subnet placement, login SSH via
    `SSHAccessCidr`, monitoring stack, OpenLDAP server + compute SSSD auto-join, Slurm jobs
    as an LDAP user with multi-node UID consistency and user-delete propagation, EFA
    on/off, CFN lint on every template.
  - Real EFA traffic (hpc8a, 2 nodes, OSU benchmarks).
  - GPU (p6-b200 ×4): NCCL all_reduce, Megatron-LM GPT-3 distributed training, GPU health
    check.
  - Least-privilege IAM: create→delete deploy-all as the cluster-admin role (no
    `iam:CreatePolicy`); SSM-only access as the cluster-user (cannot read the LDAP password).
  - Region coverage: deploy + monitoring + storage + Pyxis across 10 Regions.
- Detailed procedures and results live under `architectures/aws-pcs/tests/`.

---

## Note for maintainers

The templates fetch the nested CFN templates and boot scripts from a distribution S3
bucket (under `S3BucketName`/`S3KeyPrefix`; boot scripts under `assets/scripts/`).
**After merge, `architectures/aws-pcs/assets/` (templates + `scripts/`) must be synced to
the distribution bucket** — otherwise nested-stack fetches and node first-boot
(Enroot/Pyxis, directory setup) will fail.


architectures/aws-pcs/assets/cluster.yaml
```diff
@@ -159,16 +159,55 @@ Resources:
...
+
+              - Sid: SSMClusterParametersForThisClusterOnly
+                Effect: Allow
+                Action:
+                  - ssm:GetParameter
+                  - ssm:PutParameter
+                  - ssm:AddTagsToResource
+                Resource:
+                  - !Sub 'arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/pcs/${PCSCluster.Id}/grafana/*'
+                  - !Sub 'arn:${AWS::Partition}:ssm:${AWS::Region}:${AWS::AccountId}:parameter/pcs/${PCSCluster.Id}/ldap/*'
```
KeitaW, line 275: ### Compute nodes can read & decrypt the OpenLDAP admin password → directory takeover (blocking)

**Observation.** `setup-directory.sh` (server role) stores the generated OpenLDAP admin password in SSM at `/pcs/${CLUSTER_ID}/ldap/admin-password` as a `SecureString`. This `PcsInstanceBasePolicy` statement is *always-on* (not gated behind any directory/monitoring condition) and grants `ssm:GetParameter`/`PutParameter` on `/pcs/${PCSCluster.Id}/ldap/*`, and the sibling `KMSForSSMSecureString` statement grants `kms:Decrypt` gated only by `kms:ViaService = ssm.*`. That single instance profile is attached to **every** compute node group — `LoginNodeGroupStack`, `OnDemandCNGStack`, `PseriesCNGStack`, `P6B200CNGStack`, `P6B300CNGStack` all pass `IamProfileArn: !GetAtt ClusterStack.Outputs.InstanceProfileArn`. The compute launch templates set `HttpPutResponseHopLimit: 4`.

**Impact.** A compute node runs untrusted user Slurm jobs. With hop limit 4, a user process (even inside a container) can reach IMDSv2, assume the instance role, and run `aws ssm get-parameter --with-decryption --name /pcs/<id>/ldap/admin-password`. With the `cn=admin` bind it can then `ldappasswd`/`ldapmodify` any account — reset passwords, change `uidNumber`, inject `sshPublicKey`. That is full directory takeover, which defeats the multi-user isolation this PR exists to provide. Crucially, **compute does not need this secret at all**: the SSSD config `setup-directory.sh` writes uses an *anonymous* bind (no `ldap_default_bind_dn` / `ldap_default_authtok`), so clients resolve users without any credential.

I verified this on a live CPU-only deployment of this branch (2026-06-18): the SecureString exists, and `iam simulate-principal-policy` on the shared role (attached to compute node `PCS-cpu1`) returns **allowed** for both `ssm:GetParameter` on the ldap param and `kms:Decrypt` via SSM.

**Suggestion.** Keep the `ldap/*` SSM read/write + `kms:Decrypt` off the role attached to compute. The cleanest fix mirrors the split you already made for monitoring: give the **login/server** node group its own instance role/profile that carries the `ldap/*` + decrypt permissions, and drop the `ldap/*` statement from the shared/compute base policy entirely. (The `grafana/*` read has the same shape but far lower stakes — worth scoping to the login node in the same change, since compute never serves Grafana.) This is structural (a new role + profile + wiring through `add-cng`/`deploy-all`), so no one-click patch — happy to sketch the resource split if useful.

*Ref:* [IMDSv2 instance-role credentials are reachable by in-instance processes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html).

KeitaW (commented):
## Review Batch 1/5 — IAM & Privilege Boundaries

Three findings on the new IAM surface; the first is a blocker. Details inline.

architectures/aws-pcs/assets/cluster.yaml
```diff
@@ -159,16 +159,55 @@ Resources:
...
                   - pcs:RegisterComputeNodeGroupInstance
                 Resource: "*"
 
+              - Sid: S3GetScripts
+                Effect: Allow
+                Action:
+                  - s3:GetObject
+                Resource:
+                  - arn:aws:s3:::awsome-distributed-ai/templates/scripts/*
+                  - arn:aws:s3:::*/templates/scripts/*
```
KeitaW: ### `s3:GetObject` on `arn:aws:s3:::*/templates/scripts/*` is broader than needed

**Observation.** The base policy's `S3GetScripts` statement lists the canonical `arn:aws:s3:::awsome-distributed-ai/templates/scripts/*` **and** a wildcard `arn:aws:s3:::*/templates/scripts/*`.

**Impact.** The second ARN grants the instance role `s3:GetObject` on any object keyed `templates/scripts/*` in *any* bucket the account can reach (and any public bucket). It's there to support a custom distribution bucket, but it's much wider than the one bucket actually used.

**Suggestion.** Scope to the actual distribution bucket. `cluster.yaml` doesn't currently receive `S3BucketName` (it's not in the `ClusterStack` parameters), so the tidy fix is to plumb `S3BucketName`/`S3KeyPrefix` into this template (as you already did for `add-cng`) and use `arn:${AWS::Partition}:s3:::${S3BucketName}/${S3KeyPrefix}scripts/*`. If you'd rather not add the param now, at least replace the bare `*/...scripts/*` with the named bucket(s) you actually deploy from. (Note: dropping the wildcard outright would break the custom-bucket flow in `docs/DEPLOY-TESTING.md`, so plumb the param rather than just deleting the line.)

architectures/aws-pcs/assets/cluster-user-iam.yaml
```diff
@@ -0,0 +1,139 @@
...
+              "Resource": [
+                "arn:aws:ec2:*:*:instance/*",
+                "arn:aws:ssm:*:*:document/AWS-StartSSHSession",
+                "arn:aws:ssm:*:*:document/AWS-StartPortForwardingSession",
+                "arn:aws:ssm:*:*:document/AWS-StartPortForwardingSessionToRemoteHost",
+                "arn:aws:ssm:*:*:document/SSM-SessionManagerRunShell"
+              ],
+              "Condition": {
+                "StringLike": {
+                  "ssm:resourceTag/Name": "PCS-login*"
```
KeitaW, line 89: ### cluster-user SSM policy denies SSH-over-SSM / port-forward (the documents are tag-gated)

**Observation.** The single `SSMSessionToLoginNode` statement gates the **instance ARN and the SSM document ARNs together** (`AWS-StartSSHSession`, `AWS-StartPortForwardingSession*`, `SSM-SessionManagerRunShell`) under one `StringLike ssm:resourceTag/Name: PCS-login*` condition.

**Impact.** `ssm:resourceTag/Name` is a tag on the *instance*; the SSM **document** resources have no `Name` tag, so for the document sub-authorization the key is absent and `StringLike` (without `...IfExists`) evaluates false. `ssm:StartSession` with a custom document authorizes against *both* the instance and the document, so SSH-over-SSM and port-forwarding (the headline cluster-user feature) **fail closed**. A plain Session Manager shell to the login node still works (instance-only authz), which is why a quick smoke test passes — so the instance scoping itself is correct; only the document gating is wrong. The instance side is fine: `LoginNodeGroupStack` hard-codes `CngName: login` → `Name=PCS-login`, and compute CNGs use other names, so `PCS-login*` does isolate login from compute.

Confirmed via `iam simulate-custom-policy` against a live deployment of this policy (2026-06-18): `ssm:StartSession` on `document/AWS-StartSSHSession` → **implicitDeny**; on a login instance (`Name=PCS-login`) → allowed; on a compute instance (`Name=PCS-compute`) → deny.

**Suggestion.** Split into a tag-gated *instance* statement + an untagged *document* statement (the AWS pattern):

```json
{ "Sid": "SSMSessionToLoginNodeInstance", "Effect": "Allow", "Action": ["ssm:StartSession"],
  "Resource": "arn:aws:ec2:*:*:instance/*",
  "Condition": { "StringLike": { "ssm:resourceTag/Name": "PCS-login*" } } },
{ "Sid": "SSMSessionDocuments", "Effect": "Allow", "Action": ["ssm:StartSession"],
  "Resource": [ "arn:aws:ssm:*:*:document/AWS-StartSSHSession",
                "arn:aws:ssm:*:*:document/AWS-StartPortForwardingSession",
                "arn:aws:ssm:*:*:document/AWS-StartPortForwardingSessionToRemoteHost",
                "arn:aws:ssm:*:*:document/SSM-SessionManagerRunShell" ] }
```

I've opened this as a ready-to-merge PR against this branch — **https://github.com/DaisukeMiyamoto/awsome-distributed-ai/pull/4** — verified end-to-end with `iam simulate-custom-policy` before/after (SSH document `implicitDeny → allowed`, port-forward `→ allowed`, login instance `allowed`, compute instance still `deny`). Feel free to take the commit or the diff.

*Ref:* [Session Manager — restrict access by instance tag](https://docs.aws.amazon.com/systems-manager/latest/userguide/getting-started-restrict-access-quickstart.html) (instance vs document statements are separate in the AWS example).

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
```diff
@@ -0,0 +1,70 @@
...
+
+set -euo pipefail
+
+USERNAME="${1:?Usage: $0 <username> [uid] [gid] [ssh-pub-key]}"
+USER_UID="${2:-$((10000 + RANDOM % 50000))}"
+USER_GID="${3:-3000}"
+SSH_PUBKEY="${4:-}"
+
+# Auto-detect LDAP config from sssd.conf or environment
+LDAP_DOMAIN_SUFFIX="${LDAP_DOMAIN_SUFFIX:-$(grep ldap_search_base /etc/sssd/sssd.conf 2>/dev/null | awk -F= '{print $2}' | tr -d ' ' || echo 'dc=cluster,dc=internal')}"
```
KeitaW: ### LDAP suffix auto-detect truncates `dc=cluster,dc=internal` → `dc` (blocking)

**Observation.** When `LDAP_DOMAIN_SUFFIX` is not already in the environment, the script derives it from `sssd.conf` with `... | awk -F= '{print $2}' | tr -d ' '`. The `sssd.conf` line `setup-directory.sh` writes is `ldap_search_base = dc=cluster,dc=internal` — whose **value itself contains `=`**. `awk -F=` splits on every `=`, so `$2` is ` dc`, and after `tr -d ' '` the suffix becomes `dc`.

**Impact.** Every DN is then built wrong — `LDAP_ADMIN_DN=cn=admin,dc`, user dn `uid=<u>,ou=People,dc` — and `ldapadd`/`ldappasswd` fail, so the documented add-a-user helper fails on its default invocation. `USER-MANAGEMENT.md` shows `sudo LDAP_ADMIN_PASSWORD=... ldap-add-user.sh alice 10001 3000` (no `LDAP_DOMAIN_SUFFIX`), so this is the normal path, and it breaks precisely when `sssd.conf` is present (the common case). Reproduced against the exact line the template writes (2026-06-18): the pipeline returns `dc`.

**Suggestion.** Split only on the first `= ` so the value stays intact:

```suggestion
LDAP_DOMAIN_SUFFIX="${LDAP_DOMAIN_SUFFIX:-$(sed -n 's/^ldap_search_base[[:space:]]*=[[:space:]]*//p' /etc/sssd/sssd.conf 2>/dev/null || echo 'dc=cluster,dc=internal')}"
```

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
```diff
@@ -0,0 +1,70 @@
...
+# Usage: ./ldap-add-user.sh <username> [uid] [gid] [ssh-pub-key]
+#
+# Example:
+#   ./ldap-add-user.sh alice 10001 3000
+#   ./ldap-add-user.sh bob 10002 3000 "ssh-rsa AAAA..."
+
+set -euo pipefail
+
+USERNAME="${1:?Usage: $0 <username> [uid] [gid] [ssh-pub-key]}"
+USER_UID="${2:-$((10000 + RANDOM % 50000))}"
```
KeitaW: ### Random UID has a dead modulo and no uniqueness check

**Observation.** `USER_UID="${2:-$((10000 + RANDOM % 50000))}"`. Bash `RANDOM` is capped at 32767, so `RANDOM % 50000 == RANDOM` always — the `% 50000` is dead and the effective range is **10000–42767**, not the intended ~60000 (and `sssd.conf` sets `max_id = 60000`). There is also no check that the chosen `uidNumber` is unused.

**Impact.** Two users assigned the same `uidNumber` are the *same* POSIX principal on shared `/home`/`/fsx` — they can read and overwrite each other's files. A silent random default makes that collision possible, and UID uniqueness is foundational to the multi-user isolation this PR adds.

**Suggestion.** The documented usage always passes an explicit UID, so the simplest robust fix is to require it rather than silently randomize:

```suggestion
USER_UID="${2:?uid required — pick a cluster-unique uidNumber (see USER-MANAGEMENT.md)}"
```

If you want to keep auto-allocation, derive `max(uidNumber)+1` from an `ldapsearch` and verify it's free before `ldapadd`, instead of a bounded `RANDOM`.

KeitaW (commented):
## Review Batch 2/5 — Multi-User Directory Scripts

Two issues in `ldap-add-user.sh` inline. Non-blocking notes: directory-client discovery in `setup-directory.sh` has no retry if the login node isn't running/tagged yet when a compute node boots (eventual-consistency in practice, since PCS brings the login node up first and SSSD reconnects); and LDAP traffic is cleartext `ldap://` with no TLS — acceptable within the private VPC given anonymous-bind resolution and SSH-pubkey login, but worth a one-line note in the docs.

KeitaW (commented):
## Review Batch 3/5 — Network Access Controls

One finding inline.

KeitaW (commented):
## Review Batch 4/5 — Repository Hygiene

Header fix inline. Minor: `cfn-lint` also reports `DirectoryService` is declared but unused in the four `add-cng*` templates (only `DirectoryRole` is referenced in their UserData) — harmless dead parameter, but trivially removable.

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
KeitaW, line 1: ### MIT-0 header missing on the new shell scripts

**Observation.** `ldap-add-user.sh`, `setup-directory.sh`, and `tests/lint-docs.sh` open with the shebang and go straight to a description comment — no copyright/license header. The sibling `assets/scripts/install-enroot-pyxis.sh` (renamed in this same PR) carries it right after the shebang, and the two new IAM YAML templates added here do too.

**Suggestion.** Add the two-line header after the shebang (shown for this file; **please also add it to `setup-directory.sh` and `tests/lint-docs.sh`**, the latter with its `#!/usr/bin/env bash` shebang):

```suggestion
#!/bin/bash
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT-0
```

KeitaW (commented):
## Review Batch 5/5 — Evaluation, Positives & Sources

I deployed this branch end-to-end (CPU-only, `DirectoryService=OpenLDAP-LoginNode`, no GPU/monitoring) from a private staging bucket in us-west-2 to validate the findings; `cfn-lint` is clean (no errors) and the multi-user cluster booted (login + compute, OpenLDAP server + SSSD client). The findings above marked "confirmed live" were verified against that deployment.

### Things That Look Great
- **`tests/lint-docs.sh` is the right kind of automation.** It bans the old parameter names (`OnDemandEnableEfa`, `GrafanaPublicAccessCidr`, `DeployMonitoring=true`, the "S3 hosting is not allowed" phrase, the removed `iam/` dir) with an allow-context escape for history notes, and gates that every `deploy-all` parameter has a `PARAMETERS.md` row. I checked the complement: all eight new/renamed params are documented and the only surviving mentions of removed params are in `(renamed from …)` / `(removed …)` context — no doc drift leaks through.
- **`PostInstallScriptUrl` default fixed the right way.** Moving off the hard-coded `raw.githubusercontent.com/...main` default to empty-→-`s3://<bucket>/<prefix>scripts/...` (scheme-dispatched `aws s3 cp` vs `curl`) removes a first-boot 404 / wrong-org-pin footgun and works in private-S3 accounts.
- **`OnDemandEnableEfa` → `OnDemandEfaInterfaceCount=0/1/2` consolidation** is clean: `EfaEnabled` becomes `count != 0` and `Has2ndEfaInterface` simplifies to `count == 2` (equivalent since `2` implies enabled).
- **Multi-AZ CIDR split is arithmetically sound:** `!Cidr [CIDR1, 4, 14]` carves the `/16` into four `/18`s, indices 0/1/2 for primary/AZ2/AZ3, additional subnets condition-gated and sharing the primary route table; the breaking primary-subnet resize is correctly called out in the migration table.
- **Directory-role discovery is well-isolated** from the monitoring stack (`directory-role` tag distinct from `monitoring-role`), with a clear single-login-node constraint documented in the script and parameter help.

### Sources
- IMDSv2 hop limit / instance-role credential reach: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html
- Session Manager tag-restricted access (instance vs document statements separate): https://docs.aws.amazon.com/systems-manager/latest/userguide/getting-started-restrict-access-quickstart.html
- `bash RANDOM` is `[0, 32767]` → `RANDOM % 50000 == RANDOM` (verified live, 2026-06-18).
- LDAP suffix `awk -F=` truncation reproduced against the exact `sssd.conf` line this PR writes (verified live, 2026-06-18).
- Shared instance profile across all 5 CNG stacks; anonymous SSSD bind on clients; `Name=PCS-login` from hard-coded `CngName: login`; cluster-user SSM document gating — all verified live on a deployment of this branch (2026-06-18).
- MIT-0 header precedent: `architectures/aws-pcs/assets/scripts/install-enroot-pyxis.sh`.

KeitaW (changes requested):
Few comments + One proposal via PR.

architectures/aws-pcs/assets/cluster.yaml
DaisukeMiyamoto: Thanks for flagging this — it sent me down a useful rabbit hole.

Looking again, that inline `S3GetScripts` statement was **redundant**: the instance role already attaches the AWS-managed `AmazonS3ReadOnlyAccess` (`s3:Get*`/`List*` on `*`), so a narrower `templates/scripts/*` grant added nothing, and the wildcard bucket you flagged was already moot under the broader managed policy. I've **removed the inline statement from both roles** (base + login) rather than scope it — boot-script reads (incl. the `s3://` PostInstall path on a private bucket) are still covered by the managed policy. Verified end-to-end on a live multi-user cluster (CPU + OpenLDAP): the `s3://` post-install fetch from a private bucket still works, Pyxis installs, and compute is still denied the LDAP admin secret.

The bigger finding: the real over-grant is **`AmazonS3ReadOnlyAccess` being attached unconditionally**. Upstream [`aws-hpc-recipes` `pcs-iip-minimal`](https://github.com/aws-samples/aws-hpc-recipes/blob/main/recipes/pcs/getting_started/assets/pcs-iip-minimal.yaml) makes it an **opt-in** (`EnableS3ReadOnly`, off by default), but that gate was lost when the instance role was brought into `cluster.yaml` (in the earlier monitoring PR, not this one) — so today every node gets account-wide S3 read by default.

I'm **not** changing that in this PR: it's a pre-existing IAM-behaviour change and the training test cases (FSDP, Megatron) currently rely on reading data from S3, so tightening it needs its own validation pass. I've added it to [`docs/ROADMAP.md`](architectures/aws-pcs/docs/ROADMAP.md) to restore the upstream opt-in (or scope to named data/template buckets) in a dedicated follow-up. (`AmazonSSMManagedInstanceCore` is also unconditional, but SSM is a core feature here — connect, cluster-user policy — so it stays on by default.)

architectures/aws-pcs/assets/cluster.yaml
DaisukeMiyamoto: Fixed — and verified live. The OpenLDAP admin secret is now off the compute role:

- Split the instance role in `cluster.yaml`: a **login-only** role/profile (`PcsLoginInstanceIamRole`/`Profile`) carries the `ldap/*` SSM read/write + `kms:Decrypt`; the shared **compute** base role keeps only `grafana/*`. `deploy-all` passes `LoginInstanceProfileArn` to the login CNG and `InstanceProfileArn` to all compute CNGs.
- Important subtlety your review surfaced: removing `kms:Decrypt` from the compute role was **not enough**. `AmazonSSMManagedInstanceCore` grants `ssm:GetParameter` on `*`, and the SecureString uses the AWS-managed key `alias/aws/ssm` whose key policy lets any in-account principal decrypt via SSM — so a compute node could still `get-parameter --with-decryption` the secret. I added an explicit **Deny** on `ssm:GetParameter*` for `/pcs/<id>/ldap/*` to the compute base policy (explicit Deny overrides the managed-policy Allow).
- Verified end-to-end on a fresh compute node: `get-parameter --with-decryption` on the ldap secret → **BLOCKED**; the login node still reads it; SSSD on compute still resolves users via anonymous bind.

(One follow-up for the login role name: PCS requires the instance-profile role name to start with `AWSPCS`, so the new login role is `AWSPCS-pcs-login-<hash>-<region>-role` — without that, `CreateComputeNodeGroup` rejects it.)

architectures/aws-pcs/assets/cluster-user-iam.yaml
DaisukeMiyamoto: Merged your fix — thank you for the ready-to-go PR. I cherry-picked the commit onto this branch (instance statement tag-gated, document statement separate), so SSH-over-SSM and port-forwarding now authorize correctly for the cluster-user. Credited to you as the commit author.

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
DaisukeMiyamoto: Fixed. Replaced `awk -F=` (which split on the `=` inside the value) with `sed -n 's/^ldap_search_base[[:space:]]*=[[:space:]]*//p'`, so the suffix stays intact. Verified live: the helper now derives `dc=cluster,dc=internal` (not `dc`) and the default `ldap-add-user.sh alice 10001 3000` path works.

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
DaisukeMiyamoto: Done — went with your "require it" suggestion: `USER_UID="${2:?uid required ...}"`. The silent bounded-RANDOM default is gone (you're right that `RANDOM % 50000 == RANDOM` and that a collision means a shared POSIX principal on `/home`,`/fsx`), so the helper now fails fast if no uid is given. USER-MANAGEMENT.md already always passes an explicit uid.

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
DaisukeMiyamoto: Added the MIT-0 SPDX header to all three: `ldap-add-user.sh`, `setup-directory.sh`, and `tests/lint-docs.sh`.

DaisukeMiyamoto:
Thanks for the thorough, hardware-validated review — it caught two real security bugs. All of your findings are addressed; rundown:

**Blockers (fixed + verified live on a multi-user CPU cluster):**
- **LDAP admin secret reachable from compute** → split into a login-only IAM role/profile carrying `ldap/*` SSM + `kms:Decrypt`; compute keeps only `grafana/*`, **plus an explicit Deny** on `ldap/*` (needed because `AmazonSSMManagedInstanceCore` + the `alias/aws/ssm` key policy would otherwise let compute decrypt it). Verified: compute `--with-decryption` → BLOCKED, login still reads it.
- **LDAP suffix truncation** → `awk -F=` → `sed`; derives `dc=cluster,dc=internal` correctly.

**Your cluster-user SSM PR** → cherry-picked onto this branch (credited to you).

**Non-blocking, also addressed:**
- Random uid → now required (`${2:?...}`).
- MIT-0 headers added to the three scripts.
- Dead `DirectoryService` param removed from `add-cng*` (only `DirectoryRole`/`DirectoryDomainSuffix` are used).
- `s3:GetObject` finding → the inline statement was redundant under the role's `AmazonS3ReadOnlyAccess`, so I removed it entirely.

**Deferred to a follow-up (ROADMAP), with reasoning in the inline thread:** the real over-grant is `AmazonS3ReadOnlyAccess` being attached **unconditionally** — upstream `aws-hpc-recipes` makes it opt-in, but that gate was lost when the role moved into `cluster.yaml` (pre-existing, not this PR). Tightening it is an IAM-behaviour change that the training test cases (FSDP/Megatron read from S3) depend on, so it gets its own validation pass rather than riding this PR.

I also did a side-effect pass: confirmed the default (`DirectoryService=none`) path is unaffected (login role/profile are created unconditionally but the ldap grants simply go unused), and re-verified the full deploy after each change.

KeitaW (approved):
LGTM

DaisukeMiyamoto:
Thanks for the review and the merge! 🙏

**One follow-up needed before the published templates work: sync `assets/` to the distribution S3 bucket.**

The templates fetch their nested stacks and first-boot scripts from `s3://awsome-distributed-ai/templates/…`, but that bucket isn't updated by the merge itself. Until someone with write access runs the sync, the public bucket serves the **pre-merge** templates, so:

- **`s3://awsome-distributed-ai/templates/pcs-ml-cluster-deploy-all.yaml` is still the old version** — the one-click / `--template-url` deploy from the bucket won't have the new parameters (`MonitoringStack`, `SSHAccessCidr`, `DirectoryService`, `AdditionalSubnetAZ2/3`, …).
- **The boot scripts moved** `architectures/aws-pcs/scripts/` → `architectures/aws-pcs/assets/scripts/` in this PR. So `s3://…/templates/scripts/install-enroot-pyxis.sh` (the path the published templates and the workshop point at) currently **404s**, and a node's first-boot Enroot/Pyxis install fails until the new `assets/scripts/` content is synced. (GitHub `main` is already correct — only the S3 mirror lags.)

So the published/bucket-based path is effectively broken in the window between merge and sync. Could a maintainer run the sync from the merged `main`?

```bash
cd architectures/aws-pcs
aws s3 sync assets/ s3://awsome-distributed-ai/templates/ \
  --exclude "*" --include "*.yaml" --include "*.sh"
```

(Deploying directly from a checkout, or from a personal bucket via `S3BucketName`/`S3KeyPrefix`, works today — it's only the public-bucket default that needs the sync.)

I'll update the AI/ML PCS workshop's `POST_INSTALL_SCRIPT_URL` to the new `assets/scripts/` path in lockstep.

Issue #25, opened 2023-10-24
FIx docker install


mhuguesaws:
Docker role does not install due to missing automatic `yes` to package install

PR #1141, opened 2026-06-20
fix(aws-pcs): harden multi-user OpenLDAP for production (compute LDAP join, password idempotency, replacement docs)

DaisukeMiyamoto:
Follow-up hardening for the multi-user OpenLDAP directory shipped in #1137.

Rather than a quick smoke test, I ran the multi-user setup the way a real cluster would be used — for the better part of a day on live hardware: several LDAP users and accounts submitting Pyxis/Enroot and multi-node jobs over time, plus the operational events a long-lived cluster actually hits (compute scaling up and down, and login-node replacement). That longer soak surfaced a handful of issues that a short deploy-and-run check doesn't: they only appear once nodes have churned, credentials have aged, or the login node has been replaced.

Each item below is a concrete problem that showed up during that run, followed by the change that addresses it.

## 1. A compute node could boot before the directory server was discoverable

`setup_client()` looked up the `directory-role=server` instance once. If a compute node booted before the login node finished tagging itself (scale-up race, or right after a login-node replacement), the one-shot lookup returned nothing and the node came up **without LDAP**, staying that way until reboot.

**Fix** (`setup-directory.sh`): retry the tag lookup with backoff (up to ~5 min). Verified on a clean deploy and after a replacement — compute discovers the (new) login IP and resolves users.

## 2. Login-node replacement silently rotated the LDAP admin password

The admin password was regenerated on every boot. The user DB on `/home/ldap-db` is persistent, so a login-node replacement kept all the data but changed the password — breaking any admin credential an operator had cached, and leaving SSM out of sync with what people had.

**Fix** (`setup-directory.sh`): `setup_server()` reuses the existing SSM password if present, only generating one on first boot. Verified across a real replacement: password SHA-256 identical before/after, all users survive, admin bind with the SSM value succeeds on the new node.

## 3. A failed LDAP client setup left no signal

The client setup ran as `bash setup-directory.sh ... | tee logfile`, so the pipe's exit status was always `tee`'s `0` — a genuine failure (e.g. discovery exhausting its retries in item 1) was masked and nothing recorded it.

**Fix** (`add-cng.yaml`): redirect to the log and capture the script's real `$?`, landing an explicit `ERROR` in `/var/log/directory-setup.log`. The node is intentionally **not** drained — Slurm launches jobs by numeric UID, so a job still runs on a node where the user can't be name-resolved (verified: a job submitted as `alice` reached COMPLETED on a node where `id alice` returned "no such user"); a missing directory client is a degradation, not a hard failure, so a log signal is the right response.

## 4. Login-node replacement behaviour wasn't documented

What survives a replacement and what doesn't was undocumented, both for the directory and for monitoring.

**Fix** (docs):
- `USER-MANAGEMENT.md`: post-replacement recovery for already-running compute (stale SSSD `ldap_uri`; name resolution degrades but jobs keep running); public-IP / SSH host-key change; the transient "user not found" right after a node boots (SSSD `enumerate=true`); two accounting reporting gotchas (`sacct --state=` needs `-E now`; `sreport` lags the hourly rollup).
- `OPERATIONS.md`: a "Monitoring across a login-node replacement" subsection (compute scraping re-discovers via EC2 SD; Grafana password reused from SSM; built-in dashboards file-provisioned; but node-local Prometheus TSDB / Grafana DB lose history). Also consolidates the two AMI-related sections into one.
- `ROADMAP.md`: stable LDAP endpoint across replacement; persist the monitoring TSDB / Grafana DB.
- `tests/multi-user-test.md`: reproducible cases for password-preserved-across-replacement, stale-ldap_uri recovery, and job-runs-by-UID.
- `README.md`: a faster Quick Start -> first-job path with a **CPU** job example that runs on the default `cpu1` queue (no GPU); short user-facing warnings about the replacement and accounting-report gotchas.

## Verified end-to-end on real hardware (us-east-2)

- Fresh `deploy-all` cluster from these templates: compute joins LDAP cleanly from first boot, jobs run as the right users, the README CPU example runs (2 distinct nodes, Pyxis import), the intentional failure records FAILED.
- Login-node replacement: password preserved, DB survived, fresh compute found the new login IP; the documented recovery restored resolution on an already-running node non-disruptively.
- User lifecycle (add/group/passwd/delete) propagates to a running compute without reboot.

`tests/lint-docs.sh` PASS; `add-cng.yaml` passes `validate-template`; the rendered `runcmd` block was checked under `dash`.

> A related monitoring fix (Prometheus stops discovering compute ~6h after boot when credentials rotate) is in the separate `aws-parallelcluster-monitoring` repo — aws-samples/aws-parallelcluster-monitoring#53. This repo only consumes that stack via `MonitoringRepo` / `MonitoringVersion`.


KeitaW (commented):
## Review Batch 1/3 — Operational robustness

architectures/aws-pcs/assets/add-cng.yaml
```diff
@@ -532,7 +532,31 @@ Resources:
...
+                #  - Auto-draining a fresh node would risk a launch/drain/replace
+                #    loop: PCS replaces the "unhealthy" node, the replacement hits
+                #    the same condition and drains again. Log-and-continue is the
+                #    safe choice; an operator can act on the logged ERROR.
+                bash /tmp/setup-directory.sh "${DirectoryRole}" > /var/log/directory-setup.log 2>&1
+                dir_rc=$?
+                echo "directory setup exited $dir_rc" >> /var/log/directory-setup.log
+                if [ "$dir_rc" -ne 0 ] && [ "${DirectoryRole}" = "client" ]; then
+                  echo "ERROR: LDAP client setup failed (exit $dir_rc). The node will still join and run jobs (Slurm uses numeric UIDs), but user name resolution / home-dir creation may be degraded on this node. Check this log and the directory server; not draining (avoids a replace loop)." >> /var/log/directory-setup.log
+                fi
```
KeitaW: ### Server-role (login-node) setup failures are under-surfaced — broaden the ERROR branch

Nice fix capturing the script's real exit code instead of `tee`'s `0` — that closes a real masking bug. One asymmetry: the explicit `ERROR:` line fires only for `DirectoryRole = "client"`. A **server**-role failure (the login node, where slapd itself lives) is the *more* serious case — if `setup_server()` returns non-zero, the directory is potentially unavailable cluster-wide — yet it produces only the generic `directory setup exited N` line with no `ERROR` marker for an operator (or a log scraper) to key on.

The no-drain rationale in the comment above is genuinely client-specific (a compute node degrades but still runs jobs by UID), so it's reasonable that the *guidance text* differs by role — but both roles deserve an equally loud `ERROR` on failure. I'd suggest splitting the branch so a server failure also logs clearly:

```suggestion
                if [ "$dir_rc" -ne 0 ]; then
                  if [ "${DirectoryRole}" = "client" ]; then
                    echo "ERROR: LDAP client setup failed (exit $dir_rc). The node will still join and run jobs (Slurm uses numeric UIDs), but user name resolution / home-dir creation may be degraded on this node. Check this log and the directory server; not draining (avoids a replace loop)." >> /var/log/directory-setup.log
                  else
                    echo "ERROR: LDAP server (login-node) setup failed (exit $dir_rc) — the directory service may be unavailable cluster-wide; compute nodes will not be able to resolve directory users. Check this log." >> /var/log/directory-setup.log
                  fi
                fi
```

Severity: should-fix (observability of the higher-impact failure path).

architectures/aws-pcs/README.md
```diff
@@ -153,7 +162,7 @@ idempotent — it no-ops on a pre-baked AMI.
 > **Production tip — pin the AMI.** CloudFormation re-resolves SSM `/latest/`
 > parameters on every stack update, so a later scale-out can drift onto a newer AMI.
 > Resolve once and pass the literal `ami-xxx` as `AmiId`. Details:
-> [OPERATIONS.md §4](./docs/OPERATIONS.md#4-ami-selection-amiid--pin-in-production).
+> [OPERATIONS.md §4](./docs/OPERATIONS.md#25-ami-selection-amiid--pin-in-production).
```
KeitaW: ### Stale "§4" link label after the OPERATIONS.md renumbering

The renumbering moved AMI selection from `## 4.` to `### 2.5` in OPERATIONS.md, and you correctly updated the link *target* to `#25-ami-selection-amiid--pin-in-production` (verified: it resolves against the new `### 2.5 AMI selection (\`AmiId\`) — pin in production` heading). The visible link *text* still reads "§4", so a reader sees "§4" but lands on §2.5 — cosmetic, but the kind of mismatch that erodes trust in the docs:

```suggestion
> [OPERATIONS.md §2.5](./docs/OPERATIONS.md#25-ami-selection-amiid--pin-in-production).
```

Severity: nit.

architectures/aws-pcs/docs/PARAMETERS.md
```diff
@@ -35,7 +35,7 @@ it directly.
...
+| `AmiId` | *(empty → SSM auto-resolve)* | AMI ID for every node group. **Empty (default) auto-resolves to the latest PCS-Ready Deep Learning AMI** (Ubuntu 24.04, x86_64) from SSM (`/aws/service/pcs/ami/dlami-base-ubuntu2404/x86_64/latest/ami-id`). For production, **pin to a specific `ami-xxx`** so a later scale-out cannot drift onto a newer base. Use a custom AMI built off the PCS-Ready DLAMI base (e.g. via [`pcs-ready-dlami-with-enroot-pyxis.yaml`](../README.md#85-pre-baking-enrootpyxis-into-a-custom-ami)) when you want Enroot/Pyxis pre-baked or other customizations. See [OPERATIONS.md §4](./OPERATIONS.md#25-ami-selection-amiid--pin-in-production) |
```
KeitaW: ### Same stale "§4" label in the `AmiId` parameter row

Identical issue on the `AmiId` row — target updated to `#25-...`, label still says "§4":

```suggestion
| `AmiId` | *(empty → SSM auto-resolve)* | AMI ID for every node group. **Empty (default) auto-resolves to the latest PCS-Ready Deep Learning AMI** (Ubuntu 24.04, x86_64) from SSM (`/aws/service/pcs/ami/dlami-base-ubuntu2404/x86_64/latest/ami-id`). For production, **pin to a specific `ami-xxx`** so a later scale-out cannot drift onto a newer base. Use a custom AMI built off the PCS-Ready DLAMI base (e.g. via [`pcs-ready-dlami-with-enroot-pyxis.yaml`](../README.md#85-pre-baking-enrootpyxis-into-a-custom-ami)) when you want Enroot/Pyxis pre-baked or other customizations. See [OPERATIONS.md §2.5](./OPERATIONS.md#25-ami-selection-amiid--pin-in-production) |
```

Severity: nit.

KeitaW (commented):
## Review Batch 2/3 — Documentation accuracy

KeitaW (commented):
## Review Batch 3/3 — Evaluation, Positives & Sources

### Things That Look Great

This is a well-engineered hardening PR — the kind where the inline comments accurately describe the tradeoffs rather than hand-waving them. I verified the substantive pieces rather than taking them on faith:

- **Password idempotency is correct, not just plausible.** `setup_server()` reads the existing SSM password *before* the `debconf-set-selections` block that feeds `slapd/password1|2`, and reassigns `LDAP_ADMIN_PASSWORD` — so the fresh `cn=config` (recreated by `apt-get install slapd` on the new instance, while the data MDB on `/home/ldap-db` persists) gets an `olcRootPW` matching the password the persistent data was created under. The trailing `put-parameter --overwrite` then correctly degenerates to a no-op.
- **Shell portability is right for *both* interpreters.** `setup-directory.sh` is `#!/bin/bash`, so `local`, `seq`, and the `for` loop are fine there; the `add-cng.yaml` `runcmd` runs under dash, and the new exit-capture deliberately avoids `PIPESTATUS`/bash arrays and uses only POSIX `[ ... ]` / `$?`. The comment explaining *why* (a bashism would die with "Bad substitution" before slurmd starts) is exactly the right thing to leave for the next contributor.
- **`Fn::Sub` discipline.** `$dir_rc` and `$?` are written without braces, so CloudFormation passes them through literally while substituting only `${DirectoryRole}` — the common footgun (shell `${VAR}` getting eaten by `Fn::Sub`) is avoided.
- **Discovery retry is sound.** 30 × 10s ≈ 5 min matches the comment, and the loop guards both empty and the `describe-instances` `None` sentinel.
- **The no-drain design call is well-justified and *tested*.** "A job runs by numeric UID even when the user is unresolvable on the node" is easy to assert and hard to verify — and the PR backs it with test case E5 plus a real-hardware observation. The replace-loop avoidance reasoning (drain → PCS replaces → same condition → drains again) is the right instinct.
- **Cross-reference integrity held up under the renumbering.** Sweeping the whole `architectures/aws-pcs` tree for `OPERATIONS.md#…` links, every *target* resolves (including the tricky em-dash/slash slugs `#25-…--pin-in-production` and `#32-…-stop--replacement`), with no dangling references to the removed `## 4.`/`## 5.`–`## 8.` numbers. Only two visible *labels* lagged (Batch 2).
- **The docs match the code.** ROADMAP correctly marks the password-regen issue as fixed and scopes the remaining stale-`ldap_uri` and TSDB-persistence gaps as future work; USER-MANAGEMENT's recovery `sed`/`sss_cache -E`/`restart sssd` sequence matches the `ldap_uri = ` line `setup_client` actually writes.
- **Independently verified end-to-end on live hardware.** I deployed this branch as a CPU PCS cluster in us-east-1 and exercised the hardening directly: a login-node replacement **preserved the SSM admin password** (SHA-256 identical before/after) with `setup_server` logging "Reusing existing admin password" and the admin bind still succeeding (**E3**); an already-running compute node kept its stale `ldap_uri`, a newly-added user failed to resolve there until the documented `sed` + `sss_cache -E` + `systemctl restart sssd` recovery, and a job running throughout was unaffected (**E4**); and a job submitted for a user unresolvable on the node still reached `COMPLETED` by numeric UID with the node never draining (**E5**). All three behaved exactly as the PR describes.

One pre-existing edge case worth a *separate* tracking issue (explicitly **not** a blocker, since it's untouched by this diff): the DB-relocation guard `[ ! -d "${LDAP_DB_DIR}" ]` in `setup_server` keys on directory *existence*, not contents — so a hypothetical "SSM password survives but `/home/ldap-db` was wiped" state would reuse the password against an empty directory. That requires manual destruction of persistent `/home`, not a normal replacement.

### Sources
- Slurm upgrade policy (AMI single-version rationale): https://slurm.schedmd.com/upgrades.html
- Verified live, 2026-06-20: `setup-directory.sh` shebang is `#!/bin/bash`; all `OPERATIONS.md#…` / `USER-MANAGEMENT.md#…` anchor targets resolve; no stale section-number cross-references after the §4-removal renumbering.
- Verified live, 2026-06-21: end-to-end deploy of this branch in us-east-1 (multi-user LDAP + Slurm accounting + login-node replacement E3/E4/E5), per the bullet above.

KeitaW (approved):
Few minor comments.

DaisukeMiyamoto:
Thanks for the thorough review, @KeitaW — all three addressed in 6eb08259:

1. **Server-role failure under-surfaced (should-fix):** Agreed, the higher-impact path was the one missing a loud signal. Split the branch so a `server` (login-node) failure now logs its own `ERROR:` line noting the directory may be unavailable cluster-wide; the client branch keeps its degradation/no-drain wording. Verified the rendered `runcmd` still parses under `/bin/sh` (dash).
2. **README "§4" label (nit):** Updated the visible label to **§2.5** to match the already-corrected `#25-...` target.
3. **PARAMETERS "§4" label (nit):** Same fix on the `AmiId` row.

`tests/lint-docs.sh` PASS and `add-cng.yaml` still passes `validate-template`.

DaisukeMiyamoto:
Quick follow-up: I exercised the failure branch on a live login node (rendered `runcmd` block run under `/bin/sh`, pointed at a throwaway non-zero command so the real slapd/SSSD were untouched). Both roles now log as intended:

```
===== CLIENT role failure =====
directory setup exited 7
ERROR: LDAP client setup failed (exit 7). The node will still join and run jobs (Slurm uses numeric UIDs), but user name resolution / home-dir creation may be degraded on this node. Check this log and the directory server; not draining (avoids a replace loop).

===== SERVER role failure =====
directory setup exited 7
ERROR: LDAP server (login-node) setup failed (exit 7) — the directory service may be unavailable cluster-wide; compute nodes will not be able to resolve directory users. Check this log.
```

slapd/sssd stayed `active` throughout (no side effects — the change only adds a log line, never drains).

PR #1144, opened 2026-06-21
feat(megatron-bridge): Qwen3-235B 3-way MoE dispatcher comparison (NCCL vs DeepEP+UCCL vs DeepEP+NVSHMEM)

KeitaW:
## Summary

Adds a **Qwen3-235B-A22B (128-expert MoE) three-way expert-parallel all-to-all comparison**
to the `megatron-bridge` test case: **NCCL all-to-all** vs **DeepEP+UCCL** vs **DeepEP+NVSHMEM**,
measured as end-to-end **training-step throughput** on `p6-b300` (B300), swept across
**EP16 and EP32**. The DeepEP+NVSHMEM arm — NVIDIA DeepEP carrying its all-to-all over
**NVSHMEM-libfabric on EFA** — is new to this library and is wired in without touching
Megatron-Core.

## Motivation

The `megatron-bridge` library already compares the Megatron-Core MoE token dispatcher between
**NCCL all-to-all** and **DeepEP via UCCL** (UCCL's EFA-native `deep_ep` shadow). Two gaps
motivated this PR:

1. **"NVIDIA DeepEP can't run on EFA" is now obsolete.** That is true only of DeepEP's stock
   IB/IBGDA build. The repo's own
   [`micro-benchmarks/expert-parallelism/deepep-benchmark`](../../micro-benchmarks/expert-parallelism/deepep-benchmark)
   already patches DeepEP to run **NVSHMEM over libfabric (host-proxy, IBGDA off)** on EFA.
   This PR brings that build in as a *third* Megatron-Core dispatcher backend so DeepEP+NVSHMEM
   and DeepEP+UCCL can be compared head-to-head against the NCCL baseline on the same model.
2. **The DSV3 A/B needed 32 nodes (EP32) to fit 671B.** To make a meaningful EP sweep runnable
   at a smaller, commonly-available node count, this PR uses **Qwen3-235B-A22B** (235B total /
   22B active, 128 experts, top-8, 94 layers, hidden 4096, no shared expert), which fits a
   handful of B300 nodes at **both EP16 and EP32** — so the comparison sweeps the variable the
   all-to-all actually turns on, the **EP degree**, at a realistic operating point (EP32 is the
   point cited for Qwen3-235B in the DSV3 reference table).

## Approach

The third arm is the **same `MOE_DISPATCHER=deepep` code path** — only the importable `deep_ep`
module differs, so the arm is selected by *which image* runs. Provider is chosen with a new
**`EP_BACKEND={uccl,nvshmem}` build arg on the single shared Dockerfile** (the library's README
forbids a second Dockerfile; a dual-install would collide two `deep_ep` packages):

| Arm | `MOE_DISPATCHER` | Image (`EP_BACKEND`) | all-to-all transport |
|-----|------------------|----------------------|----------------------|
| NCCL all-to-all (baseline) | `alltoall` | uccl | NCCL all-to-all over EFA |
| DeepEP + UCCL | `deepep` | uccl | UCCL EFA-native `deep_ep` |
| DeepEP + NVSHMEM | `deepep` | **nvshmem** | NVIDIA DeepEP over NVSHMEM-libfabric/EFA |

No Megatron-Core patch; the benchmark entrypoint is identical across all three arms.

## What changed

**Shared environment (`megatron-bridge/`)**
- `Dockerfile`: `EP_BACKEND` arg. The `nvshmem` branch builds GDRCopy + NVSHMEM `v3.7.0-0`
  (libfabric, IBGDA off) + EFA-patched NVIDIA DeepEP `567632d` into `/opt/venv`, then cleans the
  build/source trees. Build-time verify uses `find_spec` for the nvshmem build (DeepEP's
  extension links `libcuda`, absent during `docker build`) and the real `import` for uccl.
- `1.build-and-push.sh`: builds/pushes both tags (`…-uccl-0dc87eb`, `…-deepep-nvshmem-567632d-cu13`).
- `deepep/setup_deepep_efa.sh`: vendored verbatim from the deepep-benchmark (Docker build-context
  constraint) with a sync note in `deepep/README.md`.
- `run-ab-rawpods.sh`: `MODEL=qwen3-235b`, `ARM_LABEL`/`EP_BACKEND` tagging (distinct run dirs for
  the two deepep transports), and `HF_HOME`/`HF_HUB_OFFLINE`/`RECOMPUTE`/`NUM_LAYERS` passthrough.
- `bench/parse-runs.py`: NVSHMEM-active log signal + `backend`/`arm_label`/`ep`/`nvshmem_ok`
  columns; `seq` read from `env.txt` for tok/s; aligned stdout table.
- `2.sanity-singlenode.sh`: backend-aware Gate 2; fixes a `fi_info | head` SIGPIPE-under-`pipefail`
  abort on 16-NIC nodes.
- `README.md`: corrects the "DeepEP can't run on EFA" claim; adds the Qwen3-235B model row.

**New model (`megatron-bridge/qwen3-235b/`)**
- `benchmarks/bench_qwen3_pretrain.py`: mirrors the validated dsv3 bench against the shipped
  `qwen3_235b_a22b_pretrain_config` recipe (mock data + random init — we measure step time). EP
  sweep, A/B validity guard, `RECOMPUTE` knob, `overlap=on` forced off at PP1, eval + checkpoint
  saving disabled.
- `benchmarks/run-qwen3-campaign.sh`: drives 3 arms × EP{16,32} × cells via `run-ab-rawpods.sh`,
  with an EP-divides-WORLD / integer-DP guard, per-run EFA gate, and end-of-campaign parse.
- `benchmarks/RESULTS.md`, `README.md`: methodology, measured numbers, and caveats.

## Validation

- **Both images build and push** to ECR; single-node gates 1–3 pass on the nvshmem image (EFA
  present, `deep_ep.Buffer` importable from `/opt/venv`, MCore flex/deepep config fields). Gate 5
  is the pre-existing known-stale standalone-dispatcher gate (the real `pretrain()` path is the
  authoritative dispatch check, per the library README).
- **Work-equivalence (no token dropping)** on a 2-node smoke (`NUM_LAYERS=8`, `LOSS_PROBE=1`):
  DeepEP+NVSHMEM iteration-1 mean loss **12.701515** vs NCCL all-to-all **12.702237** (identical
  `num_tokens=2048`; rel. diff 5.7e-5 = bf16 round-off). The NVSHMEM arm's IBGDA→host-proxy +
  put_signal patches dispatch/combine the same tokens to numerically-equivalent output.
- **Full campaign**: 12 runs, every run EFA-active on all ranks (`efa_ok`), 0 stalls, 8 steady
  iters; per-arm transport confirmed (`uccl_ok` / `nvshmem_ok`).

## Results (two fabrics, mb=4, overlap=off; Δ vs NCCL all-to-all, − = DeepEP faster)

**B300 / Qwen3-235B** — 8× p6-b300, TP8, selective recompute, EP16=PP4/DP2, EP32=PP2/DP4:

| EP | NCCL all-to-all | DeepEP+UCCL | DeepEP+NVSHMEM |
|---:|----------------:|------------:|---------------:|
| 16 | 12.21 s | +6.9% | **−5.2%** |
| 32 | 15.76 s | +21.5% | +34.0% |

**H100 / Qwen3-30B** — 8× p5.48xlarge, TP8, recompute off, same EP layout:

| EP | NCCL all-to-all | DeepEP+UCCL | DeepEP+NVSHMEM |
|---:|----------------:|------------:|---------------:|
| 16 | 8.60 s | +72% | +6.6% |
| 32 | 9.24 s | +187% | +86% |

Full tables (mb=1, overlap=on, TFLOP/s, tok/s, validity): `qwen3-235b/benchmarks/RESULTS.md`
(B300) and `qwen3-30b/benchmarks/RESULTS.md` (H100). The 4-node B300 run (capacity-constrained,
`RECOMPUTE=full`) is retained as a secondary reference in the B300 RESULTS.

## How to read this — the recommendation is **scale-dependent**

At the **node counts tested here (4–8 nodes)**, **NCCL all-to-all is the right default on EFA**:
it wins nearly every cell, the lone exception being B300 EP16 mb4 where DeepEP+NVSHMEM edges it
(−5.2%). DeepEP+NVSHMEM is the stronger of the two DeepEP backends in most cells (UCCL edges it at B300 EP32 mb4).

But this is **not** "EFA kills DeepEP" — and the small-scale result does **not** generalize:

- **It's an EFA/host-proxy constraint, not DeepEP being slow per se.** Without IBGDA, DeepEP on
  EFA is stuck on the NVSHMEM host-proxy path and can't use its GPU-initiated-RDMA advantage.
- **DeepEP's edge grows with scale (EP span / node count).** DeepEP/UCCL does *sparse* dispatch
  (tokens go only to the experts they route to), so its advantage grows with inter-node
  all-to-all span and eventually overtakes the host-proxy penalty. That trend is already visible
  here (the DeepEP gap shrinks markedly EP32→EP16), and at larger scale it flips: the repo's
  **Kimi-K2 32-node / 256-GPU / 384-expert** result shows **UCCL −34% vs NCCL at mb≥4**
  ([`kimi-k2/benchmarks/RESULTS.md`](kimi-k2/benchmarks/RESULTS.md)) — and −21% at 16 nodes,
  i.e. the win itself scales with node count.
- **Fabric is second-order.** DeepEP looked worst on H100 (p5.48xlarge, lower EFA bandwidth
  exposes the host-proxy cost); it should look better on p5en.48xlarge / newer. But it still
  lost at the B300 8-node/EP32 point, so the effect is **scale-primary, fabric-secondary**.

**Customer-facing takeaway:** on EFA, default to **NCCL all-to-all up to ~8 nodes / EP32**;
consider DeepEP (UCCL today; NVSHMEM pending large-scale validation) at **large EP span /
≳16–32 nodes**, where sparse dispatch pays for the host-proxy overhead.

## Caveats

- **Balanced routing only.** Runs use `moe_router_force_load_balancing` (mock data, random
  init); **skewed / real expert routing is not yet tested** and could shift the dispatcher
  balance.
- **n=1 per cell.** Large deltas are robust; the small B300 EP16 gaps (±5–7%) are within
  run-to-run noise (n≥3 is the recommended follow-up).
- **H100/30B is comm-bound** (hidden 2048, compute-light) so its % deltas are amplified — read
  it as the small-message worst case for DeepEP, not end-to-end MFU.
- A few cells were lost to shared-cluster node preemption (documented per RESULTS). The harness
  now skips completed cells on **all** ranks (fixes a same-`CAMPAIGN_ID` re-run log clobber).

## Follow-ups

- **NVSHMEM+DeepEP at Kimi-K2 32-node scale** — the highest-value missing cell; standalone EP
  micro-benchmarks suggest NVSHMEM ≥ UCCL in throughput mode, so it's plausibly the best
  large-scale option. Not yet run.
- Decode/inference NCCL/UCCL/NVSHMEM comparison on p5en.48xlarge (complements these training
  numbers).
- n≥3 repeats to put error bars on the close B300 EP16 deltas.

## Reviewers / acks

Thanks to **Henan Wang** (DeepEP / EFA performance) for the scale-dependence framing and the
host-proxy-vs-IBGDA context that shaped the recommendation above.

## Risk / blast radius

Additive: a new model subdir + a build-arg-gated branch on the shared image. The default
`EP_BACKEND=uccl` build is byte-for-byte the existing behavior; existing dsv3/kimi-k2 cases are
untouched except the shared `run-ab-rawpods.sh`/`parse-runs.py`, whose changes are backward
compatible (new optional env + new CSV columns). No secrets, no new network surface; versions
pinned (NVSHMEM `v3.7.0-0`, DeepEP `567632d`, GDRCopy `v2.5.2`).

### DIFF
--- a/3.test_cases/megatron/megatron-bridge/Dockerfile
+++ b/3.test_cases/megatron/megatron-bridge/Dockerfile
@@ -46,6 +46,25 @@ ARG TORCH_CUDA_ARCH_LIST="10.0a+PTX;10.3a+PTX"
 # nanobind is a UCCL build-time dependency. Pinned per the repo "no latest" rule.
 ARG NANOBIND_VERSION=2.4.0
 
+# EP backend selector — which library provides the top-level `deep_ep` module that
+# Megatron-Core's flex/deepep dispatcher (`moe_flex_dispatcher_backend="deepep"`)
+# resolves at `import deep_ep`. ONE Dockerfile, TWO image tags:
+#   uccl    (default) -> UCCL EFA-native deep_ep shadow (the existing build)
+#   nvshmem           -> NVIDIA DeepEP, EFA-patched to run NVSHMEM over libfabric
+#                        (host-proxy, no IBGDA — see micro-benchmarks/expert-
+#                         parallelism/deepep-benchmark/setup_deepep_efa.sh).
+# Both arms then run the SAME `MOE_DISPATCHER=deepep` code path; only the image
+# (hence which `deep_ep` is importable) differs. This is what lets the dsv3/qwen3
+# benchmarks compare DeepEP+UCCL vs DeepEP+NVSHMEM with no Megatron-Core changes.
+ARG EP_BACKEND=uccl
+# nvshmem-backend pins (used only when EP_BACKEND=nvshmem). DeepEP commit + NVSHMEM
+# tag match micro-benchmarks/.../deepep-benchmark (proven on p6-b300 / sm_103).
+# DEEPEP_GPU_ARCH=100 (sm_100) is the Blackwell cubin the deepep-benchmark image
+# was built with and ran on these sm_103 B300 nodes.
+ARG NVSHMEM_VERSION=v3.7.0-0
+ARG DEEPEP_COMMIT=567632d
+ARG DEEPEP_GPU_ARCH=100
+
 ARG OPEN_MPI_PATH=/opt/amazon/openmpi
 
 ######################
@@ -114,13 +133,15 @@ ENV LD_LIBRARY_PATH=/usr/local/cuda/extras/CUPTI/lib64:/opt/amazon/openmpi/lib:/
     PATH=/opt/venv/bin:/opt/amazon/openmpi/bin:/opt/amazon/efa/bin:$PATH
 
 #####################################################
-## UCCL core + UCCL-EP + deep_ep shadow wrapper
+## deep_ep provider — UCCL (default) OR NVIDIA DeepEP + NVSHMEM-over-EFA
 ##
-## UCCL provides the EFA-native expert-parallel all-to-all. Its `deep_ep_wrapper`
-## installs a TOP-LEVEL python module named `deep_ep` that shadows NVIDIA DeepEP,
-## so Megatron-Core's flex/deepep dispatcher (`moe_flex_dispatcher_backend="deepep"`)
-## resolves `import deep_ep` to the UCCL/EFA implementation.
+## Exactly one of the two RUN blocks below executes, gated on EP_BACKEND, and each
+## installs a TOP-LEVEL python `deep_ep` module into /opt/venv that shadows the
+## NVIDIA DeepEP the base ships in dist-packages (IB/NVSHMEM-bound). Megatron-Core's
+## flex/deepep dispatcher (`moe_flex_dispatcher_backend="deepep"`) then resolves
+## `import deep_ep` to whichever was installed.
 #####################################################
+# --- EP_BACKEND=uccl: UCCL core + UCCL-EP + deep_ep shadow wrapper ---
 # Use `python3 -m pip` for ALL of UCCL so packages land in the SAME env as python3
 # (/opt/venv, first on PATH, where torch + Megatron live). A bare `pip` targets
 # /usr/local; combined with the `ep` extension that setup.py installs into /opt/venv,
@@ -131,7 +152,7 @@ ENV LD_LIBRARY_PATH=/usr/local/cuda/extras/CUPTI/lib64:/opt/amazon/openmpi/lib:/
 #   nanobind (build dep) -> clone+checkout pinned SHA + submodules -> UCCL core ->
 #   UCCL-EP kernels (PER_EXPERT_BATCHING + Blackwell arch list; the ";" in
 #   TORCH_CUDA_ARCH_LIST must stay quoted) -> deep_ep shadow wrapper.
-RUN set -eux; \
+RUN if [ "${EP_BACKEND}" = "uccl" ]; then set -eux; \
     python3 -m pip install nanobind==${NANOBIND_VERSION}; \
     git clone https://github.com/uccl-project/uccl.git /opt/uccl; \
     cd /opt/uccl; \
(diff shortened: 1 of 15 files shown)


3.test_cases/megatron/megatron-bridge/qwen3-235b/benchmarks/RESULTS.md
```diff
@@ -0,0 +1,156 @@
...
+| mb | NCCL all-to-all | DeepEP+UCCL | DeepEP+NVSHMEM | UCCL Δ | NVSHMEM Δ |
+|---:|----------------:|------------:|---------------:|-------:|----------:|
+| 4 | **15.91 s · 152.4 · 65.9k** | 18.43 s · 134.5 · 56.9k | 20.41 s · 120.9 · 51.4k | +15.9% | +28.3% |
+| 1 | **33.41 s · 72.6 · 31.4k** | 46.90 s · 52.4 · 22.4k | 45.39 s · 54.2 · 23.1k | +40.4% | +35.9% |
+
+### EP32 @ PP2, overlap=on (1F1B, deployment regime) — mean iter s
+
+| mb | NCCL all-to-all | DeepEP+UCCL | DeepEP+NVSHMEM | UCCL Δ | NVSHMEM Δ |
+|---:|----------------:|------------:|---------------:|-------:|----------:|
+| 4 | **11.01 s · 225.0 · 95.3k** | 17.08 s · 150.6 · 61.4k | 17.08 s · 150.6 · 61.4k | +55.1% | +55.1% |
```
whn09: **Likely duplicated data — please verify before this row is read as a result.** DeepEP+UCCL and DeepEP+NVSHMEM here are identical to the digit on all three metrics (`17.08 s · 150.6 · 61.4k`, both `+55.1%`). Two independent transports producing byte-identical iter time / TFLOP / tok/s is essentially impossible from real measurement. Combined with the note just below (these EP32 overlap=on cells were preempted and re-run, and "logs were cosmetically overwritten by the same-campaign re-run"), this looks like one run's data landed in both columns — i.e. the exact same-`CAMPAIGN_ID` log-clobber class of bug this PR fixes in `run-ab-rawpods.sh`. Could you confirm each column's `index.csv`/`env.txt` traces to a distinct run dir? If they can't be separated, I'd mark this row unverified or re-measure.

3.test_cases/megatron/megatron-bridge/qwen3-235b/benchmarks/RESULTS.md
```diff
@@ -0,0 +1,156 @@
...
+
+| mb · overlap | NCCL all-to-all | DeepEP+UCCL | DeepEP+NVSHMEM | UCCL Δ | NVSHMEM Δ |
+|---|----------------:|------------:|---------------:|-------:|----------:|
+| 4 · off | 12.21 s · 198.6 · 85.9k | 13.06 s · 185.8 · 80.3k | **11.57 s · 209.6 · 90.6k** | +6.9% | **−5.2%** |
+| 1 · off | 34.17 s · 71.0 · 30.7k | 40.00 s · 60.7 · 26.2k | _preempted_ | +17.1% | — |
+| 4 · on  | **10.53 s · 235.0 · 99.5k** | 13.24 s · 187.1 · 79.2k | 11.32 s · 218.7 · 92.6k | +25.7% | +7.5% |
+
+> Two cells lost to mid-run node preemption (rendezvous < 8/8): `nvshmem-ep16-mb1-off` and (on
+> the first pass) the two EP32 overlap=on DeepEP cells, which were re-run successfully. The
+> EP32 overlap=off numbers above are from the first pass (their logs were cosmetically
```
whn09: This note says the **primary** EP32 overlap=off numbers (the headline table above) come from a first pass whose logs were overwritten by the same-campaign re-run. That means the headline deltas aren't reproducible from the logs the repo retains. Suggest explicitly flagging which cells are no longer log-reproducible (a short superscript/marker in the table), so a reader doesn't assume the whole primary table is traceable. Doesn't change the conclusion, just the provenance honesty.

3.test_cases/megatron/megatron-bridge/Dockerfile
```diff
@@ -131,29 +152,68 @@ ENV LD_LIBRARY_PATH=/usr/local/cuda/extras/CUPTI/lib64:/opt/amazon/openmpi/lib:/
...
+      --nvshmem-src /opt/nvshmem_src; \
+    rm -rf /opt/nvshmem_src/build /opt/deepep-efa-src; \
+    else echo "EP_BACKEND=${EP_BACKEND}: skipping NVSHMEM/DeepEP build"; fi
+
+# NVSHMEM runtime env. Harmless for the uccl image (the dirs are absent there).
+# Must PRECEDE the verify RUN so the nvshmem build's `import deep_ep` can dlopen
+# libnvshmem; LD_LIBRARY_PATH puts the from-source NVSHMEM ahead of any torch-bundled one.
+ENV NVSHMEM_DIR=/opt/nvshmem_src/install \
+    NVSHMEM_HOME=/opt/nvshmem_src/install \
+    NVSHMEM_REMOTE_TRANSPORT=libfabric \
```
whn09, line 203: Minor: `ENV` can't be conditioned on `EP_BACKEND`, so these `NVSHMEM_*` vars are baked into the **uccl** image too. You've noted it's harmless (dirs absent), and that's true today. The latent risk is if a uccl image ever ends up with any NVSHMEM bits present, these would silently point it at a libfabric transport. Fine to keep; a one-line caveat that this is intentionally unconditional would help future readers.

3.test_cases/megatron/megatron-bridge/deepep/README.md
```diff
@@ -0,0 +1,23 @@
...
+built with `--build-arg EP_BACKEND=nvshmem` — Docker cannot `COPY` files from outside
+the build context, and the build context is this test-case directory.
+
+It builds **NVIDIA DeepEP over NVSHMEM-libfabric for AWS EFA** (host-proxy, IBGDA off)
+via four source patches: a combined put+signal, an IBGDA→host-proxy shim, fake IBGDA
+device state, and `NVSHMEM_MAX_TEAMS` 7→8. See that benchmark's
+[README](../../../../micro-benchmarks/expert-parallelism/deepep-benchmark/README.md) for
+the full rationale.
+
+**Keep in sync with the canonical copy.** If the benchmark script changes, re-copy:
```
whn09, line 18: This 469-line script is vendored as a verbatim copy kept in sync by hand. Drift between this copy and the canonical `micro-benchmarks/.../setup_deepep_efa.sh` would be silent and could mean the benchmark builds something different from what's documented. Consider a tiny CI check that `diff`s the two files and fails on mismatch — cheap insurance against drift.

3.test_cases/megatron/megatron-bridge/run-ab-rawpods.sh
```diff
@@ -89,14 +118,17 @@ EOF
...
             echo overlap=${MOE_A2A_OVERLAP} force_balance=${MOE_FORCE_BALANCE} loss_probe=${LOSS_PROBE} ;
             echo image=${IMG} ; echo bench_py=${BENCH_PY} ; echo git_rev=${GIT_REV} ; echo started=\$(date -u +%FT%TZ) ; } > ${RUN_DIR}/env.txt ;"
 
 launch_pod() {
   local R="$1"
-  local PREAMBLE="mkdir -p ${LOGDIR} ;"
+  # ALL ranks must skip a completed cell, not just rank-0. If only rank-0 REFUSE-exits, ranks
+  # 1..N-1 still start torchrun, fail rendezvous (no rank-0), and OVERWRITE their rank logs —
+  # corrupting a previously-good run when a campaign is re-run with the same CAMPAIGN_ID.
+  local PREAMBLE="if [ -f ${RUN_DIR}/STATUS ]; then echo 'skip: completed run' ; exit 0 ; fi ; mkdir -p ${LOGDIR} ;"
```
whn09, line 134: Nit: the skip-guard keys on STATUS *existing*, and the NVSHMEM arm writes STATUS even though it exits 1 at NVSHMEM finalize (per RESULTS, validity is judged by `efa_ok` + `n_steady==8`, not exit code). So a same-`CAMPAIGN_ID` re-run will skip a NVSHMEM cell that exited non-zero — which is intended here, but reads like "skipping a failed run as if complete." A one-line comment noting STATUS-exists is deliberately the skip key (not exit==0) would prevent a future misread.

whn09 (approved):
Approving — clean single-Dockerfile + EP_BACKEND two-tag design, no Megatron-Core patching, and the loss-based work-equivalence check + all-rank EFA gate make the numbers trustworthy. Left a few inline comments; the only one worth a quick look before merge is the EP32 overlap=on row where the two DeepEP backends report digit-for-digit identical numbers (looks like one run in both columns).

KeitaW:
Thanks @whn09 — addressed all five.

**1. EP32 overlap=on "duplicated" row — your instinct the row was off was right, but it wasn't one run in both columns.** I traced both columns to disk: two distinct run dirs, distinct images (`uccl-0dc87eb` vs `deepep-nvshmem-567632d`), finish times 30 min apart, and distinct per-iter loss curves (UCCL mean 17.0770 s / median 18.29 / 8 steady; NVSHMEM 17.0806 / 17.78 / 7 steady). Two real runs that collapsed at 3-sig-fig display (`17.08·150.6·61.4k`). Rather than just add decimals I re-measured (see #2) — the clean overlap=on row now separates: **UCCL +68.0%, NVSHMEM +63.4%** (NVSHMEM actually faster).

**2. Primary overlap=off not log-reproducible — confirmed, and it was the real issue.** The per-iter timing only prints on the last-PP-stage rank (rank-7 here); the same-`CAMPAIGN_ID` re-run clobbered exactly those logs for the 6 EP32 overlap=off cells (rank-0 survived but carries no timing → the parser saw `nan`). So I re-ran all 9 EP32 cells clean in one fresh campaign (`qwen3-ep32-remeasure-20260624`, git `b73a2d7`, 16 iters / 12 steady, all-rank skip-guard active). Everything is now reproducible from retained rank logs (`efa_ok` 8/8, transport confirmed, iter-1 `lm loss` matches across arms to ≤2e-6). The conclusion holds — NCCL fastest at EP32 — and the re-measured numbers land within 1–4% of the earlier pass. RESULTS.md primary section + provenance note updated; the old "first-pass capture" caveat is gone.

**3. NVSHMEM_* ENV unconditional** — added a comment that it's intentionally unconditional (`ENV` can't gate on the `EP_BACKEND` ARG) and named the latent risk.

**4. Vendored `setup_deepep_efa.sh` drift** — added `.github/workflows/deepep-vendor-sync.yml`, a path-filtered PR gate that diffs the vendored copy vs the canonical source and fails on mismatch (currently identical). README points at it.

**5. STATUS skip-key** — added a comment that STATUS-*exists* (not `exit==0`) is the deliberate skip key, so NVSHMEM cells that exit 1 at teardown are treated as complete.

Thanks again — #1/#2 materially improved the provenance.

PR #1145, opened 2026-06-21
Update DeepEP micro benchmark

erezzarum:
## Changes

* Remove setup script and use a DeepEP AWS EFA patch to patch DeepEP
* Use upstream NVSHMEM 3.7
* Update k8s examples for multi-node to bound to a single AZ
* Add `NVSHMEM_NETDEVS_POLICY=EXTERNAL_SHARING_PCIE_SWITCH_NIC_EXCLUSIVE` to k8s and slurm examples.
* Cleanup Dockerfile.

## Test Plan

**Environment:**
- AWS Service: EKS
- Instance type: p5.48xlarge, p5en.48xlarge, p6-b200.48xlarge
- Number of nodes: 2

**Test commands:**
```bash
export IMAGE_URI=<HIDEEN>

export INSTANCE_TYPE=p5.48xlarge
export GPU_PER_NODE=32
export EFA_PER_NODE=32
export NUM_NODES=2
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE' < test-intranode.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-low-latency.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-internode.yaml | kubectl apply -f -

export INSTANCE_TYPE=p5en.48xlarge
export GPU_PER_NODE=16
export EFA_PER_NODE=16
export NUM_NODES=2
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE' < test-intranode.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-low-latency.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-internode.yaml | kubectl apply -f -

export INSTANCE_TYPE=p6-b200.48xlarge
export GPU_PER_NODE=8
export EFA_PER_NODE=8
export NUM_NODES=2
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE' < test-intranode.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-low-latency.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-internode.yaml | kubectl apply -f -
```

## Directory Structure

```
3.test_cases/
```

## Checklist


micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
```diff
@@ -118,56 +192,24 @@ ENV PMIX_MCA_gds=hash
...
+    export DISABLE_AGGRESSIVE_PTX_INSTRS && \
+    git clone https://github.com/deepseek-ai/DeepEP.git ${DEEPEP_PREFIX} && \
+    cd ${DEEPEP_PREFIX} && \
+    git checkout ${DEEPEP_COMMIT} && \
+    patch -p1 < /tmp/deepep_aws/deepep_aws_efa.patch && \
+    CUDA_VERSION_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') && \
+    if [ "${CUDA_VERSION_MAJOR}" -ge 13 ]; then \
+        sed -i "s|f'{nvshmem_dir}/include']|f'{nvshmem_dir}/include', '${CUDA_HOME}/include/cccl']|" "setup.py"; \
+    fi && \
+    NVSHMEM_DIR="${NVSHMEM_PREFIX}" pip3 install -vv --no-build-isolation .
```
KeitaW, line 186: ### Arch-selection knobs (`TORCH_CUDA_ARCH_LIST`, `NVCC_GENCODE`, `GPU_ARCH`) are wired through the README but not consumed in the active build path

The README's single-arch override tells users to pass `--build-arg GPU_ARCH=90 --build-arg TORCH_CUDA_ARCH_LIST=9.0 --build-arg NVCC_GENCODE=…` for "a smaller single-arch image", and all three are declared as `ARG`s and threaded through the `docker build` command. But in the active build path none of them reaches a compiler:

- `TORCH_CUDA_ARCH_LIST` is referenced only inside the `if [ "${TORCH_CUDA_ARCH_LIST}" = "9.0" ]` test that toggles `DISABLE_AGGRESSIVE_PTX_INSTRS`. It is never exported into the environment of `pip3 install .`, so torch's `CUDAExtension` never sees it. (The deleted `setup_deepep_efa.sh` *did* export it — so this is a regression introduced by the rewrite.)
- `NVCC_GENCODE` has no active consumer: the NCCL and nccl-tests builds dropped their explicit `NVCC_GENCODE=…` in this PR, and the only other reference is the commented-out from-source NVSHMEM block.
- `GPU_ARCH` is consumed only by `-DCMAKE_CUDA_ARCHITECTURES`, which is inside that same commented-out `RUN`.

So a user following the README to build a Hopper-only image silently gets whatever torch's default gencode produces — and the Hopper-only path sets `DISABLE_AGGRESSIVE_PTX_INSTRS=0` while torch may still emit `sm_100`, the mismatch the override was meant to avoid. The default both-arch build likely still works (torch falls back to its own arch list), so probably not a build-breaker — but the documented override is non-functional. Minimal fix — export it into the build env so torch honors it:

```suggestion
    TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST}" NVSHMEM_DIR="${NVSHMEM_PREFIX}" pip3 install -vv --no-build-isolation .
```

Alternatively, if the intent is always a fat binary, drop `GPU_ARCH`/`NVCC_GENCODE`/`TORCH_CUDA_ARCH_LIST` and the README override block rather than ship knobs that no-op. (The README `NVCC_GENCODE` inline comment `# … use "90" or "100"` is also a copy-paste from `GPU_ARCH` — its values are gencode strings, not `90`/`100`.)

micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
```diff
@@ -118,56 +192,24 @@ ENV PMIX_MCA_gds=hash
...
-    --cuda-home /usr/local/cuda \
-    --libfabric-home /opt/amazon/efa \
-    --gdrcopy-home /opt/gdrcopy \
-    --gpu-arch "${GPU_ARCH}" \
-    --venv /opt/deepep \
-    --deepep-src /DeepEP \
-    --nvshmem-src ${NVSHMEM_SRC}
+ARG DEEPEP_PREFIX="/DeepEP"
+
+RUN --mount=type=bind,source=deepep_aws_efa.patch,target=/tmp/deepep_aws/deepep_aws_efa.patch \
```
KeitaW, line 174: ### `RUN --mount=type=bind` and `TARGETARCH` require BuildKit, but the README build command doesn't enable it

This step now uses `RUN --mount=type=bind,source=deepep_aws_efa.patch,…` (replacing the old `COPY`), and the NVSHMEM tarball selection branches on `TARGETARCH`. Both are BuildKit-only: `TARGETARCH` is auto-populated only under BuildKit/buildx, and `RUN --mount` is a hard parse error on the legacy builder. The README's `docker build --progress=plain …` block sets no `DOCKER_BUILDKIT=1` and uses no `docker buildx`, so on a host where the legacy builder is the default the build fails outright on the `--mount` syntax. Modern Docker defaults to BuildKit so most users are fine, but the canonical build instructions shouldn't depend on that — prefixing the README build command with `DOCKER_BUILDKIT=1` (or documenting a buildx requirement) closes it.

micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
```diff
@@ -118,56 +192,24 @@ ENV PMIX_MCA_gds=hash
...
+    if [ "${TORCH_CUDA_ARCH_LIST}" = "9.0" ]; then DISABLE_AGGRESSIVE_PTX_INSTRS=0; else DISABLE_AGGRESSIVE_PTX_INSTRS=1; fi && \
+    export DISABLE_AGGRESSIVE_PTX_INSTRS && \
+    git clone https://github.com/deepseek-ai/DeepEP.git ${DEEPEP_PREFIX} && \
+    cd ${DEEPEP_PREFIX} && \
+    git checkout ${DEEPEP_COMMIT} && \
+    patch -p1 < /tmp/deepep_aws/deepep_aws_efa.patch && \
```
KeitaW: ### `patch -p1` applies with fuzz; `git apply` fails loudly if the pinned upstream ever drifts

Applying the vendored diff against a pinned commit is the right call, and `567632d` is pinned so it can't rot today. One hardening: `patch -p1` applies with a default fuzz factor and can land hunks at shifted offsets, succeeding with only a warning. `git apply` is exact — no fuzz — so the day someone bumps `DEEPEP_COMMIT` without regenerating the patch, the build fails cleanly instead of producing a subtly mis-patched binary:

```suggestion
    git apply /tmp/deepep_aws/deepep_aws_efa.patch && \
```

Worth a one-line note near `DEEPEP_COMMIT` that the patch is coupled to that commit and must be regenerated on a bump.

KeitaW (commented):
## Review Batch 1/3 — Build reproducibility & arch selection

Thanks for this — it's a real improvement over the previous bring-up: the fragile 469-line `setup_deepep_efa.sh` is gone, NVSHMEM now comes from a pinned NVIDIA redist binary, and the DeepEP changes land as a single reviewable patch against a **pinned** commit. I verified the three load-bearing build inputs resolve (NVSHMEM 3.7.0 redist tarball → HTTP 200; `torch==2.11.0+cu130` wheel exists; DeepEP `567632d` → full SHA). Findings: 2 should-fix, the rest nits, plus what looks great in the final batch. I treated the 703-line `deepep_aws_efa.patch` as upstream-derived data (not line-reviewed); the findings assume it applies cleanly against the pinned `567632d`.

**Validation Evidence — the Test Plan is empty (should fix).** The Test Plan (environment, commands, results) is all placeholders, on a change that swaps the patch mechanism, moves NVSHMEM from source-build to prebuilt binary, reworks arch handling, and changes the EFA-installer invocation. Given the GPU-arch surface, the key thing to state: which instance types and node counts did this build *and run* on — Hopper (`p5`/`p5en`) and Blackwell (`p6-b300`), since the README claims both — and did the internode/low-latency benchmarks pass there? That also settles how the arch-selection finding below should be weighted. A couple of lines is all it takes.

KeitaW (commented):
## Review Batch 2/3 — Container & k8s hygiene

**`.gitignore` now un-ignores every `*.patch` in the repo (nit).** Tracking `deepep_aws_efa.patch` does require un-ignoring it, but deleting the blanket `*.patch` line removes the ignore repo-wide, so future stray patch files (`git format-patch` output, failed-rebase `.patch` leftovers) become accidentally committable. A surgical negation keeps the global guard:

```gitignore
*.patch
!micro-benchmarks/expert-parallelism/deepep-benchmark/deepep_aws_efa.patch
```

micro-benchmarks/expert-parallelism/deepep-benchmark/kubernetes/test-internode.yaml
```diff
@@ -94,6 +115,9 @@ spec:
           - image: ${IMAGE_URI}
             imagePullPolicy: IfNotPresent
             name: deepep-internode-worker
+            env:
+              - name: NVIDIA_GDRCOPY
```
KeitaW, line 119: ### `NVIDIA_GDRCOPY=enabled` is a portability safeguard here, not a requirement — and it reads more naturally in the Dockerfile

`NVIDIA_GDRCOPY=enabled` is an NVIDIA Container Toolkit (≥ v1.15) directive that injects `/dev/gdrdrv` into a runtime-isolated container — useful on a cluster whose GPU stack does *not* auto-inject it (a bare device-plugin with gdrcopy off, or plain `docker run --gpus`). But it isn't a correctness requirement on a GPU-Operator cluster: I tested on live H100+EFA hardware and a GPU pod saw `/dev/gdrdrv` **with and without** the variable, because the NVIDIA device plugin injects the device per `nvidia.com/gpu` allocation. (On the Slurm/pyxis side the device is likewise already present, via enroot's `/dev` passthrough.)

If you keep it as a safeguard, the cleaner single home is a Dockerfile `ENV NVIDIA_GDRCOPY=enabled` rather than repeating it across the three k8s manifests: the container runtime reads `NVIDIA_GDRCOPY` from the container's *merged* env, which includes image-baked `ENV` (the same way NGC images bake `NVIDIA_DRIVER_CAPABILITIES`). That collapses three copies to one line and covers non-k8s launches too. It only ever *has* effect on the legacy/jit-cdi runtime path (a GPU requested + host `gdrdrv` module loaded); it's moot under the device-plugin/CDI path and ignored by enroot. The real thing to confirm is whether NVSHMEM actually *engages* GDRCopy at runtime (`NVSHMEM_DEBUG=INFO`), not whether the node is visible.

micro-benchmarks/expert-parallelism/deepep-benchmark/README.md
```diff
@@ -45,24 +43,27 @@ image and re-exported by the launchers):
 FI_PROVIDER=efa
 NVSHMEM_REMOTE_TRANSPORT=libfabric
 NVSHMEM_LIBFABRIC_PROVIDER=efa
+NVSHMEM_NETDEVS_POLICY=EXTERNAL_SHARING_PCIE_SWITCH_NIC_EXCLUSIVE
```
KeitaW: ### `NVSHMEM_NETDEVS_POLICY` is already baked into the image `ENV` — the k8s/sbatch copies are redundant

`NVSHMEM_NETDEVS_POLICY=EXTERNAL_SHARING_PCIE_SWITCH_NIC_EXCLUSIVE` is already set as a Dockerfile `ENV` (on `main`, and kept here), next to its `NVSHMEM_REMOTE_TRANSPORT` / `NVSHMEM_LIBFABRIC_PROVIDER` siblings. It's a library variable read by the NVSHMEM process, so once it's in the image `ENV` every rank in the container inherits it — which makes the new copies in the three k8s manifests (`mpirun -x NVSHMEM_NETDEVS_POLICY`) and three sbatch files (`export …`) redundant. I'd drop the launch-surface copies and rely on the single image-`ENV` home (otherwise the hardcoded duplicates can drift from the image), unless launch-time override is the intent.

Separately, since it *is* a load-bearing knob, a one-line rationale where it's defined would help: it controls NIC↔PE assignment on multi-NIC EFA nodes and is an NVSHMEM 3.7 libfabric-transport setting (it isn't in the NVSHMEM 3.6.5 env-var reference — verified live). It's the only NVSHMEM var in that README block without surrounding prose.

KeitaW (commented):
## Review Batch 3/3 — Documentation & what looks great

### Things that look great
- **Resolves the prior round's `setup.py` issue cleanly.** Deleting `setup.py`, `src/deepep_efa_setup_script/__init__.py`, and `setup_deepep_efa.sh` together removes the orphaned, unbuildable packaging shim — and no in-repo reference to it survives in the changed files. (Heads-up only: the deleted docstring named an external consumer, `ElasticCollectivesNightlyTests`, that resolved the script from the installed package — worth a ping so that downstream isn't surprised by its removal.)
- **Single-AZ binding is correct for EFA.** The `podAffinity` with `topologyKey: topology.kubernetes.io/zone` on launcher + workers keeps internode/low-latency runs in one AZ, and you correctly left it off `test-intranode.yaml` (single-node, no benefit).
- **Every pin resolves** (verified live): DeepEP `567632d`, the `libnvshmem-linux-x86_64-3.7.0_cuda13` redist tarball, and `torch==2.11.0+cu130`.
- **Build hygiene** — `rm -rf /tmp/gdrcopy` / `aws-efa-installer` cleanup after install, and `ARG`-parameterized prefixes (`GDRCOPY_PREFIX`, `NVSHMEM_PREFIX`), keep layers tidy.
- **Moving from a build-time sed/setup-script to a pinned patch + prebuilt NVSHMEM** is a real reproducibility and build-speed win.

### Suggested priority
1. Fill the Test Plan (instance types + node counts for Hopper **and** Blackwell, and whether the benchmarks passed).
2. Either export `TORCH_CUDA_ARCH_LIST` into the DeepEP build or drop the non-functional arch knobs + README override block.

Everything else is nits — none blocking.

### Sources
- PyTorch `CUDAExtension` honors `TORCH_CUDA_ARCH_LIST` for gencode selection — https://pytorch.org/docs/stable/cpp_extension.html
- Docker BuildKit `TARGETARCH` / `RUN --mount` — https://docs.docker.com/build/buildkit/ and https://docs.docker.com/reference/dockerfile/#run---mounttypebind
- NVSHMEM 3.6.5 environment-variable reference (no `NVSHMEM_NETDEVS_POLICY`; verified live) — https://docs.nvidia.com/nvshmem/api/gen/env.html
- NVIDIA Container Toolkit GDRCopy discovery (`NVIDIA_GDRCOPY=enabled`, added v1.15.0) — https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/CHANGELOG.md
- Verified live: NVSHMEM 3.7.0 redist tarball → HTTP 200; `torch-2.11.0+cu130` wheel present; `deepseek-ai/DeepEP@567632d` → full SHA; `/dev/gdrdrv` present in a GPU pod with and without `NVIDIA_GDRCOPY` on a GPU-Operator EKS cluster.

KeitaW (changes requested):
Thank you @erezzarum ! Few comments

micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
erezzarum: both `TORCH_CUDA_ARCH_LIST` and `NVCC_GENCODE` are being exported, as the ARG definition export explicitly and there is no need to export or declare them explicitly.

Running with build arg `TORCH_CUDA_ARCH_LIST="9.0"` and `NVCC_GENCODE="-gencode=arch=compute_90,code=sm_90"`

`NVCC_GENCODE` can be used as an ARG which is used as an exportable environment variable during all build stages, no need to declare it explicitly during build command.

```
#16 [stage-0 12/17] RUN git clone -b v2.30.4-1 https://github.com/NVIDIA/nccl.git  /opt/nccl     && cd /opt/nccl     && make -j $(nproc) src.build CUDA_HOME="/usr/local/cuda"
#16 0.498 Cloning into '/opt/nccl'...
...
...
#16 2.035 make -C src build BUILDDIR=/opt/nccl/build
#16 2.036 make[1]: Entering directory '/opt/nccl/src'
#16 2.043 NVCC_GENCODE is -gencode=arch=compute_90,code=sm_90
```

DeepEP `setup.py` uses `TORCH_CUDA_ARCH_LIST`, you can see this passed without explicitly exporting it, as the ARG definition is exporting the environment variable during all build stages.

```
#21 [stage-0 17/17] RUN --mount=type=bind,source=deepep_aws_efa.patch,target=/tmp/deepep_aws/deepep_aws_efa.patch     if [ "9.0" = "9.0" ]; then DISABLE_AGGRESSIVE_PTX_INSTRS=0; else DISABLE_AGGRESSIVE_PTX_INSTRS=1; fi &&     export DISABLE_AGGRESSIVE_PTX_INSTRS &&     git clone https://github.com/deepseek-ai/DeepEP.git /DeepEP &&     cd /DeepEP &&     git checkout 567632d &&     patch -p1 < /tmp/deepep_aws/deepep_aws_efa.patch &&     CUDA_VERSION_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') &&     if [ "${CUDA_VERSION_MAJOR}" -ge 13 ]; then         sed -i "s|f'{nvshmem_dir}/include']|f'{nvshmem_dir}/include', '/usr/local/cuda/include/cccl']|" "setup.py";     fi &&     NVSHMEM_DIR="/opt/nvshmem" pip3 install -vv --no-build-isolation .
#21 0.395 Cloning into '/DeepEP'...
#21 0.691 Note: switching to '567632d'.
...
...
#21 1.009   Added file:///DeepEP to build tracker '/tmp/pip-build-tracker-r0v2_q3u'
#21 1.009   Created temporary directory: /tmp/pip-modern-metadata-9o7eiqml
#21 1.010   Preparing metadata (pyproject.toml): started
#21 1.010   Running command Preparing metadata (pyproject.toml)
#21 2.167   W0623 09:28:52.892000 40 torch/utils/cpp_extension.py:140] No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda'
#21 2.185   Build summary:
#21 2.186    > Sources: ['csrc/deep_ep.cpp', 'csrc/kernels/runtime.cu', 'csrc/kernels/layout.cu', 'csrc/kernels/intranode.cu', 'csrc/kernels/internode.cu', 'csrc/kernels/internode_ll.cu']
#21 2.186    > Includes: ['csrc/', '/opt/nvshmem/include', '/usr/local/cuda/include/cccl']
#21 2.186    > Libraries: ['/opt/nvshmem/lib']
#21 2.186    > Compilation flags: {'cxx': ['-O3', '-Wno-deprecated-declarations', '-Wno-unused-variable', '-Wno-sign-compare', '-Wno-reorder', '-Wno-attributes'], 'nvcc': ['-O3', '-Xcompiler', '-O3', '-rdc=true', '--ptxas-options=--register-usage-level=10'], 'nvcc_dlink': ['-dlink', '-L/opt/nvshmem/lib', '-lnvshmem_device']}
#21 2.186    > Link flags: ['-lcuda', '-l:libnvshmem_host.so', '-l:libnvshmem_device.a', '-Wl,-rpath,/opt/nvshmem/lib']
#21 2.186    > Arch list: 9.0
#21 2.186    > NVSHMEM path: /opt/nvshmem
```

Running with build arg `TORCH_CUDA_ARCH_LIST="9.0;10.0;10.3"` and `NVCC_GENCODE="-gencode=arch=compute_90,code=sm_90 -gencode=arch=compute_100,code=sm_100 -gencode=arch=compute_103,code=sm_103"`

```
#16 [stage-0 12/17] RUN git clone -b v2.30.4-1 https://github.com/NVIDIA/nccl.git  /opt/nccl     && cd /opt/nccl     && make -j $(nproc) src.build CUDA_HOME="/usr/local/cuda"
#16 0.656 Cloning into '/opt/nccl'...
...
...
#16 2.200 make -C src build BUILDDIR=/opt/nccl/build
#16 2.201 make[1]: Entering directory '/opt/nccl/src'
#16 2.208 NVCC_GENCODE is -gencode=arch=compute_90,code=sm_90 -gencode=arch=compute_100,code=sm_100 -gencode=arch=compute_103,code=sm_103

#21 [stage-0 17/17] RUN --mount=type=bind,source=deepep_aws_efa.patch,target=/tmp/deepep_aws/deepep_aws_efa.patch     if [ "9.0;10.0;10.3" = "9.0" ]; then DISABLE_AGGRESSIVE_PTX_INSTRS=0; else DISABLE_AGGRESSIVE_PTX_INSTRS=1; fi &&     export DISABLE_AGGRESSIVE_PTX_INSTRS &&     git clone https://github.com/deepseek-ai/DeepEP.git /DeepEP &&     cd /DeepEP &&     git checkout 567632d &&     patch -p1 < /tmp/deepep_aws/deepep_aws_efa.patch &&     CUDA_VERSION_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') &&     if [ "${CUDA_VERSION_MAJOR}" -ge 13 ]; then         sed -i "s|f'{nvshmem_dir}/include']|f'{nvshmem_dir}/include', '/usr/local/cuda/include/cccl']|" "setup.py";     fi &&     NVSHMEM_DIR="/opt/nvshmem" pip3 install -vv --no-build-isolation .
#21 0.469 Cloning into '/DeepEP'...
#21 0.755 Note: switching to '567632d'.
...
...
#21 1.087   Added file:///DeepEP to build tracker '/tmp/pip-build-tracker-cci3l8zh'
#21 1.088   Created temporary directory: /tmp/pip-modern-metadata-7wvdaqq6
#21 1.088   Preparing metadata (pyproject.toml): started
#21 1.088   Running command Preparing metadata (pyproject.toml)
#21 2.238   W0623 09:46:01.329000 40 torch/utils/cpp_extension.py:140] No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda'
#21 2.257   Build summary:
#21 2.257    > Sources: ['csrc/deep_ep.cpp', 'csrc/kernels/runtime.cu', 'csrc/kernels/layout.cu', 'csrc/kernels/intranode.cu', 'csrc/kernels/internode.cu', 'csrc/kernels/internode_ll.cu']
#21 2.257    > Includes: ['csrc/', '/opt/nvshmem/include', '/usr/local/cuda/include/cccl']
#21 2.257    > Libraries: ['/opt/nvshmem/lib']
#21 2.257    > Compilation flags: {'cxx': ['-O3', '-Wno-deprecated-declarations', '-Wno-unused-variable', '-Wno-sign-compare', '-Wno-reorder', '-Wno-attributes', '-DDISABLE_AGGRESSIVE_PTX_INSTRS'], 'nvcc': ['-O3', '-Xcompiler', '-O3', '-rdc=true', '--ptxas-options=--register-usage-level=10', '-DDISABLE_AGGRESSIVE_PTX_INSTRS'], 'nvcc_dlink': ['-dlink', '-L/opt/nvshmem/lib', '-lnvshmem_device']}
#21 2.257    > Link flags: ['-lcuda', '-l:libnvshmem_host.so', '-l:libnvshmem_device.a', '-Wl,-rpath,/opt/nvshmem/lib']
#21 2.257    > Arch list: 9.0;10.0;10.3
#21 2.257    > NVSHMEM path: /opt/nvshmem
```

I will remove `GPU_ARCH` and building NVSHMEM from source section, no need to build NVSHMEM from source.

micro-benchmarks/expert-parallelism/deepep-benchmark/kubernetes/test-internode.yaml
erezzarum: NVIDIA device plugin 0.19.0 introduced the default enablement of mofed and gdrcopy, they were unaware and this caused breaking changes, after reporting this to NVIDIA this was reverted in 0.19.3.
The cluster you are running likely used 0.19.0 or explicitly configured `GDRCOPY_ENABLED=true` in the device plugin configuration.

https://github.com/NVIDIA/k8s-device-plugin/issues/1692
https://github.com/NVIDIA/k8s-device-plugin/pull/1837

"The real thing to confirm is whether NVSHMEM actually engages GDRCopy at runtime (NVSHMEM_DEBUG=INFO), not whether the node is visible." -> Without GDRCopy driver (gdrdrv) NVSHMEM will fail for this workload. (confirmed)
```
[1,0]<stderr>:/dvs/p4/build/sw/rel/gpgpu/toolkit/r13.2/main_nvshmem/src/modules/transport/libfabric/libfabric.cpp:nvshmemt_libfabric_get_mem_handle:2559: GDRCopy support not enabled. Unable to register gpu memory handle info.
```

KeitaW (approved):

PR #1148, opened 2026-06-22
Automate architecture template publishing

KeitaW:
## Summary

- add a GitHub Actions workflow that publishes manifest-listed architecture templates after merges to `main`
- publish current PCS assets under `s3://awsome-distributed-ai/templates/aws-pcs/...` instead of flat `templates/...` keys
- update PCS launch URLs and `S3KeyPrefix` defaults to `templates/aws-pcs/` so nested stacks and scripts resolve inside the namespace
- add manifest-driven deletion of the legacy flat PCS objects after the namespaced upload succeeds
- add a staging/validation helper that detects CloudFormation templates structurally, tolerates CloudFormation YAML tags, checks duplicate destinations, and validates nested references against the architecture namespace

## Validation

- `python3 .github/scripts/stage-template-publish.py --manifest .github/template-publish-manifest.yml --stage-dir /tmp/template-publish-stage --cfn-list /tmp/template-publish-cfn-files.txt --shell-list /tmp/template-publish-shell-files.txt --obsolete-list /tmp/template-publish-obsolete-keys.txt`
- `python3 -m py_compile .github/scripts/stage-template-publish.py`
- `bash architectures/aws-pcs/tests/lint-docs.sh`
- staged shell scripts validated with `bash -n`
- verified staged output is under `/tmp/template-publish-stage/aws-pcs/...` and obsolete key list contains the old flat PCS keys

Note: local `aws cloudformation validate-template` was not run because this workspace has no default AWS credentials. The workflow runs it after GitHub OIDC role assumption.

## IAM note

The GitHub Actions role now needs `s3:DeleteObject` on `arn:aws:s3:::awsome-distributed-ai/templates/*` in addition to upload/list permissions, because this migration removes the legacy flat PCS keys.

### DIFF
--- a/.github/scripts/stage-template-publish.py
+++ b/.github/scripts/stage-template-publish.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python3
+"""Validate and stage architecture assets for S3 template publishing."""
+
+from __future__ import annotations
+
+import argparse
+import re
+import shutil
+import sys
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+
+class CfnTagLoader(yaml.SafeLoader):
+    """YAML loader that preserves structure while tolerating CloudFormation tags."""
+
+
+def _construct_unknown(loader: CfnTagLoader, node: yaml.Node) -> Any:
+    if isinstance(node, yaml.ScalarNode):
+        return loader.construct_scalar(node)
+    if isinstance(node, yaml.SequenceNode):
+        return loader.construct_sequence(node)
+    if isinstance(node, yaml.MappingNode):
+        return loader.construct_mapping(node)
+    raise TypeError(f"Unsupported YAML node: {type(node).__name__}")
+
+
+CfnTagLoader.add_multi_constructor("", lambda loader, _tag, node: _construct_unknown(loader, node))
+
+
+S3_KEY_RE = re.compile(r"^[A-Za-z0-9!_.*'()/.-]+$")
+PREFIX_REF_RE = re.compile(r"\$\{S3KeyPrefix\}([A-Za-z0-9._/-]+\.(?:ya?ml|sh))")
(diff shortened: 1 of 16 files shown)


DaisukeMiyamoto (approved):
The templates-path change makes sense, and thank you for **retaining the legacy flat `templates/...` keys** (`obsolete_keys: []`) rather than deleting them — that preserves compatibility for already-deployed stacks, which resolve nested templates and bootstrap scripts (e.g. `${S3KeyPrefix}scripts/install-enroot-pyxis.sh`) against the flat layout on stack updates and node replacement. The manifest design that lets a flat key be retired later, one line at a time once nothing references it, is the right way to stage that.

My main concern was whether moving the `S3KeyPrefix` default to `templates/aws-pcs/` would break the **own-bucket / dev** deploy flow (where templates are synced flat under a chosen prefix). I checked it both ways and it's fine:

- **Code:** the templates only ever prepend `${S3KeyPrefix}` (no hard-coded `aws-pcs/`), for both nested template URLs and the script references (`${S3KeyPrefix}scripts/...`, including the default `PostInstallScriptUrl` construction). So passing `S3KeyPrefix=templates/` against a flat layout resolves exactly as before.
- **Live:** deployed deploy-all from a private dev bucket with templates synced flat under `templates/` and `S3KeyPrefix=templates/`. Nested stacks resolved and created (Prerequisites → Cluster → CNGs), and the login node fetched `s3://<bucket>/templates/scripts/install-enroot-pyxis.sh` and installed Enroot/Pyxis successfully (`Post-install script completed (exit 0)`; `/usr/bin/enroot` + `spank_pyxis.so` present) — no 404, no path mismatch.

So the new default is purely a default; own-bucket users keep working by passing their own `S3KeyPrefix`. LGTM.

PR #1149, opened 2026-06-22
fix(eks-observability): non-existed custom-dcgm-metrics.csv cause failure in deploy-obs.sh

ybalbert001:
## Purpose

- Use default dcgm metrics if custom-dcgm-metrics.csv doesn't exist.

## Changes

- Document custom metrics as optional in README and reference ../dcgm-metrics.csv as the template to copy.
- Switch `helm upgrade` to `--reset-then-reuse-values` so cleared overrides actually take effect.

## Test Plan

**Environment:**
- AWS Service:  AMG + AMP
- Instance type:  g6.2xlarge
- Number of nodes: 1

**Test commands:**
Verified it by Manual Test

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/deploy-obs.sh
+++ b/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/deploy-obs.sh
@@ -358,24 +358,37 @@ echo ""
 echo -e "${YELLOW}Step 7: Configure Custom DCGM Metrics${NC}"
 echo ""
 
-# Create ConfigMap
-kubectl create configmap custom-dcgm-metrics \
-  --from-file=dcgm-metrics.csv=custom-dcgm-metrics.csv \
-  -n gpu-operator 2>/dev/null || echo "ConfigMap already exists, deleting and recreating..."
+# If a local custom-dcgm-metrics.csv exists, create a ConfigMap from it and
+# point DCGM Exporter at it. Otherwise leave the exporter on the GPU Operator's
+# built-in default metrics rather than pointing it at a non-existent ConfigMap
+# (which would wedge the pod in Init:0/1 on a failed mount).
+if [ -f "custom-dcgm-metrics.csv" ]; then
+    echo "Using custom DCGM metrics from: custom-dcgm-metrics.csv"
 
-if [ $? -ne 0 ]; then
-    kubectl delete configmap custom-dcgm-metrics -n gpu-operator
+    # Recreate the ConfigMap. The key must be exactly dcgm-metrics.csv for the
+    # exporter to pick it up.
+    kubectl delete configmap custom-dcgm-metrics -n gpu-operator --ignore-not-found
     kubectl create configmap custom-dcgm-metrics \
       --from-file=dcgm-metrics.csv=custom-dcgm-metrics.csv \
       -n gpu-operator
-fi
 
-# Update GPU Operator
-echo "Updating GPU Operator..."
-helm upgrade gpu-operator nvidia/gpu-operator \
-  -n gpu-operator \
-  --reuse-values \
-  --set dcgmExporter.config.name=custom-dcgm-metrics
+    echo "Updating GPU Operator to use custom metrics ConfigMap..."
+    helm upgrade gpu-operator nvidia/gpu-operator \
+      -n gpu-operator \
+      --reset-then-reuse-values \
+      --set dcgmExporter.config.name=custom-dcgm-metrics
+else
+    echo -e "${YELLOW}No custom-dcgm-metrics.csv found; using GPU Operator default DCGM metrics.${NC}"
+
+    # Drop any stale custom ConfigMap and clear the override (=null removes the
+    # key) so the exporter reverts to the operator's built-in default metrics.
+    kubectl delete configmap custom-dcgm-metrics -n gpu-operator --ignore-not-found
+    echo "Updating GPU Operator to use default metrics..."
+    helm upgrade gpu-operator nvidia/gpu-operator \
+      -n gpu-operator \
+      --reset-then-reuse-values \
+      --set dcgmExporter.config.name=null
+fi
 
 # Restart DCGM Exporter
 echo "Restarting DCGM Exporter..."
--- a/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/deploy-obs.sh
+++ b/4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/deploy-obs.sh
@@ -12,6 +12,15 @@ echo -e "${GREEN}EKS GPU Observability Setup${NC}"
 echo -e "${GREEN}========================================${NC}"
 echo ""
 
+# Helm >= 3.14 is required for `--reset-then-reuse-values` (introduced in
+# v3.14.0) used by the GPU Operator upgrades in Step 7. Fail loudly up front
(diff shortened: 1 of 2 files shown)


4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/deploy-obs.sh
```diff
@@ -358,24 +358,37 @@ echo ""
...
-# Update GPU Operator
-echo "Updating GPU Operator..."
-helm upgrade gpu-operator nvidia/gpu-operator \
-  -n gpu-operator \
-  --reuse-values \
-  --set dcgmExporter.config.name=custom-dcgm-metrics
+    echo "Updating GPU Operator to use custom metrics ConfigMap..."
+    helm upgrade gpu-operator nvidia/gpu-operator \
+      -n gpu-operator \
+      --reset-then-reuse-values \
```
KeitaW, line 387: ### Document the new Helm ≥ 3.14 requirement

This PR switches all four `helm upgrade` calls from `--reuse-values` to `--reset-then-reuse-values` — both here in `deploy-obs.sh` (lines 378 and 389) and in the README's two manual commands. That flag was only added in **Helm v3.14.0**. The README Prerequisites list `kubectl`/AWS CLI but pin no Helm version.

On any host with Helm < 3.14, `helm upgrade` fails with `Error: unknown flag: --reset-then-reuse-values`. Under `set -e` that aborts the script mid-Step-7 — and because the default branch deletes the ConfigMap *before* this upgrade runs (see the other comment), it aborts in exactly the wedged state this PR sets out to remove.

I verified this on a live EKS cluster (NVIDIA GPU Operator v26.3.2): with Helm 3.13 Step 7 aborts with `unknown flag: --reset-then-reuse-values`; with Helm ≥ 3.14 it succeeds.

I'd suggest adding a Helm floor to the README Prerequisites, e.g. `- Helm ≥ 3.14 (for --reset-then-reuse-values)`, and optionally failing loudly up front near the top of the script:

```bash
helm version --short | grep -qE 'v3\.(1[4-9]|[2-9][0-9])' || { echo 'Helm >= 3.14 required for --reset-then-reuse-values'; exit 1; }
```

Citation: [Helm v3.14.0 release notes](https://github.com/helm/helm/releases/tag/v3.14.0) — flag introduced.

4.validation_and_observability/4.prometheus-grafana/eks-managed-observability/deploy-obs.sh
```diff
@@ -358,24 +358,37 @@ echo ""
...
+    echo -e "${YELLOW}No custom-dcgm-metrics.csv found; using GPU Operator default DCGM metrics.${NC}"
+
+    # Drop any stale custom ConfigMap and clear the override (=null removes the
+    # key) so the exporter reverts to the operator's built-in default metrics.
+    kubectl delete configmap custom-dcgm-metrics -n gpu-operator --ignore-not-found
+    echo "Updating GPU Operator to use default metrics..."
+    helm upgrade gpu-operator nvidia/gpu-operator \
+      -n gpu-operator \
+      --reset-then-reuse-values \
+      --set dcgmExporter.config.name=null
```
KeitaW: ### Default branch deletes the ConfigMap before dereferencing it — reorder to dereference-then-delete

In the `else` (no-csv / default-metrics) branch, `kubectl delete configmap custom-dcgm-metrics` runs *first* (line 385), then `helm upgrade … --set dcgmExporter.config.name=null` removes the release's reference to it. The `if` branch gets the order right — it creates the ConfigMap *before* pointing Helm at it — but this branch is the mirror image: it removes the object before removing the thing that references it.

On a custom→default re-run, if that `helm upgrade` fails, the script exits (`set -e`) with the ConfigMap already gone but the release still referencing `dcgmExporter.config.name=custom-dcgm-metrics` → DCGM Exporter wedges in `Init:0/1` on the missing mount — the state the comment just above says it's avoiding. The most realistic trigger for that failed upgrade is the Helm < 3.14 case in the other comment, so the two compound.

I reproduced exactly this on a live cluster: forcing the upgrade to fail in this branch left the exporter pod in `Init:0/1` with `MountVolume.SetUp failed for volume "metrics-config": configmap "custom-dcgm-metrics" not found`. Reordering to dereference-then-delete avoided it.

Dereference first, then delete — so the ConfigMap is only removed once nothing points at it:

```suggestion
    # Point the exporter back at the operator's built-in defaults first, so
    # nothing references the custom ConfigMap before we remove it.
    echo "Updating GPU Operator to use default metrics..."
    helm upgrade gpu-operator nvidia/gpu-operator \
      -n gpu-operator \
      --reset-then-reuse-values \
      --set dcgmExporter.config.name=null

    # Now that the release no longer references it, drop any stale ConfigMap
    # (=null above removed the key).
    kubectl delete configmap custom-dcgm-metrics -n gpu-operator --ignore-not-found
```

KeitaW (commented):
## Review Batch 1/2 — Deployment Pipeline & Operational Correctness

Two operational-correctness items on Step 7, both reproduced on a live EKS + GPU Operator cluster. Neither is a hard blocker, but they compound: the version gap is the most likely way to trigger the ordering wedge.

KeitaW (commented):
## Review Batch 2/2 — Positives & Sources

### Things That Look Great
- **The actual bug is correctly diagnosed and fixed.** The old block — `kubectl create … || echo …; if [ $? -ne 0 ]; then …` — could never recreate the ConfigMap: `$?` reflected the trailing `echo` (always 0), so the delete/recreate path was dead. Replacing it with an explicit `if [ -f custom-dcgm-metrics.csv ]` branch is the right call and removes the dead `$?` check entirely.
- **The `--set dcgmExporter.config.name=null` revert is sound.** `--set` (not `--set-string`) parses `null` as YAML nil, and the gpu-operator chart leaves `dcgmExporter.config.name` unset by default — so clearing it reverts the exporter to the operator's built-in default metrics. Verified live: the default-metrics branch leaves the exporter healthy on `dcp-metrics-included.csv`, and the custom branch correctly mounts the ConfigMap and emits the extra profiling/power/thermal metrics.
- **The `../dcgm-metrics.csv` template reference resolves.** Both the README link and the `cp ../dcgm-metrics.csv custom-dcgm-metrics.csv` instruction point at a real file in the parent directory — no broken relative link.
- **Docs move in lockstep with the code.** Step 7 is re-titled "(Optional)", the quick-setup block gains an opt-in comment, and a clear Note explains both branches.
- **Good, accurate rationale comments.** The "would wedge the pod in Init:0/1 on a failed mount" comment captures the *why* — and it's exactly the state I reproduced when the upgrade fails mid-branch.

### Sources
- [Helm v3.14.0 release notes](https://github.com/helm/helm/releases/tag/v3.14.0) — `--reset-then-reuse-values` introduced in 3.14.0.
- Verified live on EKS + NVIDIA GPU Operator v26.3.2 (g6.2xlarge / L4): default-path and custom-path both leave the DCGM Exporter healthy; the Helm-<3.14 + delete-before-dereference combination reproduces the `Init:0/1` `FailedMount` wedge.

KeitaW (changes requested):
Thank you! Few comments

ybalbert001:
> Thank you! Few comments

Thanks Keita, Updated.

KeitaW (approved):
LGTM!

PR #1150, opened 2026-06-22
Add PointWorld 3D world model pre-training test case (EKS)

mvinci12:
## What

Adds a new PyTorch test case under `3.test_cases/pytorch/pointworld/` for distributed pre-training and evaluation of [PointWorld](https://github.com/NVlabs/PointWorld) (NVlabs, CVPR 2026 Highlight) — a large 3D world model that forecasts full-scene 3D point flow for robotic manipulation (PTv3 backbone + DINOv3 scene featurizer).

This first cut is **Kubernetes-only** (EKS / SageMaker HyperPod EKS).

## Agenda

PointWorld is designed as *pre-train once, no post-training* (a single checkpoint drives a real robot via MPC). The test case demonstrates:

1. **Pre-train** the world model (multi-node DDP) on BEHAVIOR, large PTv3.
2. **Evaluate** the resulting (or a released) checkpoint.

No fine-tuning stage by design; real-robot MPC deployment is off-cluster and out of scope.

**Scope:** this test case stages and trains the **BEHAVIOR** domain only. The committed defaults are a small **2-node / 16-GPU** smoke run over a BEHAVIOR subset, so the full data → train → evaluate → visualize loop runs on a modest footprint; `env_vars` knobs scale it to more nodes and the full dataset. PointWorld's other domain, **DROID**, is a ~3.9 TB monolithic split archive and is **out of scope** here (the data-prep Job does not stage it); `scripts/0.download_dataset.py` + upstream cover DROID for users who want it.

## Validated end-to-end on Amazon EKS

Exercised on a live EKS cluster (Kubeflow **Trainer v2.0.0**, p5en.48xlarge / H200):

- [x] **Infra**: 2-node / 16-GPU `torchrun` NCCL all-reduce as a Trainer v2 TrainJob — gang-scheduled on p5en, distributed init OK.
- [x] **Container build**: `pointworld.Dockerfile` built on-cluster (Kaniko), pushed to ECR (~14.9 GB), imports cleanly on H200 (torch 2.5.1+cu124, flash-attn 2.7.4.post1, PTv3, DINOv3).
- [x] **Pre-training**: 1-node / 8-GPU micro-run (BEHAVIOR, `--ptv3_size=large`, `--max_train_steps=2`) — distributed init -> DINOv3 backbone -> WebDataset dataloader -> forward/backward with loss, clean exit.
- [x] **Evaluation**: `eval.py` loaded a released checkpoint (`nvidia/PointWorld_models` large-droid+behavior, 13 GB) and produced metrics on the BEHAVIOR test split.

### Fixes found during validation (in this branch)

- **Dockerfile**: `apt-get update` before the EFA installer (needs environment-modules/tcl); add `zstd` (dataset recovery needs it); purge the base image's stale opencv 4.7.0 cv2 tree and install `opencv-python-headless==4.11.0.86` (fixes `import cv2` / `cv2.dnn.DictValue`).
- **Kubernetes**: added [`kubernetes/trainer-v2/`](https://github.com/awslabs/awsome-distributed-ai/tree/main/3.test_cases/pytorch/pointworld/kubernetes/trainer-v2) (TrainJob + ClusterTrainingRuntime) for Trainer v2 clusters — requires the `trainer.kubeflow.org/trainjob-ancestor-step: trainer` label and a `torchrun` entrypoint. Corrected EFA request for p5en.48xlarge to **16** (was 32).
- **Scripts**: `1.convert_wds.py` now runs integrity-check -> make_wds_manifest -> convert_wds in order and matches the upstream CLI; `2.download_dinov3.sh` keeps the canonical weight filename `dinov3_vitl16_pretrain_lvd1689m-8aa4cbdd.pth` (the scene featurizer rejects renamed files).

## Contents

| Path | Purpose |
|---|---|
| `pointworld.Dockerfile` | CUDA 12.4 base; clones NVlabs/PointWorld at a pinned commit; pinned deps + flash-attn |
| `env_vars.template` | Copy to `env_vars`; one place for region/account/cluster/run params (sourced before deploy) |
| `kubernetes/deploy.sh` | Render manifests with `envsubst` (explicit allowlist) + apply; `--dry-run`/`--delete` |
| `kubernetes/build-image.{sh,yaml}` | On-cluster Kaniko build → ECR (linux/amd64; builds on x86_64, no local Docker) |
| `kubernetes/pointworld-data-prep.yaml` | Job: stage DINOv3 + WebDataset shards onto FSx (runs in-cluster) |
| `kubernetes/pointworld-pretrain.yaml` | PyTorchJob v1 — multi-node DDP via torchrun, EFA + FSx (default 2x p5en / 16x H200, ${NUM_NODES}-driven) |
| `kubernetes/pointworld-eval.yaml` | single-GPU Job running `eval.py` (domain selected via `EVAL_DOMAIN`) |
| `kubernetes/pointworld-viz.yaml` | interactive `viser` 3D viewer Pod (stdin/tty + port 8080), driven via `kubectl attach` + `port-forward` |
| `kubernetes/trainer-v2/` | Kubeflow Trainer v2 TrainJob + ClusterTrainingRuntime variants (validated) |
| `scripts/0-2 + parse_benchmark.py` | dataset download, H5->WDS, gated DINOv3 fetch, throughput parser |

## Notes

- DINOv3 weights are **gated by Meta** — documented prerequisite.
- **Config via `env_vars` + `envsubst`** — fill in region/account/cluster/run params once, then `./kubernetes/deploy.sh <manifest>` renders and applies (explicit allowlist preserves in-container vars like `$RANK`/`$LOCAL_DATASET_DIR`). Matches the repo convention (cf. `nvrx`).
- **`linux/amd64` only** — the NGC base image, PyG `torch-scatter+pt25cu124` wheels, and p5en nodes are all x86_64. Build via the on-cluster Kaniko helper (`./kubernetes/build-image.sh`) or `docker buildx --platform linux/amd64`; native arm64 builds are not possible.
- **FSx is not mounted locally** — data/weight staging runs in-cluster via `kubernetes/pointworld-data-prep.yaml` (mounts the `fsx-claim` PVC); the `scripts/*` are documented manual equivalents.
- Targets H200; B200 needs NCCL >= 2.29 (noted under Known Limitations).
- DROID is a ~3.9 TB monolithic split archive and is **out of scope** for this test case; BEHAVIOR (task-sharded, subsets cleanly) is what the data-prep Job stages and what the committed defaults train on.
- **Interactive visualization** — `eval.py --eval_viz_num > 0` serves a live [`viser`](https://github.com/nerfstudio-project/viser) 3D viewer on port 8080. Because it is a live web server *and* pauses on an interactive TTY prompt, it ships as a dedicated `stdin/tty` Pod (`pointworld-viz.yaml`) rather than a batch Job, driven via `kubectl attach` + `kubectl port-forward`. The image bakes the upstream viz extras (`viser==1.0.10`, `open3d==0.16.0`, `matplotlib==3.10.6`, pinned to `environments/train_eval_viz.yml`).

## Validation update

The committed defaults match what was validated: a small **2-node / 16-GPU** BEHAVIOR smoke run (3 tasks / `MAX_CLIPS=200`, `NUM_EPOCHS=10`). The same path was also exercised at a larger scale (10 tasks, 4000 train / 1000 test clips) on p5en/H200 (Kubeflow Trainer v2, EFA `efa-direct`): distributed init, DINOv3 load, WebDataset streaming, decreasing loss, and a `model-last.pt` checkpoint on the epoch cutoff. The interactive `viser` viewer was validated end-to-end against a trained checkpoint (server live on :8080, paused at the per-sample prompt).

Marked **draft** for maintainer review. BEHAVIOR is the intended scope (DROID is out of scope, documented). Commits are DCO signed-off.


bluecrayon52 (commented):
Thanks for this — it's a thorough, well-documented contribution that genuinely follows the repo conventions (pinned upstream commit, numbered scripts, self-contained README, `env_vars` + `envsubst` à la `nvrx`). I verified the external deps are live: the pinned commit `05484826` and the `data` branch both exist in `NVlabs/PointWorld`, and the HF datasets/checkpoints + the DINOv3 repo all resolve. Security hygiene is good too (gated DINOv3 URL kept out of files, `env_vars` gitignored, short-lived ECR cred secret with a cleanup trap).

I have one substantive concern worth resolving before merge, plus a few minor items.

## Main concern: the documented *default* path (v1 PyTorchJob) may not have been tested multi-node

All the validation evidence in the PR — the README "Validation status", `kubernetes/trainer-v2/README.md`, and the PR description — is for **Kubeflow Trainer v2.0.0**. But the top-level README presents `kubernetes/pointworld-pretrain.yaml` (**PyTorchJob v1**, `kubeflow.org/v1`) as the primary path and Trainer v2 as the alternative.

The v1 manifest sets neither `elasticPolicy` nor a spec-level `nprocPerNode`:

```yaml
spec:
  pytorchReplicaSpecs:
    Worker:
      replicas: ${NUM_NODES}
```

For a `torchrun --nnodes=2 --nproc_per_node=8` launch, the launcher needs a `node_rank` and a rendezvous endpoint. It gets these from the `PET_*` env vars, which the Training Operator only injects when the job is in **elastic/torchrun mode** — i.e. when `elasticPolicy` (`rdzvBackend: c10d`) or a spec-level `nprocPerNode` is set. Every existing FSDP PyTorchJob in this repo does exactly that (e.g. `FSDP/kubernetes/llama3_1_8b-fsdp.yaml` uses `elasticPolicy`; the hpto variants use `nprocPerNode`).

Without one of those, the operator runs in **classic mode** and injects only pod-level `MASTER_ADDR/PORT/WORLD_SIZE/RANK` (where `WORLD_SIZE` = replica count and `RANK` = pod index, designed for one-proc-per-pod). `torchrun` then sees no `PET_*` rendezvous and no `node_rank`, falls back to its static defaults (`node_rank=0`, `master_addr=127.0.0.1`), and **each pod rendezvouses as a standalone node 0** — so a 2-node run becomes two independent 8-GPU jobs (or errors at init) rather than one gang-scheduled 16-GPU job.

Notably, the comment at `pointworld-pretrain.yaml:71-73` states the v1 operator "injects ... the `PET_*` elastic variables" — that's only true in elastic/`nprocPerNode` mode, which this manifest doesn't enable. The Trainer v2 path gets `PET_*` from the v2 torch plugin instead, so the validated v2 runs would have worked regardless of this — which is what makes me suspect the v1 path wasn't actually exercised multi-node.

**Could you confirm whether `pointworld-pretrain.yaml` was run multi-node, or only the Trainer v2 path?** Easiest empirical check: on a running v1 worker pod, `env | grep -E 'PET_|RANK|WORLD'` and confirm `WORLD_SIZE=16` (correct) vs `8` (split-brain), or check that NCCL init reports 16 ranks across 2 nodes. If it wasn't tested, please either (a) add `elasticPolicy: { rdzvBackend: c10d, minReplicas, maxReplicas, maxRestarts }` (or a spec-level `nprocPerNode`) to the v1 manifest and validate, or (b) make Trainer v2 the documented default and clearly mark v1 as untested. CONTRIBUTING.md asks that test cases be "tested and run at the target scale."

## Minor

- **Inconsistent EFA env between the two paths.** `pointworld-pretrain.yaml` (v1) sets `FI_EFA_SET_CUDA_SYNC_MEMOPS=0` but not `FI_EFA_USE_DEVICE_RDMA` / `FI_EFA_FORK_SAFE`; the v2 runtime sets `FI_EFA_USE_DEVICE_RDMA=1` + `FI_EFA_FORK_SAFE=1` but not the sync-memops one. Same hardware and workload — these should match (and since v2 is the validated path, the v1 EFA config is likely the unverified one).
- **`scripts/parse_benchmark.py` regex defaults likely won't match out-of-the-box.** Throughput is the headline deliverable for a distributed-training test case, but the script's own docstring notes "PointWorld logs do not include per-step time by default; throughput will be skipped unless this matches," and the default `--step_regex=B=(\d+)` / `--loss_regex=Loss=...` are described as guesses. Could you paste a real log line from a validated run and confirm the defaults produce a throughput number without manual regex tweaking?
- **Eval `backoffLimit: 2` with a placeholder `MODEL_PATH`.** The default `env_vars` `MODEL_PATH` is a literal `<wandb-run-name>` placeholder; if a user forgets to set it, the eval Job retries 3× before failing. A fail-fast guard (like the data-prep Job's `DINOV3_URL` check) would be friendlier.

## Nits

- `scripts/0.download_dataset.py` is documented as the manual helper but isn't wired into the data-prep Job (which inlines its own `snapshot_download`). Intentional and fine, but the duplication can drift — a one-line note would help.
- The `sed 's/^ \{14\}//'` dedent in the data-prep heredoc is correct today (verified the body is uniformly 14-space indented) but fragile to future edits; a `textwrap.dedent` or a ConfigMap-mounted script would be more robust. Non-blocking.

## Verified clean

- `sed` dedent indentation math ✓
- `.gitignore` covers `env_vars`, checkpoints, datasets ✓
- PNG committed raw (~1.08 MB), consistent with the repo's other raw PNGs and not LFS-required per `.gitattributes` ✓
- MIT-0 headers on all new files ✓
- Pinned versions throughout, no `latest` ✓

mvinci12:
Thanks for the careful review, Nathan — especially the rendezvous analysis. You're right on the main concern, and I've pushed `0775f49` addressing everything.

## Main concern — you nailed it

The v1 `pointworld-pretrain.yaml` was **not** exercised multi-node; all my validation was on Trainer v2 (the v2 torch plugin supplies `PET_*`, which masked the gap). The v1 manifest had `replicas` but no `elasticPolicy`/`nprocPerNode`, so a 2-node run would have split-brained exactly as you describe.

Rather than claim a multi-node v1 run I didn't do, I went with a hybrid of your two options:

- **Made Trainer v2 the documented default** (top-level README §5a, `kubernetes/README.md` table + warning callout) since that's the validated path.
- **Fixed the v1 manifest anyway** so it's correct for Training-Operator clusters: added `elasticPolicy: { rdzvBackend: c10d, minReplicas: 1, maxReplicas: ${NUM_NODES}, maxRestarts: 100 }` (matching the FSDP convention you cited) and corrected the misleading `PET_*` comment.
- **Explicitly marked the v1 path "not validated multi-node"** with a callout pointing at the empirical check you suggested (`env | grep -E 'PET_|WORLD_SIZE'` → expect `WORLD_SIZE=16`). Honest per CONTRIBUTING.md rather than asserting an untested scale.

## Minor

- **EFA env mismatch** — aligned both paths on `FI_EFA_USE_DEVICE_RDMA=1`. I dropped `FI_EFA_SET_CUDA_SYNC_MEMOPS` (v1) and `FI_EFA_FORK_SAFE` (v2) rather than unioning them, to keep the two configs identical and minimal (NCCL's OFI layer reports it already sets `FORK_SAFE` internally — visible in `run.log`).
- **`parse_benchmark.py` regex** — confirmed the defaults **do** match out-of-the-box. Real line from a validated run:
  ```
  [22:48:38 dummy-5b5pxh8p], Epoch=2.354/200, B=16, S=352, Loss=2.936e+03
  ```
  `--step_regex=B=(\d+)` and `--loss_regex=Loss=...` match as-is. There's no per-step *duration* token, so throughput is derived from the leading `[HH:MM:SS]` prefix via `--timestamp_regex` (not `--time_regex`). I corrected the docstring, which previously described a `step=<N>` format that doesn't match reality.
- **Eval `backoffLimit` + placeholder** — added a fail-fast guard: if `MODEL_PATH` is still the `<wandb-run-name>` placeholder (or the checkpoint doesn't exist), the Job exits immediately with a clear message instead of burning retries. Kept `backoffLimit: 2` for genuine transient failures (FSx mount races) and documented why.

## Nits

- Added a note that the data-prep Job inlines its own `snapshot_download` rather than calling `scripts/0.download_dataset.py` (drift caveat), as you suggested.
- Left the `sed` dedent as-is per your "non-blocking" — agreed it's correct today; happy to switch to a ConfigMap-mounted script if you'd prefer.

The one thing I have **not** done is re-validate the v1 path multi-node on a live cluster — flagging that openly. If you'd rather gate merge on that, I can schedule a 2-node p5en run; otherwise the v1 path stays clearly labeled unvalidated with v2 as the default.

bluecrayon52 (approved):
Approved with Trainer v2 the documented default.

PR #1111, opened 2026-05-27
feat: add Cosmos Reason vLLM inference test case

mvinci12:
## Summary

Adds an online inference test case for NVIDIA's Cosmos Reason physical-reasoning Vision-Language Model, served by vLLM on Amazon EKS or SageMaker HyperPod EKS.

## What this adds

`3.test_cases/pytorch/vllm/cosmos-reason/` — 14 files, ~1300 lines.

### Two parallel deployment paths (both `kubectl`-only)

| Path | Container | Use case |
|------|-----------|----------|
| `kubernetes/` | Upstream `vllm/vllm-openai:v0.21.0` | Plain EKS users, no HyperPod required |
| `hyperpod-eks/` | AWS-managed vLLM DLC (`vllm:0.17-gpu-py312`) | HyperPod EKS clusters using the Inference Operator — auto KEDA + Karpenter scale-to-zero, managed KV cache, intelligent routing |

### Three example use cases

| Script | Pattern |
|--------|---------|
| `examples/image_vqa.py` | Single-image visual Q&A |
| `examples/video_qa.py` | Short video clip Q&A |
| `examples/auto_label.py` | SDG auto-labeling with structured JSON output + retry logic |

### Sample data

`examples/download_samples.sh` fetches CC-licensed media from Unsplash and Wikimedia Commons. Sample files are gitignored.

## Validation

Validated end-to-end on a SageMaker HyperPod EKS cluster:
- Hardware: g5.8xlarge (1× NVIDIA A10G 24 GB)
- Model: `nvidia/Cosmos-Reason1-7B`
- Container: `vllm/vllm-openai:v0.21.0`
- Both deployment paths exercised

## Design decisions

- Default model is Cosmos-Reason1-7B (not Reason2) because Reason2 is HF-gated
- No HPA included — inference is GPU-bound; CPU-based scaling is misleading. README points at KEDA + `vllm:num_requests_running` for production.
- TLS verification defaults to ON; `--insecure` flag available for self-signed certs
- `auto_label.py` includes `--max-retries` with exponential backoff for transient HTTP errors
- Pre-flight `envsubst` dry-run validation step documented in Quick Start

PR #1155, opened 2026-06-24
feat(dynamo): add NVIDIA Dynamo (SGLang) LLM inference test case on HyperPod EKS

matheus-1618:
## Purpose

Add an online **LLM inference** test case using [NVIDIA Dynamo](https://docs.nvidia.com/dynamo/)
with the SGLang backend on **SageMaker HyperPod EKS**. It serves two models —
GPT-OSS-20B and Qwen3.6-27B-FP8 — each in **aggregated** and **disaggregated**
(prefill/decode split + NIXL KV transfer) flavors, via the Dynamo operator and the
`DynamoGraphDeployment` (DGD) CRD.

Complements the existing vLLM inference case (`pytorch/vllm/cosmos-reason`, #1111) by
adding Dynamo's disaggregated serving and NIXL-based KV cache transfer.

## Changes

- New test case `3.test_cases/pytorch/nvidia-dynamo/`:
  - 4 `DynamoGraphDeployment` scenarios (gpt-oss & qwen3.6, each aggregated + disaggregated)
  - Numbered scripts `01`–`10` (install platform, download model, deploy, test, interactive
    streaming chat, AIPerf benchmark, cleanup, uninstall)
  - `platform/values.yaml` (Dynamo platform Helm values: operator on control nodes, etcd/NATS on GPU workers)
  - `env_vars.example`, `hf-token-secret.yaml.example`
  - `README.md` (self-contained, with a Troubleshooting table), `BENCHMARKS.md`, `OBSERVABILITY.md`
- SPDX/MIT-0 headers on all scripts and manifests; all container images pinned
  (`nvcr.io/nvidia/ai-dynamo/sglang-runtime:1.2.1`, platform chart `0.7.0`) — no `:latest`.

## Test Plan

**Environment:**
- Instance type: `ml.g6e.4xlarge` (NVIDIA L40S 48GB) GPU workers + `ml.c6i.xlarge` control
- Number of nodes: 2 GPU workers + 2 control

**Test commands:**
```bash
cp env_vars.example env_vars && source env_vars
./scripts/01-install-platform.sh
./scripts/02-download-model.sh gpt-oss
./scripts/02-download-model.sh qwen3.6
./scripts/03-deploy-gpt-oss-agg.sh        # (and 04/05/06 for the other scenarios)
./scripts/07-test-inference.sh            # auto-detects the deployed scenario
./scripts/08-benchmark.sh <scenario>
```

## Test Results

All four scenarios deploy, register, and serve OpenAI-compatible chat completions; the
disaggregated scenarios complete the NIXL prefill→decode handshake. AIPerf sweep
(ISL 200 / OSL 256) on a single L40S — full tables in `BENCHMARKS.md`:

- **GPT-OSS-20B**: ITL ≈7–16 ms (≈60–140 tok/s/user); ~398 tok/s aggregate at c=10 (agg),
  ~447 tok/s at c=10 (disagg, lower TTFT under load).
- **Qwen3.6-27B-FP8**: ITL ≈45–52 ms (hybrid Gated DeltaNet/SSM decode); ~140 tok/s
  aggregate at c=10 (disagg).

> Disaggregation here uses NIXL over **TCP** on g6e (no EFA). For RDMA, deploy on
> EFA-enabled instances (e.g. p5).

## Directory Structure

```
3.test_cases/pytorch/nvidia-dynamo/
├── README.md
├── BENCHMARKS.md
├── OBSERVABILITY.md
├── env_vars.example
├── hf-token-secret.yaml.example
├── platform/values.yaml
├── scenarios/
│   ├── gpt-oss-agg/      gpt-oss-20b-agg.yaml
│   ├── gpt-oss-disagg/   gpt-oss-20b-disagg.yaml
│   ├── qwen3.6-agg/      qwen3.6-27b-fp8-agg.yaml
│   ├── qwen3.6-disagg/   qwen3.6-27b-fp8-disagg.yaml
│   └── observability/    pod-monitor.yaml
└── scripts/              01..10 (install, download, deploy, test, benchmark, cleanup)
```

### DIFF
--- a/3.test_cases/pytorch/nvidia-dynamo/OBSERVABILITY.md
+++ b/3.test_cases/pytorch/nvidia-dynamo/OBSERVABILITY.md
@@ -80,4 +80,4 @@ The Dynamo frontend exposes these metric families:
 - Use `rate(...[5m])` for counters and histograms to see per-second rates
 - Use `histogram_quantile(0.50, ...)` for p50, `0.99` for p99
 - Filter by model: add `{model="openai/gpt-oss-20b"}` to any query
-- Run a benchmark (`./scripts/05a-test-benchmark.sh`) to generate traffic and see metrics in action
+- Run a benchmark (`./scripts/08-benchmark.sh <scenario>`) to generate traffic and see metrics in action
--- a/3.test_cases/pytorch/nvidia-dynamo/OBSERVABILITY.md
+++ b/3.test_cases/pytorch/nvidia-dynamo/OBSERVABILITY.md
@@ -1,0 +1,83 @@
+# Observability Guide
+
+Dynamo frontend exposes Prometheus metrics at `/metrics`. This guide covers how to get those metrics into the HyperPod Grafana dashboard.
+
+## Setup
+
+Run after installing the platform (step 01):
+
+```bash
+./scripts/01a-setup-observability.sh
+```
+
+This creates:
+1. A stable Kubernetes Service (`dynamo-metrics`) that always points to the active frontend pod via label selector
+2. A `customServiceScrapeTarget` in the HyperPod `ObservabilityConfig` CR, which tells the OTEL collector to scrape Dynamo metrics and send them to AMP
+
+Metrics flow: **Dynamo pods → dynamo-metrics Service → OTEL collector → AMP → Grafana**
+
+## Accessing Grafana
+
+The HyperPod cluster creates an Amazon Managed Grafana workspace automatically.
+
+### 1. Find the Grafana URL
+
+```bash
+aws grafana list-workspaces --region <REGION> --no-cli-pager \
+  --query 'workspaces[*].{name:name,endpoint:endpoint,status:status}' --output table
+```
+
+Open the `endpoint` URL in your browser. Login is via AWS IAM Identity Center (SSO).
+
+### 2. Get Admin access
+
+By default, users are added as **Viewer** — which doesn't have access to Explore or dashboard creation. To promote to Admin:
+
+```bash
+# List current permissions
+aws grafana list-permissions --workspace-id <WORKSPACE_ID> --region <REGION> --no-cli-pager
+
+# Promote your user to Admin (use the user ID from the list above)
+aws grafana update-permissions --workspace-id <WORKSPACE_ID> --region <REGION> --no-cli-pager \
+  --update-instruction-batch '[{"action":"ADD","role":"ADMIN","users":[{"id":"<USER_ID>","type":"SSO_USER"}]}]'
+```
+
+After updating, **logout and login again** for the new role to take effect.
+
+## Viewing Metrics
+
+### Explore (ad-hoc queries)
+
+1. Go to **Explore** (compass icon in the sidebar — only visible to Editor/Admin)
+2. Select the **prometheus** data source
+3. Try these queries:
+
+| Query | What it shows |
+|---|---|
+| `dynamo_frontend_inflight_requests` | Current in-flight requests |
+| `rate(dynamo_frontend_input_sequence_tokens_count[5m])` | Request rate (req/s) |
+| `histogram_quantile(0.50, rate(dynamo_frontend_time_to_first_token_seconds_bucket[5m]))` | TTFT p50 |
+| `histogram_quantile(0.99, rate(dynamo_frontend_time_to_first_token_seconds_bucket[5m]))` | TTFT p99 |
+| `histogram_quantile(0.50, rate(dynamo_frontend_inter_token_latency_seconds_bucket[5m]))` | ITL/TPOT p50 |
+| `histogram_quantile(0.50, rate(dynamo_frontend_request_duration_seconds_bucket[5m]))` | Request duration p50 |
+
+### Available Metrics
+
+The Dynamo frontend exposes these metric families:
+
+| Metric | Type | Description |
+|---|---|---|
+| `dynamo_frontend_inflight_requests` | Gauge | Number of in-flight requests |
+| `dynamo_frontend_disconnected_clients` | Gauge | Disconnected client count |
+| `dynamo_frontend_input_sequence_tokens` | Histogram | Input sequence length distribution |
+| `dynamo_frontend_output_sequence_tokens` | Histogram | Output sequence length distribution |
+| `dynamo_frontend_time_to_first_token_seconds` | Histogram | TTFT distribution |
+| `dynamo_frontend_inter_token_latency_seconds` | Histogram | Inter-token latency (TPOT) distribution |
+| `dynamo_frontend_request_duration_seconds` | Histogram | Total request duration distribution |
+
+### Tips
+
+- Use `rate(...[5m])` for counters and histograms to see per-second rates
+- Use `histogram_quantile(0.50, ...)` for p50, `0.99` for p99
+- Filter by model: add `{model="openai/gpt-oss-20b"}` to any query
+- Run a benchmark (`./scripts/05a-test-benchmark.sh`) to generate traffic and see metrics in action
--- a/3.test_cases/pytorch/nvidia-dynamo/README.md
+++ b/3.test_cases/pytorch/nvidia-dynamo/README.md
@@ -1,0 +1,165 @@
+# NVIDIA Dynamo (SGLang) — LLM Inference on SageMaker HyperPod EKS
+
+[NVIDIA Dynamo](https://docs.nvidia.com/dynamo/) is a distributed inference framework
+providing KV cache-aware routing, disaggregated prefill/decode, and NIXL-based KV
+transfer for LLM serving on Kubernetes. This test case serves two models on a SageMaker
+HyperPod EKS cluster with the SGLang backend, each in **aggregated** and **disaggregated**
+flavors — four self-contained scenarios driven by the Dynamo operator and its
+`DynamoGraphDeployment` (DGD) CRD.
+
+| Scenario | Model | Pattern | GPUs |
+|---|---|---|---|
+| `gpt-oss-agg`    | GPT-OSS-20B      | Aggregated (1 worker = prefill+decode) | 1 |
+| `gpt-oss-disagg` | GPT-OSS-20B      | Disaggregated (prefill + decode, NIXL) | 2 |
+| `qwen3.6-agg`    | Qwen3.6-27B-FP8  | Aggregated                             | 1 |
+| `qwen3.6-disagg` | Qwen3.6-27B-FP8  | Disaggregated (prefill + decode, NIXL) | 2 |
+
+- **Aggregated**: one worker handles the full request; the frontend routes by KV-cache overlap.
+- **Disaggregated**: prefill and decode run on separate workers; the KV cache is transferred
+  prefill→decode via NIXL (RDMA where EFA is available, otherwise TCP).
+
+See [BENCHMARKS.md](BENCHMARKS.md) for measured TTFT / ITL / throughput, and
+[OBSERVABILITY.md](OBSERVABILITY.md) for wiring Dynamo metrics into the HyperPod Grafana.
+
+## Prerequisites
+
+| Requirement | Detail |
+|---|---|
+| HyperPod EKS cluster | Two instance groups (see below); `kubectl` configured, `helm` v3+ |
+| Control nodes | 2x `ml.c6i.xlarge` — platform components (operator, etcd, NATS) |
+| GPU nodes | 2x `ml.g6e.4xlarge` (L40S 48GB), instance group named **`dynamo-workers`** |
+| Shared filesystem | FSx for Lustre (model weights, mounted at `/fsx`) |
+| Hugging Face token | **Not required** — both models are public. Only needed for gated models. |
+
+> The GPU instance group **must** be named `dynamo-workers` — every scenario YAML uses
+> this label for scheduling. For production disaggregation, use EFA-enabled instances
+> (e.g. p5) so NIXL uses RDMA instead of TCP.
+
+## 0. Configure environment
+
+```bash
+cp env_vars.example env_vars
+# edit env_vars: AWS_REGION, AWS_ACCOUNT_ID, EKS_CLUSTER_NAME (defaults are validated values)
+source env_vars
+aws eks update-kubeconfig --name "$EKS_CLUSTER_NAME" --region "$AWS_REGION"
+kubectl get nodes --show-labels | grep instance-group-name
+```
+
+## 1. Install the Dynamo platform
+
+Installs the EBS CSI driver (with HyperPod volume permissions), the Dynamo CRDs, and the
+platform — **operator** on the control nodes, **etcd + NATS** on the GPU workers (the
+control nodes' pod cap can't host them). etcd + NATS provide the cross-node discovery the
+operator wires into every scenario.
+
+```bash
+./scripts/01-install-platform.sh
+# optional: send Dynamo metrics to the HyperPod Grafana dashboard (see OBSERVABILITY.md)
+./scripts/01a-setup-observability.sh
+```
+Verify: `kubectl get pods -n dynamo-system` → operator, etcd, nats all `Running`.
+
+## 2. Download a model
+
+One-time download to FSx Lustre, shared across pods. Public models — no token needed.
+
+```bash
+./scripts/02-download-model.sh gpt-oss     # -> /fsx/models/openai-gpt-oss-20b
+./scripts/02-download-model.sh qwen3.6     # -> /fsx/models/qwen3.6-27b-fp8
+```
+
+## 3. Deploy a scenario
+
+```bash
+./scripts/03-deploy-gpt-oss-agg.sh
+./scripts/04-deploy-gpt-oss-disagg.sh
+./scripts/05-deploy-qwen3.6-agg.sh
+./scripts/06-deploy-qwen3.6-disagg.sh
+```
+
+Each scenario is a `DynamoGraphDeployment` (CRD) reconciled by the Dynamo operator into
+the frontend + worker pods. Workers load the model from FSx (~2 min). Watch:
+```bash
+kubectl get dynamographdeployment -n dynamo-system
+kubectl get pods -n dynamo-system -l nvidia.com/dynamo-graph-deployment-name=<name> -w
+```
+> DGD names are dot-free: `gpt-oss-agg`, `gpt-oss-disagg`, `qwen36-agg`, `qwen36-disagg`.
+
+> **GPU budget:** each aggregated scenario uses 1 GPU; each disaggregated scenario uses 2.
+> With 2 GPU workers you can run two aggregated scenarios at once, or one disaggregated.
+
+## 4. Inference
+
+```bash
+./scripts/07-test-inference.sh     # one-shot: models + a chat (auto-detects the deployed scenario)
+./scripts/07b-chat.sh              # interactive multi-turn streaming chat (auto-detects)
+```
+Pass a name (`gpt-oss-agg` | `gpt-oss-disagg` | `qwen36-agg` | `qwen36-disagg`) to override auto-detect.
+
+Or connect directly (OpenAI-compatible) via port-forward to the operator-created frontend
+service `<name>-frontend`:
+```bash
+kubectl port-forward -n dynamo-system svc/<name>-frontend 8000:8000
+curl localhost:8000/v1/chat/completions -H 'Content-Type: application/json' \
+  -d '{"model":"<id>","messages":[{"role":"user","content":"Hello"}],"max_tokens":200}'
+```
+
+## 5. Benchmark
+
+AIPerf concurrency sweep, saved to `bench_results/<scenario>/`:
+```bash
+./scripts/08-benchmark.sh <scenario>
+# custom: ./scripts/08-benchmark.sh qwen36-agg 2000 256
+# custom: CONCURRENCIES="1,5,10,50" ./scripts/08-benchmark.sh gpt-oss-agg
+```
+
+## 6. Cleanup
+
+```bash
+./scripts/09-cleanup-inference.sh <scenario>   # or: all
+./scripts/10-uninstall-platform.sh             # remove the platform (optional)
+```
+
+## Architecture notes
+
+### HyperPod specifics
+- Node labels use `sagemaker.amazonaws.com/instance-group-name` (not `node-group-name`).
+- EBS CSI needs `sagemaker:AttachClusterNodeVolume` / `DetachClusterNodeVolume` (handled by `01`).
+- Model weights live on shared FSx Lustre — no repeated downloads on pod restarts.
+- The SGLang runtime image requires the NVIDIA runtime, so frontend pods also run on GPU
+  nodes (without requesting a GPU).
+
+### Per-model notes
+- **GPT-OSS-20B**: MXFP4 weights; on L40S (Ada) it requires the `triton` attention backend.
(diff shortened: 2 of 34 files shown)


matheus-1618:
@paragao can you review this PR?

It adds a **NVIDIA Dynamo (SGLang) LLM inference** test case on SageMaker HyperPod EKS — **GPT-OSS-20B** and **Qwen3.6-27B-FP8**, each in **aggregated** and **disaggregated** (prefill/decode split + NIXL KV transfer) flavors, via the Dynamo operator and the `DynamoGraphDeployment` CRD. It complements the vLLM inference case (#1111) by adding Dynamo's disaggregated serving.

Validated end-to-end on 2× `ml.g6e.4xlarge` (NVIDIA L40S 48GB); all images are pinned (no `:latest`), every script/manifest has the SPDX/MIT-0 header, and benchmarks are in `BENCHMARKS.md`. Let me know what you'd like changed.

KeitaW (commented):
## Review Batch 1/4 — Structure, Deployment Pipeline & Operational Correctness

Thanks for this — a genuinely thorough, self-contained Dynamo test case (the Troubleshooting table especially is exemplary). My notes below are mostly consistency/portability; nothing here is a correctness blocker. Severity is in the prose, and I've left a few inline `suggestion` blocks for one-click fixes.

### Test case doesn't follow the repo's directory-layout standard
**Observation:** The repo's documented layout (CONTRIBUTING.md + the root README "Test cases" structure) is `3.test_cases/<framework>/<library>/<model>/` with deployment artifacts collected under a manifests subdirectory rather than scattered as top-level folders (the sibling `pytorch/vllm/cosmos-reason` keeps its under `hyperpod-eks/` + `kubernetes/`). This PR instead places top-level `scenarios/`, `scripts/`, and `platform/` folders directly under `pytorch/nvidia-dynamo/`.
**Impact (should-fix):** Inconsistent with every other test case — harder for users (and the scaffolding tooling / root index tables) to locate the deployment path.
**Suggestion:** Add a `manifests/` directory and move the deployment manifests there (scenario DGDs, platform values, observability). Keep them **platform-agnostic** — the same manifests should work on HyperPod EKS *and* standard EKS (see the node-selector note on `platform/values.yaml`) — so a generic `manifests/` home fits better than a HyperPod-specific `hyperpod-eks/` directory. Since Dynamo targets Kubernetes only, you also don't need the `slurm/`-vs-`kubernetes/` split. Keep the `*.md` docs at the case root.

### Kubernetes objects are embedded as heredocs inside shell scripts
**Observation:** Several first-class objects are generated as inline `cat <<EOF | kubectl apply -f -` heredocs rather than committed YAML: the FSx `PersistentVolume` + `PersistentVolumeClaim` and the model-download `Job` in `02-download-model.sh`, the benchmark `Job` in `08-benchmark.sh`, and the `dynamo-metrics` `Service` in `01a-setup-observability.sh`. The four scenario manifests, by contrast, are proper standalone YAML — the pattern to follow.
**Impact (should-fix):** Heredoc'd manifests aren't reviewable as manifests (no schema validation / `kubectl --dry-run` / kubeconform in CI), shell interpolation makes them easy to break silently, and they hide real resources (a cluster-scoped PV, two Jobs) from anyone reading the tree.
**Suggestion:** Move each into its own YAML under the manifests tree (e.g. `manifests/fsx-pv.yaml`, `jobs/download-model.yaml`, `jobs/benchmark.yaml`, `scenarios/observability/metrics-service.yaml`) and `envsubst` + `kubectl apply -f` them with an explicit allow-list (`envsubst '$MODEL_ID $MODEL_PATH $NS'`).

3.test_cases/pytorch/nvidia-dynamo/scripts/07-test-inference.sh
```diff
@@ -0,0 +1,33 @@
...
+  elif [ -t 0 ]; then echo "Multiple deployed — pick one:"; select s in $DET; do [ -n "$s" ] && { NAME="$s"; break; }; done
+  else echo "❌ Multiple deployed ($(echo $DET)). Pass one as arg."; exit 1; fi
+fi
+case "$NAME" in
+  gpt-oss-*) MODEL="openai/gpt-oss-20b" ;;
+  qwen36-*)  MODEL="Qwen/Qwen3.6-27B-FP8" ;;
+  *) echo "Unknown scenario: $NAME"; exit 1 ;;
+esac
+
+POD=$(kubectl get pod -l "nvidia.com/dynamo-graph-deployment-name=${NAME},nvidia.com/dynamo-component=Frontend" -n "$NS" -o jsonpath='{.items[0].metadata.name}')
```
KeitaW, line 24: ### Frontend pod selector — correction: verified working, only a cosmetic inconsistency

**Update — I tested this live** on the pinned `0.7.0` operator (deployed on an EKS cluster) and my original concern was **wrong**, so I'm walking it back. The Dynamo operator stamps **both** labels on the frontend pod:

- `nvidia.com/dynamo-component=Frontend` (capitalized — what this script uses)
- `nvidia.com/dynamo-component-type=frontend` (lowercase — what the Service/PodMonitor use)

Both selectors resolve to the frontend pod, and `07-test-inference.sh` runs end to end (the `/v1/models` list and a chat completion both succeed via this exact selector). So this line is **not** broken — your testing was right.

The only thing left is cosmetic: the PR uses two different label *keys* for the same concept across files. Standardizing on one would be marginally cleaner, but it's entirely optional since both resolve. Apologies for the false alarm on this one.

3.test_cases/pytorch/nvidia-dynamo/manifests/platform/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+#
+# NOTE: the 0.7.0 chart pins a kube-rbac-proxy sidecar image that no longer exists
+# (gcr.io/kubebuilder/...). 01-install-platform.sh patches it to the quay.io/brancz
+# mirror right after install (the image tag isn't exposed as a Helm value).
+
+dynamo-operator:
+  enabled: true
+  controllerManager:
+    nodeSelector:
+      sagemaker.amazonaws.com/instance-group-name: dynamo-control
```
KeitaW, line 28: ### Node selectors are HyperPod-only — consider supporting plain EKS too

**Observation:** This (and all four scenario manifests) pins scheduling to `sagemaker.amazonaws.com/instance-group-name: dynamo-control` / `dynamo-workers`. That label is injected only on **SageMaker HyperPod** nodes; a standard (non-HyperPod) EKS cluster has no such label, so every pod stays `Pending` with no obvious cause.
**Impact (should-fix):** The README presents HyperPod as the only target, but the workload itself isn't HyperPod-specific.
**Suggestion:** Select on the standard `node.kubernetes.io/instance-type` label, present on both platforms — its value just carries the `ml.` prefix on HyperPod (`ml.g6e.4xlarge`) vs the bare EC2 type on standard EKS (`g6e.4xlarge`). A single `nodeAffinity` listing both values works everywhere with no per-cluster edit:

```yaml
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
        - matchExpressions:
            - key: node.kubernetes.io/instance-type
              operator: In
              values: ["ml.g6e.4xlarge", "g6e.4xlarge"]   # HyperPod / standard EKS
```

(`nodeSelector` only does exact single-value match, so it can't list both — `nodeAffinity … In` is what covers both prefixes.) This also drops the hard dependency on the `dynamo-workers` / `dynamo-control` instance-group names.

3.test_cases/pytorch/nvidia-dynamo/scripts/02-download-model.sh
```diff
@@ -0,0 +1,95 @@
...
+      nodeSelector: { sagemaker.amazonaws.com/instance-group-name: dynamo-workers }
+      tolerations: [{ effect: NoSchedule, key: nvidia.com/gpu, operator: Exists }]
+      containers:
+        - name: download
+          image: python:3.11-slim
+          command: ["/bin/bash","-c"]
+          args:
+            - |
+              set -e
+              pip install -q "huggingface_hub[hf_transfer]"; export HF_HUB_ENABLE_HF_TRANSFER=1
```
KeitaW: ### Unpinned `huggingface_hub` installed at job runtime

**Observation:** The download Job runs `pip install -q "huggingface_hub[hf_transfer]"` (unpinned) inside a stock `python:3.11-slim` image at run time.
**Impact (should-fix):** Non-reproducible — a future `huggingface_hub` release can change `snapshot_download` behavior or break the job — and it requires worker egress to PyPI at run time.
**Suggestion:** Pin the version you validated, e.g. `huggingface_hub[hf_transfer]==<x.y.z>`. A pinned one-shot install is the proportionate fix here — baking a dedicated image would be heavier than this job warrants.

3.test_cases/pytorch/nvidia-dynamo/env_vars.example
```diff
@@ -0,0 +1,31 @@
...
+# ---- AWS / cluster context ----
+export AWS_REGION="us-west-2"
+export AWS_ACCOUNT_ID="123456789012"
+export EKS_CLUSTER_NAME="<your-hyperpod-eks-cluster>"
+export NAMESPACE="dynamo-system"
+
+# ---- Serving image (pinned; do NOT use :latest) ----
+# SGLang runtime that backs the Dynamo workers and frontend. Pinned in every
+# scenario YAML (the `image:` field) — keep this in sync if you bump it there.
+export DYNAMO_IMAGE="nvcr.io/nvidia/ai-dynamo/sglang-runtime:1.2.1"
```
KeitaW: ### `DYNAMO_IMAGE` is exported but never consumed (the version split itself is fine)

**Update — I verified the image split is legitimate, so narrowing this.** I checked the serving image and the `benchmarks` harness module is **not** present in `sglang-runtime:1.2.1` — it lives in the `0.7.0` image that `08-benchmark.sh` uses. So the `0.7.0` (benchmark) vs `1.2.1` (serving) split is deliberate and correct, and your `08-benchmark.sh` comment ("benchmark harness lives here") already documents it. No change needed there.

The one real (minor) residue: this `DYNAMO_IMAGE` export is read by no script or manifest (the YAMLs hardcode the tag), yet its comment says "keep this in sync if you bump it there" — which can't be honored since nothing consumes the variable. Either wire `DYNAMO_IMAGE` through, or drop the export + the "keep in sync" note to avoid the dead-config trap. Cosmetic.

KeitaW (commented):
## Review Batch 2/4 — Versions, Pinning & Reproducibility

Pinning is otherwise excellent across this PR (no `:latest` anywhere, even the benchmark image) — two small consistency notes.

3.test_cases/pytorch/nvidia-dynamo/scripts/01-install-platform.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+# Install Dynamo 0.7.0 Platform on HyperPod EKS
+# Prerequisites: kubectl configured, helm installed
+set -e
```
KeitaW: ### Use `set -euo pipefail` (all 12 scripts use bare `set -e`)

**Observation:** Every script (`01`–`10`, `01a`, `07b`) opens with `set -e` only; the repo convention is `set -euo pipefail`.
**Impact (should-fix):** Unset variables don't error (an unsourced `env_vars` leaves region/cluster vars empty and the script proceeds), and a failure mid-pipe is masked — e.g. the region/cluster `sed`-pipelines here swallow upstream failures.

```suggestion
set -euo pipefail
```

The scripts already guard positionals with `${1:?}`/`${1:-}` and env with `${VAR:-default}`, so `-u` should be safe — worth a quick pass to confirm.

KeitaW (commented):
## Review Batch 3/4 — Shell Robustness & Documentation Consistency

3.test_cases/pytorch/nvidia-dynamo/OBSERVABILITY.md
```diff
@@ -0,0 +1,83 @@
...
+| `dynamo_frontend_time_to_first_token_seconds` | Histogram | TTFT distribution |
+| `dynamo_frontend_inter_token_latency_seconds` | Histogram | Inter-token latency (TPOT) distribution |
+| `dynamo_frontend_request_duration_seconds` | Histogram | Total request duration distribution |
+
+### Tips
+
+- Use `rate(...[5m])` for counters and histograms to see per-second rates
+- Use `histogram_quantile(0.50, ...)` for p50, `0.99` for p99
+- Filter by model: add `{model="openai/gpt-oss-20b"}` to any query
+- Run a benchmark (`./scripts/05a-test-benchmark.sh`) to generate traffic and see metrics in action
```
KeitaW: ### Points at a script that doesn't exist

**Observation:** There is no `05a-test-benchmark.sh` — the benchmark script is `08-benchmark.sh`. (This is the only dangling script reference; every other doc/script path resolves.)
**Impact (should-fix):** Dead reference on the documented observability path.

```suggestion
- Run a benchmark (`./scripts/08-benchmark.sh <scenario>`) to generate traffic and see metrics in action
```

3.test_cases/pytorch/nvidia-dynamo/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+See [BENCHMARKS.md](BENCHMARKS.md) for measured TTFT / ITL / throughput, and
+[OBSERVABILITY.md](OBSERVABILITY.md) for wiring Dynamo metrics into the HyperPod Grafana.
+
+## Prerequisites
+
+| Requirement | Detail |
+|---|---|
+| HyperPod EKS cluster | Two instance groups (see below); `kubectl` configured, `helm` v3+ |
+| Control nodes | 2x `ml.c6i.xlarge` — platform components (operator, etcd, NATS) |
```
KeitaW: ### Prereq table contradicts the install section on etcd/NATS placement

**Observation:** This row says the control nodes host "platform components (operator, etcd, NATS)". But the install section just below — and `platform/values.yaml` — place **etcd + NATS on the GPU workers** ("the control nodes' pod cap can't host them"), with only the operator on control.
**Impact (should-fix):** Directly contradictory; this row is wrong and will mislead capacity planning.

```suggestion
| Control nodes | 2x `ml.c6i.xlarge` — platform operator (etcd + NATS run on the GPU workers) |
```

KeitaW: ### `dynamo-control` instance-group name is mandatory but undocumented

**Observation:** `platform/values.yaml` hard-pins the operator to `sagemaker.amazonaws.com/instance-group-name: dynamo-control`. The README is emphatic that the GPU group "**must** be named `dynamo-workers`", but never states that the control group must be named `dynamo-control`.
**Impact (should-fix):** A user who names their control group anything else gets the operator pod stuck `Pending` with no warning from the prereqs.
**Suggestion:** Add to the prereqs/callout, alongside the `dynamo-workers` requirement: "The control instance group **must** be named `dynamo-control`." (Or template both names so they aren't load-bearing magic strings.)

KeitaW (commented):
## Review Batch 4/4 — Nits, What Looks Great & Sources

### Nits
- **`--trust-remote-code` on every worker without a one-line rationale.** It's legitimately required for these models, but the repo convention is a short "why this is needed/trusted" comment next to it.
- **PodMonitor scrapes a `system` port that may not exist.** `pod-monitor.yaml` scrapes named ports `http` and `system`; the scenario YAMLs name `http` only on the Frontend and define `system` nowhere. If the operator doesn't inject a `system` metrics port, that endpoint silently no-ops. Worth a quick confirm; low priority.
- **`select s in $DET` / `$(echo $DET)` rely on unquoted word-splitting.** Fine today (DGD names are whitespace-free), but brittle once `set -u`/`pipefail` lands. Optional.

### Things that look great
- **Everything is pinned** — serving image `sglang-runtime:1.2.1`, platform chart `0.7.0`, even the rbac-proxy mirror `quay.io/brancz/kube-rbac-proxy:v0.15.0`. No `:latest` anywhere. This is the single most-violated rule in this repo and the PR gets it right.
- **MIT-0 SPDX headers on every script and manifest**, including the `.yaml.example` files.
- **The Troubleshooting table is exemplary** — each row is a real bring-up failure with symptom → cause → fix-already-applied (dot-free DGD names for the `dyn://` endpoint, `--host 0.0.0.0` for the NIXL bootstrap bind, the Qwen disagg `--max-running-requests 16` + `--mem-fraction-static 0.80` OOM cap, the rbac-proxy image patch). The disagg manifest's caps match its own note exactly.
- **Secrets handled correctly** — HF token via a `kubectl` Secret (not plaintext in a manifest), `*.example` files committed with "do not commit the filled-in version", and an honest "token not required — models are public" callout.
- **FSx PV uses `persistentVolumeReclaimPolicy: Retain`** with explicit stale-`claimRef` clearing on reuse — no accidental data loss on namespace delete.
- **Benchmarks reported honestly** — real AIPerf tables with raw TTFT/ITL/throughput per concurrency, ISL/OSL stated, and the TCP-not-RDMA (no-EFA-on-g6e) caveat called out rather than buried.

### Sources
- Dynamo K8s metrics / labels: https://docs.nvidia.com/dynamo/latest/guides/dynamo_deploy/k8s_metrics.html
- Dynamo Kubernetes operator docs: https://docs.nvidia.com/dynamo/latest/kubernetes/README.html
- Repo conventions: https://github.com/awslabs/awsome-distributed-ai/blob/main/CONTRIBUTING.md
- Sibling layout precedent: `3.test_cases/pytorch/vllm/cosmos-reason` (`hyperpod-eks/` + `kubernetes/`)

KeitaW (changes requested):
Few comments

matheus-1618:
> Few comments

Thanks @KeitaW , working on it.

matheus-1618:
Hello @KeitaW. Addressed.

3.test_cases/pytorch/nvidia-dynamo/manifests/jobs/download-model.yaml
```diff
@@ -0,0 +1,28 @@
...
+#   envsubst '$JOB $NS $MODEL_ID $MODEL_PATH $PVC' | kubectl apply -f -
+apiVersion: batch/v1
+kind: Job
+metadata: { name: ${JOB}, namespace: ${NS} }
+spec:
+  backoffLimit: 2
+  template:
+    spec:
+      restartPolicy: Never
+      nodeSelector: { sagemaker.amazonaws.com/instance-group-name: dynamo-workers }
```
KeitaW: ### The node-selector portability fix didn't reach the two Jobs

**Observation:** The four scenario manifests got the portable `nodeAffinity`, but this download Job (and `manifests/jobs/benchmark.yaml`) still pins `nodeSelector: { sagemaker.amazonaws.com/instance-group-name: dynamo-workers }` — the HyperPod-only label.
**Impact:** On a standard (non-HyperPod) EKS cluster the download and benchmark Jobs stay `Pending` (no node carries that label), even though the scenarios schedule fine — the same gap the scenarios just fixed, left on the two Jobs. (Keeping these Jobs on the GPU nodes is correct, by the way — the download Job needs a node that can mount FSx Lustre, which the GPU nodes can.)
**Suggestion:** Apply the same `nodeAffinity` you used in the scenarios:

```suggestion
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: node.kubernetes.io/instance-type
                    operator: In
                    values: ["ml.g6e.4xlarge", "g6e.4xlarge"]   # HyperPod (ml.) / standard EKS
```

3.test_cases/pytorch/nvidia-dynamo/manifests/jobs/benchmark.yaml
```diff
@@ -0,0 +1,31 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata: { name: ${JOB}, namespace: ${NS} }
+spec:
+  backoffLimit: 0
+  ttlSecondsAfterFinished: 1800
+  template:
+    spec:
+      restartPolicy: Never
+      nodeSelector: { sagemaker.amazonaws.com/instance-group-name: dynamo-workers }
```
KeitaW: ### Same fix here on the benchmark Job

**Observation:** `manifests/jobs/benchmark.yaml` has the same HyperPod-only `nodeSelector` as the download Job.
**Impact:** The AIPerf benchmark Job won't schedule on a standard-EKS cluster.
**Suggestion:**

```suggestion
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: node.kubernetes.io/instance-type
                    operator: In
                    values: ["ml.g6e.4xlarge", "g6e.4xlarge"]   # HyperPod (ml.) / standard EKS
```

KeitaW (commented):
## Review Batch 1/2 — Round 2: Resolution & Follow-up

Thanks for the thorough turnaround (commits `7d1b896` + `e91d185`) — this addressed essentially everything, and the `manifests/` restructure is cleanly done. Scoreboard:

**Resolved ✅**
- **Directory layout** → all deploy artifacts now live under a single platform-agnostic `manifests/{fsx,jobs,observability,platform,scenarios}/` tree. Exactly right.
- **Node selectors** → portable `nodeAffinity` on `node.kubernetes.io/instance-type` matching both `ml.g6e.4xlarge` (HyperPod) and `g6e.4xlarge` (standard EKS), on all four scenarios.
- **Heredoc'd manifests** → extracted to templated YAML (`fsx-pv`, `fsx-pvc`, `download-model`, `benchmark`, `metrics-service`) applied with `envsubst` + explicit allow-lists; FSx discovery stays in the script. This is the pattern I was hoping for.
- **Runtime pip** → `huggingface_hub[hf_transfer]==1.20.1` (verified — that's the current latest on PyPI, a valid resolvable pin).
- **`set -euo pipefail`** → adopted across all 12 scripts.
- **OBSERVABILITY.md** → now points at `08-benchmark.sh`.
- **README etcd/NATS placement** → table corrected to "platform operator (etcd + NATS run on the GPU workers)".
- **`dynamo-control` prereq** → now documented as a required instance-group name.

**Previously retracted / narrowed (no action needed):** the `07-test-inference.sh` label selector (I verified live that it works — the operator stamps both label keys) and the benchmark image skew (the `0.7.0` harness image is legitimate).

One follow-up below — the node-selector portability fix didn't reach the two Jobs.

KeitaW (commented):
## Review Batch 2/2 — Nit & Close

### Nit
- **`DYNAMO_IMAGE` is still exported in `env_vars.example` but consumed by nothing** (the manifests hardcode the tag). You dropped the misleading "keep in sync" note — thanks — so it's now harmless; you could remove the dead export entirely for tidiness. Cosmetic.

### Close
Nice work — once the two Jobs get the same `nodeAffinity`, the whole test case is portable HyperPod ↔ standard EKS. Everything else from round 1 is resolved.

For the record, I deployed the scenarios end-to-end on a live EKS cluster while reviewing — both gpt-oss flavors (aggregated and disaggregated / NIXL) serve correctly. I ran on Blackwell GPUs rather than your L40S target; the qwen3.6 FP8 path hit a kernel gap specific to that newer hardware, unrelated to your manifests — nothing that affects your validated L40S setup.

### Sources
- `huggingface_hub` latest version (verified `1.20.1`, 2026-06-24): https://pypi.org/project/huggingface-hub/
- Prior round: https://github.com/awslabs/awsome-distributed-ai/pull/1155#pullrequestreview-4562375050

matheus-1618:
@KeitaW can you review again?

KeitaW (approved):
LGTM, thank you!

PR #1158, opened 2026-06-25
DeepEP - Simplify DeepEP on EFA using a script

erezzarum:
## Purpose

The script simplify installation of DeepEP on EFA by
1. Installing NVSHMEM from NVIDIA upstream
2. Cloning DeepEP at pre-v2 commit
3. Patching DeepEP to work with AWS EFA and NVHSMEM 3.7
4. Installing DeepEP

## Changes

- Removed patch
- Removed NVSHMEM installation step
- Added the `setup_deepep_efa.sh` script to simplify installation.

## Test Plan

**Environment:**
- AWS Service: EKS
- Instance type: p6-b200.48xlarge
- Number of nodes: 2

**Test commands:**
```bash
export IMAGE_URI=<HIDDEN>
export INSTANCE_TYPE=p6-b200.48xlarge
export GPU_PER_NODE=8
export EFA_PER_NODE=8
export NUM_NODES=2

envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE' < test-intranode.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE' < test-intranode.yaml | kubectl delete -f -

envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-low-latency.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-low-latency.yaml | kubectl delete -f -

envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-internode.yaml | kubectl apply -f -
envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' < test-internode.yaml | kubectl delete -f -
```

## Directory Structure

```
5.test_cases/
```

## Checklist

### DIFF
--- a/micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
+++ b/micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
@@ -145,12 +145,13 @@ RUN CUDA_VER=$(nvcc --version | grep -oP 'release \K[0-9]+\.[0-9]+' | tr -d '.')
 ARG DEEPEP_PREFIX="/DeepEP"
 
 RUN --mount=type=bind,source=setup_deepep_efa.sh,target=/tmp/setup_deepep_efa.sh \
-    ./tmp/setup_deepep_efa.sh --deepep-prefix "${DEEPEP_PREFIX}" && \
+    /tmp/setup_deepep_efa.sh --deepep-prefix "${DEEPEP_PREFIX}" && \
     pip3 uninstall -y nvidia-nvshmem-cu13 nvidia-nvshmem-cu12 nvidia-nvshmem
 
 ENV LD_LIBRARY_PATH="/opt/amazon/nvshmem/lib:${LD_LIBRARY_PATH}"
 ENV PATH="/opt/amazon/nvshmem/bin:${PATH}"
 
+ENV FI_PROVIDER=efa
 ENV NVSHMEM_REMOTE_TRANSPORT=libfabric
 ENV NVSHMEM_LIBFABRIC_PROVIDER=efa
 ENV NVSHMEM_NETDEVS_POLICY=EXTERNAL_SHARING_PCIE_SWITCH_NIC_EXCLUSIVE
--- a/micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
+++ b/micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
@@ -10,12 +10,7 @@ ARG GDRCOPY_VERSION=v2.5.2
 ARG EFA_INSTALLER_VERSION=1.48.0
 ARG NCCL_VERSION=v2.30.4-1
 ARG NCCL_TESTS_VERSION=v2.18.3
-ARG NVSHMEM_VERSION=3.7.0
 ARG TORCH_VERSION=2.11.0
-# NOTE: deepep_aws_efa.patch is generated against this exact commit. If you bump
-# DEEPEP_COMMIT, regenerate the patch too -- `git apply` is exact and the build
-# will fail on a mismatch.
-ARG DEEPEP_COMMIT=567632d
 
 # CUDA architecture(s), semicolon-separated:
 # 9.0 = Hopper (H100, sm_90), 10.0 and 10.3 = Blackwell (B200/B300, sm_100,sm_103). Defaults to
@@ -126,27 +121,6 @@ RUN git clone -b ${NCCL_TESTS_VERSION} https://github.com/NVIDIA/nccl-tests.git 
     CUDA_HOME="${CUDA_HOME}" \
     NCCL_HOME=/opt/nccl/build
 
-###################################################
-## Install NVSHMEM
-ARG NVSHMEM_PREFIX="/opt/nvshmem"
-ARG LIBFABRIC_HOME="/opt/amazon/efa"
-
-RUN CUDA_VERSION_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') && \
-    if [ "${TARGETARCH}" = "arm64" ]; then NVSHMEM_ARCH="linux-sbsa"; else NVSHMEM_ARCH="linux-x86_64"; fi && \
-    NVSHMEM_ARCHIVE="libnvshmem-${NVSHMEM_ARCH}-${NVSHMEM_VERSION}_cuda${CUDA_VERSION_MAJOR}-archive.tar.xz" && \
-    NVSHMEM_DOWNLOAD_URL="https://developer.download.nvidia.com/compute/nvshmem/redist/libnvshmem/${NVSHMEM_ARCH}/${NVSHMEM_ARCHIVE}" && \
-    curl -fSL "${NVSHMEM_DOWNLOAD_URL}" -o "/tmp/${NVSHMEM_ARCHIVE}" && \
-    mkdir -p "$NVSHMEM_PREFIX" && \
-    tar -xf "/tmp/${NVSHMEM_ARCHIVE}" --strip-components=1 -C "$NVSHMEM_PREFIX" && \
-    rm "/tmp/${NVSHMEM_ARCHIVE}"
-
-ENV LD_LIBRARY_PATH="${NVSHMEM_PREFIX}/lib:${LD_LIBRARY_PATH}"
-ENV PATH="${NVSHMEM_PREFIX}/bin:${PATH}"
-
-ENV NVSHMEM_REMOTE_TRANSPORT=libfabric
-ENV NVSHMEM_LIBFABRIC_PROVIDER=efa
-ENV NVSHMEM_NETDEVS_POLICY=EXTERNAL_SHARING_PCIE_SWITCH_NIC_EXCLUSIVE
-
 RUN rm -rf /var/lib/apt/lists/*
 
 ## Set Open MPI variables to exclude network interface and conduit.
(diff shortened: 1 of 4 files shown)


micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
```diff
@@ -165,24 +139,20 @@ ENV LD_PRELOAD=/opt/nccl/build/lib/libnccl.so
...
-    git checkout ${DEEPEP_COMMIT} && \
-    git apply --check /tmp/deepep_aws/deepep_aws_efa.patch && \
-    git apply /tmp/deepep_aws/deepep_aws_efa.patch && \
-    CUDA_VERSION_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') && \
-    if [ "${CUDA_VERSION_MAJOR}" -ge 13 ]; then \
-        sed -i "s|f'{nvshmem_dir}/include']|f'{nvshmem_dir}/include', '${CUDA_HOME}/include/cccl']|" "setup.py"; \
-    fi && \
-    NVSHMEM_DIR="${NVSHMEM_PREFIX}" pip3 install -vv --no-build-isolation .
+RUN --mount=type=bind,source=setup_deepep_efa.sh,target=/tmp/setup_deepep_efa.sh \
+    ./tmp/setup_deepep_efa.sh --deepep-prefix "${DEEPEP_PREFIX}" && \
```
KeitaW: ### Documented build-args dropped, and a relative path that works only by luck

Two issues land on this `RUN` — one should-fix, one nit.

**1. The README's `--build-arg NVSHMEM_VERSION` / `--build-arg DEEPEP_COMMIT` are now silently ignored.** This PR removes both `ARG` declarations from the Dockerfile, and this `RUN` forwards only `--deepep-prefix` to the script — but the README's build command still passes `--build-arg="NVSHMEM_VERSION=…"` and `--build-arg="DEEPEP_COMMIT=…"`. With the ARGs gone, BuildKit prints *"one or more build-args … were not consumed"* and drops them. And re-adding the ARGs alone wouldn't help: `parse_args` assigns `NVSHMEM_VERSION="3.7.0"` and `DEEPEP_COMMIT="567632d"` as unconditional literals (not `${VAR-default}` env reads), so they only move via the `--nvshmem-version` / `--deepep-commit` flags — unlike `TORCH_CUDA_ARCH_LIST`, which the script reads from the env. The default build still succeeds (defaults match the docs), so this isn't build-breaking — but a user bumping either value per the README silently gets the old one.

The two knobs aren't symmetric, so the fix differs:

- **`NVSHMEM_VERSION` is genuinely variable** (`resolve_nvshmem_archive` looks it up in NVIDIA's manifest; the README says "≥ 3.7.0"). Keep it overridable: re-declare `ARG NVSHMEM_VERSION=3.7.0` near the other version ARGs (~line 13) and forward it as `--nvshmem-version "${NVSHMEM_VERSION}"`.
- **`DEEPEP_COMMIT` is effectively a fixed pin, not a knob** — `commit_compat_gate` aborts on any commit but `567632d` unless `--force`, and I confirmed the embedded patch only `git apply`s cleanly at that commit (it fails at/after the EPv2 release `b306af0`, which restructured/removed the patched files). So drop `DEEPEP_COMMIT` from the README's build command and note the commit is pinned by the patch.

(Docker's unconsumed-build-arg behavior — warn, not fail — is documented at https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg)

**2. `./tmp/setup_deepep_efa.sh` is a relative path; the mount target is the absolute `/tmp/setup_deepep_efa.sh`.** It resolves today only because the Dockerfile sets no `WORKDIR` (default cwd is `/`, so `./tmp/…` equals `/tmp/…`). Adding any `WORKDIR` above this line would silently break it. The one-click fix below switches to the absolute path; the `--nvshmem-version` change above can be folded in on top.

```suggestion
    /tmp/setup_deepep_efa.sh --deepep-prefix "${DEEPEP_PREFIX}" && \
```

KeitaW (commented):
## Review Batch 1/3 — Build Configuration & Reproducibility

Thanks for this — replacing the checked-in `.patch` with a self-contained, validated installer is a real improvement, and I verified the whole thing end-to-end (see Batch 3). One inline note below on the DeepEP `RUN`: a should-fix (documented `--build-arg`s silently dropped after the ARG removal) and a nit (relative `./tmp/` invocation path).

KeitaW (commented):

### The README's version constraint `<= 567632d` overstates the supported range — pin the exact commit

The "⚠️ Version constraints" block reads "DeepEP v1 with the NVSHMEM backend **`<= 567632d`**". The `<=` implies any earlier DeepEP v1 commit is acceptable, and combined with the `--deepep-commit` flag / `DEEPEP_COMMIT` build-arg it reads as "anything up to this commit works." It doesn't — the embedded patch matches one specific file context, so the working range is a narrow band, not an open-ended `<=`. I verified each end with `git apply --check` against a fresh DeepEP clone:

| Commit | Date | `git apply --check` |
|--------|------|------|
| `92fe2de` and earlier v1 | ≤ 2025-11-05 | **fails** |
| `29d31c0` (last commit to touch the 4 patched files) | 2026-01-20 | applies |
| **`567632d`** (the pin) | 2026-02-03 | applies — and builds + all 3 benchmarks pass on B300 |
| `b306af0` EPv2 onward | ≥ 2026-04-26 | **fails** (files restructured/removed) |

So older v1 commits that the `<=` invites **fail to patch**, and the script's own `commit_compat_gate` hard-pins to exactly `567632d` anyway. I'd suggest stating the exact requirement instead of a range:

```markdown
> This works only with **NVSHMEM ≥ 3.7.0** and DeepEP at **exactly commit
> [`567632d`](https://github.com/deepseek-ai/DeepEP/tree/567632dd59810d77b3cc05553df953cc0f779799)**
> (pre-EPv2). The embedded EFA patch is validated only against this commit; the
> build aborts on any other. EPv2 restructures the kernel sources and is out of scope.
```

This is the doc-side complement to the `DEEPEP_COMMIT`-isn't-a-knob point in Batch 1 — same root cause, two surfaces.

### Pre-existing documentation drift (informational — not introduced by this PR, not blocking)

These predate this PR (the README diff only touches the "How the EFA support works" bullets, not the build-command/prereqs region), so none are regressions. But since the PR reworks this build surface, they're cheap to fix in the same pass:

- **`TORCH_CUDA_ARCH_LIST="9.0;10.0"` in the README build example drops B300/`sm_103`.** The Dockerfile `ARG` default is `9.0;10.0;10.3` and the prose advertises B300, but copying the README verbatim builds *without* `sm_103` codegen. Unlike the two dead build-args above, this one **does** take effect (the ARG is still declared), so it silently produces a non-B300 image. Align the example to `9.0;10.0;10.3`.
- **`--build-arg GPU_ARCH=…` is documented but never declared** (README prerequisites + build command). The Dockerfile has no `GPU_ARCH` ARG, so it's a no-op + BuildKit warning. Drop it from the README or wire it in.
- **`FI_PROVIDER=efa` is listed in the runtime env table as image-baked `ENV`, but the Dockerfile doesn't set it** (only `NVSHMEM_REMOTE_TRANSPORT` / `NVSHMEM_LIBFABRIC_PROVIDER` / `NVSHMEM_NETDEVS_POLICY` are `ENV`). Either add `ENV FI_PROVIDER=efa` or correct the table.
- **`ARG NVCC_GENCODE` is declared but no longer consumed.** Nothing in the Dockerfile or the script's prebuilt-NVSHMEM path reads it (the script derives CMake archs from `TORCH_CUDA_ARCH_LIST` only on the source-build path the image doesn't use). It's a dead build ARG — remove it, or document it as advisory-only.

KeitaW (commented):
## Review Batch 3/3 — Strengths & Sources

### Things That Look Great

- **Supply-chain integrity on the NVSHMEM download.** `resolve_nvshmem_archive` reads NVIDIA's official `redistrib_<ver>.json` manifest and `verify_sha256` checks the archive against the manifest's own digest, with `download_with_retry` for transient failures. A real improvement over the previous hardcoded archive URL — corruption / CDN mismatch now fails loudly. (Manifest `redistrib_3.7.0.json` verified live, HTTP 200.)
- **The embedded patch is rot-resistant, and it fails *safe*.** `commit_compat_gate` pins to `567632d` with a `--force` escape hatch, and `apply_efa_patch` runs `git apply --check` before `git apply` — so an upstream drift or a corrupted heredoc fails the build with a clear message rather than mis-applying. `git apply` (exact) over `patch -p1` (fuzzy) is the right call. I verified the failure mode: against EPv2/HEAD the script aborts with `ERROR: EFA patch does not apply cleanly; the patch was validated against commit 567632d`, and a non-supported `--deepep-commit` is rejected up front — you never get a silently-broken image. (My only ask is to make the README's commit framing match this exactness — Batches 1 & 2.)
- **End-to-end verified on hardware.** I built the image from this PR (default args, CUDA 13) and ran all three benchmarks on 2× `p6-b300.48xlarge` (EFA): `test_intranode`, `test_internode` (EFA RDMA), and `test_low_latency` all pass. The build log confirms the new script's path — prebuilt NVSHMEM 3.7.0 installed and sha256-verified, EFA patch applied (4 files), CUDA-13 cccl fix applied, DeepEP installed, and the bundled `nvidia-nvshmem-cu13 3.4.5` correctly uninstalled.
- **The BuildKit requirement is now explicitly documented** — the README `[!NOTE]` calls out `RUN --mount` + `TARGETARCH`, gives the `DOCKER_BUILDKIT=1` command, and names `docker buildx build` as the alternative.
- **Solid shell hygiene throughout** — `set -euo pipefail`, shebang-before-MIT-0-header ordering, `die`/`log`/`warn` helpers to stderr, explicit `--option requires a value` validation, an idempotent `apply_cuda13_cccl_fix` (greps before `sed`), and a `SETUP_DEEPEP_EFA_LIB=1` source-guard so the functions are unit-testable without running `main`. Reads like production tooling.
- **Removing the checked-in `.patch` and the now-dead `.gitignore` negation** is clean — the patch lives with its applier and the `*.patch` ignore stays intact.

### Sources

- DeepEP commit `567632d` resolves (`gh api repos/deepseek-ai/DeepEP/commits/567632d`), verified live.
- NVSHMEM redist manifest `redistrib_3.7.0.json` → HTTP 200, verified live.
- Docker unconsumed-build-arg behavior: https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg
- Patch-applicability matrix — `git apply --check` of the embedded patch vs a fresh DeepEP clone: fails ≤ `92fe2de`, applies `29d31c0`→pre-EPv2, fails ≥ `b306af0` (EPv2, removed all 4 patched files). Verified live.
- e2e: image built from PR head; `test_intranode` / `test_internode` / `test_low_latency` all succeeded on 2× p6-b300.48xlarge (EKS). Verified live.

KeitaW (changes requested):
Thank you! Few comments

micro-benchmarks/expert-parallelism/deepep-benchmark/deepep.Dockerfile
erezzarum: Thanks, great catch.
Fixed and removed reference for `NVSHMEM_VERSION` and `DEEPEP_COMMIT`, the script automatically install the versions we support, so no need to expose those explicitly.

erezzarum:
> Thank you! Few comments

Thanks for the feedback, fixed.

KeitaW (approved):

Issue #1163, opened 2026-07-02
[Feature]: B300 (Blackwell, sm_103) support for the SLIME GRPO test case


littlemex:
### Update: implemented and open for review

Since filing this, I implemented every fix below, verified them end-to-end on both H200 and B300, and opened the PRs. This issue now tracks a completed, reviewable set rather than a proposal. **I would appreciate a review of the epic PR (#1164) and its focused sub-PRs.**

The SLIME GRPO test case (`3.test_cases/pytorch/slime`, added in #1129) now runs the full GRPO loop (rollout → train → weight sync → eval) for both Qwen3-4B (dense, colocated) and Qwen3-30B-A3B (MoE, disaggregated) on the CUDA-13 base image, after a small set of focused fixes. All fixes are landed on the epic branch behind #1164:

- **#1164** — `[Don't merge]` epic / integration PR that tracks the whole set.
- **#2** — `--sglang-log-level WARN` → lowercase `warning` (uvicorn startup).
- **#3** — cu13 `torch_memory_saver` `.so` selection (self-neutralizing build-time patch).
- **#4** — pin `numpy<2` (Megatron init vs numpy 2.x).
- **#5** — enable the 30B MoE disaggregated recipe (mbridge, GPU-less-driver validate guard, drop dead `--sglang-enable-ep-moe` + set `CUDA_DEVICE_MAX_CONNECTIONS=1`).

(PRs #2–#5 are on my fork against the epic branch; #1164 is the integration PR here.)

Two corrections vs my earlier description, from further verification: I dropped the large-HBM `--sglang-cuda-graph-max-bs` tuning (I could not reproduce a capture slowdown on H200, so I removed it as unverified rather than ship a preventive flag), and the "30B MoE online weight sync returns 400" item was wrong — no 400 occurs on SGLang's defaults; `--sglang-enable-ep-moe` turned out to be a dead flag that SGLang 0.5.12 silently ignores, so it is simply dropped. The real 30B blocker was elsewhere (Megatron `validate_args` probing the GPU on the GPU-less Ray driver), captured below.

### Feature Description & Motivation

[...]

PR #1164, opened 2026-07-03
Enable the SLIME GRPO test case to run end-to-end on B300 / H200

littlemex:
## Purpose

Relates to #1163.

This is the integration PR for a small epic: making the SLIME GRPO test case (`3.test_cases/pytorch/slime`, added in #1129) run through to training on NVIDIA B300 (p6) and H200 (p5en). Bringing the test case up surfaced a series of issues, each of which is a small, self-contained fix. Rather than one large "B300 support" change, the epic lands them as focused commits on this branch so each is easy to review, and this PR tracks the overall progress.

Thanks to the authors and reviewers of #1129 for the foundation this builds on: the container image, the reward-service split, the RayCluster manifests and the docs. The reward path works as designed, and every change here is scoped to getting training to start and run.

## How this PR is organized

Each fix is one focused commit. The mechanism, reproduction commands, and evidence for a fix live in a dedicated comment on this PR, not in this description. This description only tracks the overall goal, the scope, and the progress checklist.

## Scope and progress

The findings from #1163 are grouped by root cause. Most are general (they affect the documented p5/HyperPod path too — CUDA 13 and dependency issues, not GPU-generation-specific); a few are scoped to the 30B MoE recipe and are tracked separately below. Each item links to its detail comment as it lands.

General fixes (affect the documented p5/HyperPod path too; CUDA 13 / dependency issues, not GPU-generation-specific):

- [x] MODEL_ARGS reaches `train.py` as zero elements, so training never starts. Fixed by launching through a small launcher script instead of `ray job submit -- bash -c "..."`.
  - [x] detail comment: https://github.com/awslabs/awsome-distributed-ai/pull/1164#issuecomment-4872323165
  - [x] test checked on H200
- [x] `--sglang-log-level WARN` breaks uvicorn startup (uvicorn accepts lowercase log levels only), so the rollout HTTP server never binds and training hangs. Fixed by using lowercase `warning` (preserving the original intended verbosity).
  - [x] https://github.com/littlemex/awsome-distributed-training/pull/2
  - [x] test checked on H200
- [x] The train worker dies on CUDA 13 because SLIME `LD_PRELOAD`s a cu12-linked `torch_memory_saver` `.so`. This is a CUDA-13 issue, not GPU-generation-specific: it reproduces on any CUDA-13 image (the test case's NGC 26.02 base), on H100/H200/B300 alike. Proper fix filed upstream in SLIME (delegate `.so` selection to `torch_memory_saver`'s own CUDA-aware resolver); test-case side applies the same fix in place via a self-neutralizing build-time patch (no fork, no forked URL, no-op once upstream merges it).
  - [x] https://github.com/littlemex/awsome-distributed-training/pull/3
  - [x] test checked on H200 (built from `slime.Dockerfile`; resolves the `_cu13` `.so`, loads cleanly, train worker reaches training)
- [x] Megatron aborts on numpy 2.x pulled in transitively by `sglang[all]`. Fixed by pinning `numpy<2` in `requirements.txt` (the last dependency step that touches numpy), matching upstream SLIME's own Dockerfile. numpy is a pure-CPU package, so this is GPU-generation- and CUDA-independent (identical on H100/H200/B300).
  - [x] https://github.com/littlemex/awsome-distributed-training/pull/4
  - [x] test checked on H200 (built from `slime.Dockerfile`; numpy 1.26.4, Megatron init passes)
30B MoE recipe only (not exercised by the Qwen3-4B path; needed only for `run_grpo_qwen3_30b_a3b.sh`). All three are bundled into one experiment-scoped feature PR because the recipe does not reach training until all are in place:

- [x] 30B MoE `torch_dist` conversion needs `mbridge`, which the `--no-deps` slime install does not pull. Fix: `pip install --no-deps mbridge` in `slime.Dockerfile`.
- [x] Megatron `validate_args` eagerly probes the CUDA device (`get_device_capability` for `--moe-grouped-gemm`; `get_device_arch_version` for the TP/CP note), but SLIME runs `validate_args` on the GPU-less Ray driver, so it crashes with `Found no NVIDIA driver`. Only the MoE + TP/CP>1 recipe reaches these probes. Fix: a self-neutralizing build-time patch guards both probes in SLIME's `validate_args` wrapper only when `torch.cuda.is_available()` is False (real GPU actors unchanged); the permanent fix belongs upstream in Megatron.
- [x] `--sglang-enable-ep-moe` is a dead flag on SGLang 0.5.12 (the arg was removed; SLIME parses `--sglang-*` leniently, so it is silently ignored, not rejected) and the recipe omits `CUDA_DEVICE_MAX_CONNECTIONS=1`, which Megatron asserts for TP/CP>1. Fix: drop the dead flag (SGLang defaults serve the MoE correctly, verified end-to-end — no replacement flag needed) and add `CUDA_DEVICE_MAX_CONNECTIONS=1` to the recipe runtime-env.
  - [x] https://github.com/littlemex/awsome-distributed-training/pull/5
  - [x] test checked on H200 (30B MoE disaggregated reaches a full GRPO loop; see the end-to-end result below)

## Validation environment (common to the fixes here)

Everything above the cluster layer (image, recipe, model, versions, flags) is the test case as-is; only the cluster layer differs from the documented HyperPod/p5 target. Per-fix results are in each fix's comment.

- AWS Service: Amazon EKS
- Instance type: `p5en.48xlarge` (NVIDIA H200, 8 GPU/node)
- Number of nodes: 2

| Component | This validation | Test case default | Same? |
| --- | --- | --- | --- |
| Container image | built from this branch's `slime.Dockerfile` (with the fixes in this PR) | base image, versions and install path unchanged | fixes only |
| Ray | 2.55.1 | 2.55.1 (pinned) | yes |
| PyTorch / CUDA | 2.11.0+cu130 / 13.0 | NGC 26.02 (torch 2.11 / CUDA 13) | yes |
| SGLang | 0.5.12.post1 | 0.5.12.post1 (pinned) | yes |
| SLIME | v0.2.4 | v0.2.4 (pinned) | yes |
| Python | 3.12.3 | 3.12 (NGC) | yes |
| Model / data | Qwen3-4B, dapo-math-17k, aime-2024 (torch_dist) | same | yes |
| GPU generation | H200 (also observed on B300 per #1163) | H100 | cluster-layer diff |
| Orchestration | Amazon EKS | SageMaker HyperPod EKS | cluster-layer diff |

### End-to-end result so far

With the general fixes above, the Qwen3-4B colocated GRPO recipe (`recipe/run_grpo_qwen3_4b.sh`, built-in reward) now runs end-to-end. On 2x `p5en.48xlarge` (16x H200), an image built from this branch's `slime.Dockerfile` completed **3 full GRPO loops** (rollout generation → ref/actor log-probs → Megatron training step → weight sync back to SGLang), with no regressions (`libcudart.so.12`, numpy-2.x assert, uvicorn `KeyError`, `hidden_size None` all absent). Inter-node NCCL used EFA (`NET/OFI Selected provider is efa, fabric is efa-direct (found 16 nics)`, repeated across all 16 ranks). This exercises the documented quick-start path (4B dense, colocated, in-process reward).

**`remote_rm` HTTP reward service** (`reward_service/`, `kubernetes/reward-service.yaml`): also validated. I built the CPU-only reward image from `reward_service.Dockerfile`, ran the service (`math_verify` backend) on a non-GPU node, and re-ran the 4B recipe with `RM_TYPE=remote_rm` / `RM_URL=http://slime-reward.<ns>.svc.cluster.local:8000/score`. SLIME picked up `rm_type=remote_rm` and the rollout reached the reward service over HTTP: 250+ `POST /score` requests, all `200 OK`, zero non-200, and zero `remote_rm` retry/failure log lines on the SLIME side. The `/score` contract behaved as specified (correct answer -> 1.0, wrong -> 0.0). Environment delta: `reward-service.yaml` targets a HyperPod CPU instance group via `nodeSelector: sagemaker.amazonaws.com/instance-group-name`, which does not exist on plain EKS; I kept the manifest's GPU-exclusion `nodeAffinity` and dropped the instance-group `nodeSelector` so the pod lands on a non-GPU node, and added an `imagePullSecrets` for the cross-account ECR. These are cluster-layer adaptations; the reward service code and the recipe wiring are unchanged.

**30B MoE disaggregated recipe** (`recipe/run_grpo_qwen3_30b_a3b.sh`): also validated, on the same 2x `p5en.48xlarge` (16x H200), in a disaggregated topology (12 train GPUs at TP=2/EP=2/CP=2, 4 rollout GPUs). After the three 30B-only fixes above (mbridge, GPU-less driver validate guard, drop dead `--sglang-enable-ep-moe` + add `CUDA_DEVICE_MAX_CONNECTIONS=1`), the recipe reached a **full GRPO loop**: `Final collected 96 samples from rollout to train` → `Timer train start` → `Timer ref_log_probs end (27.4s)` → `Timer actor_train end (55.4s)` → `Timer train end (93.2s)` → `Timer update_weights end (12.1s)`, with zero occurrences of `Found no NVIDIA driver`, `not divisible by micro batch size`, `libcudart.so.12`, or `400 Bad Request` (the MoE online weight sync succeeded on SGLang's defaults). This is the disaggregated + MoE path, complementary to the 4B colocated run above.

## Open questions for the maintainers

1. Is the focused-commit approach on one `[Don't merge]` epic PR the shape you prefer, or would you rather each fix be its own separate PR?
2. For the 30B MoE recipe, do you prefer env/flag deltas in the existing recipe over a separate subdir, given the #1138 divergence discussion?
3. For the cu13 `torch_memory_saver` `.so`, is a thin test-case-side fallback acceptable until the CUDA-major selection lands upstream in SLIME?

## Checklist

- [x] New test cases follow the expected directory structure.

### DIFF
--- a/3.test_cases/pytorch/slime/README.md
+++ b/3.test_cases/pytorch/slime/README.md
@@ -532,8 +532,10 @@ slime/                                    # 3.test_cases/pytorch/slime
 │   ├── reward-service.yaml              # CPU reward service Deployment + Service
 │   └── data-prep-pod.yaml               # Utility pod for data preparation
 ├── recipe/
-│   ├── run_grpo_qwen3_4b.sh             # GRPO training launcher (Qwen3-4B, colocated)
-│   └── run_grpo_qwen3_30b_a3b.sh        # GRPO training launcher (Qwen3-30B-A3B MoE, disaggregated)
+│   ├── run_grpo_qwen3_4b.sh             # GRPO submit script (Qwen3-4B, colocated)
+│   ├── run_grpo_qwen3_30b_a3b.sh        # GRPO submit script (Qwen3-30B-A3B MoE, disaggregated)
+│   └── launcher/
+│       └── grpo_launch.sh               # Ray job entrypoint: sources the model script, expands MODEL_ARGS, execs train.py
 └── scripts/
     ├── convert_checkpoint.sh            # HF <-> Megatron conversion helper
     └── evaluate.sh                      # Evaluation launcher
--- a/3.test_cases/pytorch/slime/README.md
+++ b/3.test_cases/pytorch/slime/README.md
@@ -522,6 +522,8 @@ slime/                                    # 3.test_cases/pytorch/slime
 ├── env_vars.colocated.example           # Base config: colocated train+rollout, built-in reward
 ├── env_vars.disaggregated.example       # Overlay: reward model on a CPU pool + heavier GRPO
 ├── slime.Dockerfile                     # SLIME + SGLang + Megatron + EFA image
+├── patches/
+│   └── apply_slime_patches.py           # Self-neutralizing in-place fixes for the pinned upstream SLIME checkout (no-op once upstream merges them)
 ├── requirements.txt                     # Pinned Python RL dependencies
 ├── reward_service.Dockerfile            # CPU-only image for the remote reward service
 ├── reward_service/
--- a/3.test_cases/pytorch/slime/patches/apply_slime_patches.py
+++ b/3.test_cases/pytorch/slime/patches/apply_slime_patches.py
@@ -259,6 +259,48 @@ _VALIDATE_INJECT = '''def validate_args(args):
     _megatron_validate_args(args)'''
 
 
+def _megatron_probe_is_unguarded() -> bool:
+    """True if the installed Megatron's validate_args still probes the CUDA
+    device eagerly without a torch.cuda.is_available() guard.
+
+    This is the real defect this patch works around. When Megatron guards the
+    probe upstream (the permanent fix), this returns False and the SLIME-side
+    patch self-neutralizes. Locating Megatron via import keeps this robust to
+    the install path. If Megatron cannot be located or its arguments module has
+    been restructured beyond recognition, we conservatively assume the probe is
+    still unguarded (better a harmless is_available()-gated guard than a crash).
+    """
+    try:
+        import importlib.util
+
+        spec = importlib.util.find_spec("megatron.training.arguments")
+        if spec is None or not spec.origin:
+            return True
+        src = Path(spec.origin).read_text()
+    except Exception:
+        return True
+
+    if "get_device_capability()" not in src and "get_device_arch_version()" not in src:
+        # Probe removed/renamed entirely -> upstream changed it; nothing to guard.
+        return False
+
+    # A probe is considered guarded if a torch.cuda.is_available() check appears
+    # on the probe's own line OR within the few lines immediately preceding it
+    # (upstream's likely fix is `if ...is_available(): dc = get_device_capability()`
+    # -- the guard sits on the enclosing `if`, not the probe line itself). If any
+    # eager get_device_capability() call has no is_available() nearby, the probe
+    # is still unguarded and the SLIME-side workaround is needed.
+    lines = src.splitlines()
+    WINDOW = 3
+    for i, line in enumerate(lines):
+        if "get_device_capability()" not in line:
+            continue
+        context = lines[max(0, i - WINDOW): i + 1]
+        if not any("is_available" in c for c in context):
+            return True
+    return False
+
+
 def patch_gpuless_driver_validate(slime_root: Path) -> PatchResult:
     """Fix: let Megatron validate_args run on the GPU-less Ray driver.
 
@@ -277,14 +319,25 @@ def patch_gpuless_driver_validate(slime_root: Path) -> PatchResult:
     if "_slime_cpu_guard" in src:
         return PatchResult(name, "already-applied", f"{target} already guards the driver probes")
 
-    # If upstream (or SLIME) has guarded the probes with is_available(), the
-    # bare validate_args anchor may still be present but the fix is unnecessary;
-    # keep this conservative -- only skip when we truly cannot find the anchor.
-    if _VALIDATE_ANCHOR not in src:
+    # True self-neutralization: the real defect is Megatron's *unguarded* eager
+    # device probe in validate_args. If Megatron has guarded it upstream (the
+    # permanent fix), this patch is unnecessary and must not touch SLIME. Detect
+    # the unguarded probe in the installed Megatron; if it is gone, no-op.
+    if not _megatron_probe_is_unguarded():
         return PatchResult(
             name,
             "already-fixed-upstream",
-            "validate_args wrapper not in the expected shape; leaving it untouched",
+            "Megatron validate_args no longer has an unguarded CUDA device probe; leaving SLIME untouched",
+        )
+
+    # The probe is still unguarded upstream, so the SLIME-side guard is needed.
+    # It must be injectable into SLIME's validate_args wrapper in its known shape.
+    if _VALIDATE_ANCHOR not in src:
+        return PatchResult(
+            name,
+            "error",
+            "Megatron probe is unguarded but SLIME validate_args is not in the expected shape to inject a guard; "
+            "refusing to edit (needs a refreshed anchor)",
         )
     if src.count(_VALIDATE_ANCHOR) != 1:
         return PatchResult(
--- a/3.test_cases/pytorch/slime/patches/apply_slime_patches.py
+++ b/3.test_cases/pytorch/slime/patches/apply_slime_patches.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python3
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+"""Self-neutralizing patches for the pinned upstream SLIME checkout.
+
+This test case installs SLIME straight from upstream (``THUDM/slime`` at the
+pinned ``SLIME_VERSION`` -- no fork, no forked URL). A handful of upstream bugs
+still block a clean end-to-end run on B300 / H200 (CUDA 13). Rather than fork
+SLIME or hard-code a fork URL, this script applies each fix *in place* against
+the upstream checkout, and only when the unfixed pattern is actually present.
+
+Design goals (why this shape):
+  * Default is upstream. The image installs upstream SLIME as-is; this runs
+    afterwards as a thin, auditable layer.
+  * Self-neutralizing. Every patch first checks whether the upstream code still
+    exhibits the bug. Once upstream merges the corresponding fix, the pattern is
+    gone and the patch becomes a no-op automatically -- nothing to remember, no
+    version pin to bump, no fork to track. Upstream simply wins.
+  * Idempotent. Re-running (or running against an already-patched tree) is safe;
+    an applied patch is detected and skipped.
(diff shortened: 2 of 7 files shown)


littlemex:
### First fix in this epic: `MODEL_ARGS` reaches `train.py` as zero elements

Detail for the first commit on this branch (issue #1163, item 4). This one blocks the very first step of training, so it goes first.

Running either GRPO recipe exactly as documented (`bash recipe/run_grpo_qwen3_4b.sh`) fails immediately with `AssertionError: hf_validate_args failed: hidden_size in hf config 2560 is not equal to hidden_size None ...`.

A note on how this was scoped in #1163: it was originally listed there as Class C item 7 with a guess that it was plain-EKS-specific ("HyperPod's single `bash -c` doesn't hit this"). That guess was wrong. After reading the pinned Ray 2.55.1 source and reproducing the mechanism without SLIME, it is generation- and platform-independent: it happens on any KubeRay `ray job submit` path, including HyperPod. So it is really a general (Class A) bug.

**Root cause — a shell-expansion boundary, not a modeling issue.** The recipe builds the whole `train.py` invocation into one string and submits it as `ray job submit ... -- bash -c "${TRAIN_CMD}"`, with the model config deferred as `\${MODEL_ARGS[@]}`:

- https://github.com/awslabs/awsome-distributed-ai/blob/3e4edacd21eece8d6a0b70116c361c484c9ce4af/3.test_cases/pytorch/slime/recipe/run_grpo_qwen3_4b.sh#L69-L70
- https://github.com/awslabs/awsome-distributed-ai/blob/3e4edacd21eece8d6a0b70116c361c484c9ce4af/3.test_cases/pytorch/slime/recipe/run_grpo_qwen3_4b.sh#L152

`MODEL_ARGS` is a bash array defined by `source scripts/models/<model>.sh` (inside the image, from SLIME `v0.2.4`, pinned at `slime.Dockerfile` `ARG SLIME_VERSION=v0.2.4`). With Ray pinned to `ray[default]==2.55.1` (`3.test_cases/pytorch/slime/requirements.txt`), `ray job submit` re-joins everything after `--` with `subprocess.list2cmdline` and runs it through an outer shell via `Popen(shell=True)` → `/bin/sh -c` (Ray 2.55.1: `python/ray/dashboard/modules/job/cli.py`, `job_supervisor.py`). That outer shell expands `${MODEL_ARGS[@]}` before the inner `bash -c` sources the model script; `MODEL_ARGS` is undefined in the outer shell, so it expands to zero elements. `train.py` receives none of the model-shape flags and aborts in `hf_validate_args` (`hidden_size None`).

A bash array cannot survive being embedded in a string that a second shell re-parses, so the fix is to stop crossing that boundary rather than add another layer of escaping. Escaping cannot work here because `list2cmdline` always double-quotes the payload, so the outer shell always expands it.

**The fix — follow SLIME's own launch pattern.** SLIME's upstream launchers expand `${MODEL_ARGS[@]}` in the same shell that sourced it and pass everything as real argv tokens after `--`, never as one `bash -c` string:

- https://github.com/THUDM/slime/blob/v0.2.4/scripts/run-qwen3-4B.sh#L38 (source)
- https://github.com/THUDM/slime/blob/v0.2.4/scripts/run-qwen3-4B.sh#L146-L152 (`-- python3 train.py ${MODEL_ARGS[@]} ...`)

This commit adopts the same shape. A new `recipe/launcher/grpo_launch.sh` does `cd /opt/slime`, sources the model script, then `exec python3 train.py "${MODEL_ARGS[@]}" "$@"`, and it fails fast if `MODEL_ARGS` is empty. Both recipes assemble `train.py` flags as a bash array and submit with `ray job submit ... --working-dir <launcher-dir> -- bash grpo_launch.sh "${TRAIN_ARGS[@]}"`. The tokens after `--` are just `bash grpo_launch.sh <scalar flags>`, so no bash array crosses the `ray job submit` boundary and the whole class of outer-shell expansion bug is gone. `--working-dir` uploads the small launcher; SLIME itself is already in the image at `/opt/slime`, and `MODEL_SCRIPT` is forwarded via the runtime env so the launcher sources the right model definition.

Only the launch mechanism changes here. `train.py` flags are byte-for-byte identical to `main` (99 tokens for 4B, 103 for 30B, including `--expert-tensor-parallel-size`, `--rollout-num-gpus`, `--sglang-enable-ep-moe`, and `--sglang-log-level`). The `--sglang-log-level` value itself is a separate follow-up in the checklist.

**Reproduce it yourself.** Set `NAMESPACE` first; nothing else is environment-specific.

```bash
export NAMESPACE=<your-namespace>
HEAD=$(kubectl -n "$NAMESPACE" get pod -l ray.io/node-type=head -o jsonpath='{.items[0].metadata.name}')

# 1. The array IS defined in a single shell (control) -> prints count=25
kubectl -n "$NAMESPACE" exec "$HEAD" -- bash -lc \
  'source /opt/slime/scripts/models/qwen3-4B.sh && echo count=${#MODEL_ARGS[@]}'

# 2. The same command through "ray job submit -- bash -c" loses the array. Submit it,
#    then read the job log: count is 0.
JOB=$(kubectl -n "$NAMESPACE" exec "$HEAD" -- bash -lc \
  'ray job submit --address=http://127.0.0.1:8265 -- \
     bash -c "source /opt/slime/scripts/models/qwen3-4B.sh && echo count=\${#MODEL_ARGS[@]}"' \
  2>&1 | grep -oE 'raysubmit_[A-Za-z0-9]+' | head -1)
kubectl -n "$NAMESPACE" exec "$HEAD" -- \
  ray job logs "$JOB" --address http://127.0.0.1:8265 2>&1 | grep count=   # -> count=0

# 3. With the patched recipe, the launcher logs "MODEL_ARGS count=25" and Megatron
#    prints hidden_size 2560 (no "hidden_size None").

# Static check: no bash array in the submitted entrypoint
bash -n recipe/run_grpo_qwen3_4b.sh && bash -n recipe/launcher/grpo_launch.sh
```

**Result (H200).** The same reproduction, summarized:

| Launch path | how `${MODEL_ARGS[@]}` expands | args reaching the process |
| --- | --- | --- |
| single `bash -c "..."` | one shell | 25 (correct) |
| `ray job submit -- bash -c "..."` (current recipe) | outer `/bin/sh` expands first | 0 (`hidden_size None`) |
| `ray job submit -- bash grpo_launch.sh ...` (this fix) | inside the launcher | 25 (correct) |

End-to-end with the real `train.py`, Megatron receives the config and passes `hf_validate_args`, so training starts:

```
[grpo_launch] MODEL_SCRIPT=qwen3-4B.sh MODEL_ARGS count=25
[grpo_launch] launching: python3 train.py <25 model args> 100 recipe args
  ffn_hidden_size ................................. 9728
  hidden_size ..................................... 2560      # was: None
```

allela-roy (approved):
@littlemex , thanks for the PR. These are clean, well-engineered fixes and verified the end-end execution on my cluster. LGTM. Approved.

littlemex:
@allela-roy Thank you so much for taking the time to thoroughly verify the changes end-to-end on your cluster — I really appreciate it!

PR #1166, opened 2026-07-07
Add DeepSeek-V4-Flash intra-node 3P1D and GLM-5.2 4x tp=2 SGLang examples

ybalbert001:
Follow-up to #1128, adding two more SGLang inference examples for HyperPod EKS:

## dsv4flash-b300-intra-3p1d

Prefill/decode disaggregation on a single 8-GPU B300 node — 3 prefill + 1 decode engines (tp=2 each) in one pod, KV transfer over NIXL, router sidecar on 127.0.0.1.

## glm5.2-b300-tp2-dp4

Four independent tp=2 engines fill the node, fronted by an SGLang router (cache-aware load balancing, Kubernetes service discovery) — cluster-level dp=4 with per-replica scaling and rolling updates.

## Other changes

- Updated the examples table: added both samples and corrected the GLM-5.2 engine image (`dev-glm52-nvfp4`, not yet in a tagged release).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/inference/sglang/README.md
+++ b/examples/inference/sglang/README.md
@@ -14,9 +14,13 @@ manifest with `kubectl`.
 | [`qwen3.5-27b-b300-intra-pd`](./qwen3.5-27b-b300-intra-pd) | 1× B300 (8 GPU) | Intra-node PD — 6 prefill + 2 decode in one pod, NIXL, SGLang router sidecar | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
 | [`kimi2.6-h200-1p1d`](./kimi2.6-h200-1p1d) | 2× H200 nodes | Node-level 1P1D — prefill + decode StatefulSets, NIXL over EFA | Custom ECR build from [`Dockerfile.efa`](./Dockerfile.efa) (lmsysorg/sglang:v0.5.12.post1-cu130 + EFA layer), inter-node RDMA enabled |
 | [`dsv4pro-b300-single-node`](./dsv4pro-b300-single-node) | 1× B300 (8 GPU) | Unified (non-PD) baseline | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
+| [`dsv4flash-b300-intra-3p1d`](./dsv4flash-b300-intra-3p1d) | 1× B300 (8 GPU) | Intra-node PD — 3 prefill + 1 decode (tp=2 each) in one pod, NIXL, SGLang router sidecar | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
+| [`glm5.2-b300-tp2-dp4`](./glm5.2-b300-tp2-dp4) | 1× B300 (8 GPU) | 4× independent tp=2 engines behind an SGLang router (cache-aware LB, cluster-level dp=4) | `lmsysorg/sglang:dev-glm52-nvfp4` (GLM-5.2 NVFP4 support not yet in a tagged release) |
 
-All three serve on the same upstream `lmsysorg/sglang:v0.5.12.post1-cu130` image
-(Kimi adds only an EFA layer on top for inter-node RDMA) — no per-model image.
+All samples except GLM-5.2 serve on the same upstream
+`lmsysorg/sglang:v0.5.12.post1-cu130` image (Kimi adds only an EFA layer on top
+for inter-node RDMA); the GLM-5.2 sample uses the `dev-glm52-nvfp4` image until
+NVFP4 support for that model lands in a tagged release.
 
 ## Shared helpers
 
--- a/examples/inference/sglang/README.md
+++ b/examples/inference/sglang/README.md
@@ -15,7 +15,7 @@ SPDX-License-Identifier: MIT-0
 | [`dsv4flash-b300-intra-3p1d`](./dsv4flash-b300-intra-3p1d) | 1× B300 (8 GPU) | Intra-node PD — 3 prefill + 1 decode (tp=2 each) in one pod, NIXL, SGLang router sidecar | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
 | [`glm5.2-b300-tp2-dp4`](./glm5.2-b300-tp2-dp4) | 1× B300 (8 GPU) | 4× independent tp=2 engines behind an SGLang router (cache-aware LB, cluster-level dp=4) | `lmsysorg/sglang@sha256:bafcd0…` (the `dev-glm52-nvfp4` tag pinned by digest — GLM-5.2 NVFP4 support not yet in a tagged release) |
 
-> The intra-node PD samples deliberately run several engine processes in one pod — not the usual one-process-per-pod shape. Intra-node KV transfer rides NVLink via CUDA IPC, which requires all engines to share an IPC namespace and see each other's GPUs; 
+> The intra-node PD samples deliberately run several engine processes in one pod — not the usual one-process-per-pod shape. Intra-node KV transfer rides NVLink via CUDA IPC, which requires all engines to share an IPC namespace and see each other's GPUs — impossible across separate pods or containers, so the engines share one container (each sample's README explains the full rationale).
 
 > All samples except GLM-5.2 serve on the same upstream `lmsysorg/sglang:v0.5.12.post1-cu130` image (Kimi adds only an EFA layer on top
 for inter-node RDMA); the GLM-5.2 sample uses the `dev-glm52-nvfp4` image (pinned by digest in its manifest, since `dev-*` tags are mutable) until NVFP4 support for that model lands in a tagged release.
--- a/examples/inference/sglang/README.md
+++ b/examples/inference/sglang/README.md
@@ -5,9 +5,7 @@ SPDX-License-Identifier: MIT-0
 
 # SGLang test cases
 
-[SGLang](https://github.com/sgl-project/sglang) deployments on AWS EKS /
-SageMaker HyperPod. Each sub-directory is a self-contained sample — apply its
-manifest with `kubectl`.
+[SGLang](https://github.com/sgl-project/sglang) deployments on AWS EKS / SageMaker HyperPod. Each sub-directory is a self-contained sample — apply its manifest with `kubectl`.
 
 | Test case | Hardware | Topology | Engine image |
 | --- | --- | --- | --- |
@@ -17,10 +15,10 @@ manifest with `kubectl`.
 | [`dsv4flash-b300-intra-3p1d`](./dsv4flash-b300-intra-3p1d) | 1× B300 (8 GPU) | Intra-node PD — 3 prefill + 1 decode (tp=2 each) in one pod, NIXL, SGLang router sidecar | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
 | [`glm5.2-b300-tp2-dp4`](./glm5.2-b300-tp2-dp4) | 1× B300 (8 GPU) | 4× independent tp=2 engines behind an SGLang router (cache-aware LB, cluster-level dp=4) | `lmsysorg/sglang:dev-glm52-nvfp4` (GLM-5.2 NVFP4 support not yet in a tagged release) |
 
-All samples except GLM-5.2 serve on the same upstream
-`lmsysorg/sglang:v0.5.12.post1-cu130` image (Kimi adds only an EFA layer on top
-for inter-node RDMA); the GLM-5.2 sample uses the `dev-glm52-nvfp4` image until
-NVFP4 support for that model lands in a tagged release.
+> The intra-node PD samples deliberately run several engine processes in one pod — not the usual one-process-per-pod shape. Intra-node KV transfer rides NVLink via CUDA IPC, which requires all engines to share an IPC namespace and see each other's GPUs; 
+
+> All samples except GLM-5.2 serve on the same upstream `lmsysorg/sglang:v0.5.12.post1-cu130` image (Kimi adds only an EFA layer on top
+for inter-node RDMA); the GLM-5.2 sample uses the `dev-glm52-nvfp4` image until NVFP4 support for that model lands in a tagged release.
 
 ## Shared helpers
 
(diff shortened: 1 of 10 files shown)


ybalbert001:
1.  Launching multiple processes in a single pod is not Kubernetes native

The intra-node PD samples deliberately run several engine processes in one pod — not the usual one-process-per-pod shape. Intra-node KV transfer rides NVLink via CUDA IPC, which requires all engines to share an IPC namespace and see each other's GPUs;

2. port specification could conflict from what manifest promises.

Fixed by : https://github.com/awslabs/awsome-distributed-ai/pull/1166/changes/f27237ecf73f41773d5c6e61c113a2f64523e736#diff-c46bb3bf0724f7f1dcf3b8255d2f75b8343e96e51a18229bd68e22d7acb7d3deR112-R118

examples/inference/sglang/dsv4flash-b300-intra-3p1d/dsv4flash-pd-deploy.yaml
```diff
@@ -0,0 +1,286 @@
...
+            nvidia.com/gpu: 8
+        # Probe the foreground prefill (30200) — it keeps the container alive, so
+        # if it wedges the pod should restart. The 1800s initial delay clears the
+        # long cold start (4 engines load + DeepGEMM JIT + CUDA-graph capture,
+        # gated behind the decode-ready wait loop) so liveness never fires during
+        # warmup.
+        livenessProbe:
+          httpGet:
+            path: /health
+            port: 30200
```
KeitaW: **Health gating: a dead background engine is silently unrecovered — I don't think "can't serve anyway" holds.**

Both probes target only `30200` (the foreground prefill). The README defends this ("probes watch the foreground engine and the whole group restarts together … a half-alive PD group can't serve anyway"), so I tested it live on a B300 node — killed the background prefill on `30000` inside the container:

- the engine went dead (`/health` → connection refused), **but** the pod stayed `Ready 2/2` with `RESTARTS 0`, and the router's own health-checks routed around it → **12/12 completions still succeeded**.

So two parts of the rationale don't hold empirically:
1. *"can't serve anyway"* — it did keep serving, at silently reduced prefill capacity (3→2), with no recovery (nothing restarts the dead engine until a manual `kubectl delete pod`). "Restarts together" only kicks in when the **foreground** engine (30200) dies.
2. The **decode** (30300) is a single point with no redundancy — if it dies, the router has nothing to route to → the pod serves zero tokens while still reporting `Ready`, and never restarts.

Could we either gate the pod on all four engines (a small wrapper `/health` that ANDs 30000/30100/30200/30300, or health-check the backgrounded prefills before the foreground `exec`, mirroring the decode wait-loop), or — if all-or-nothing restart is the intent — point liveness at the **decode** (the real SPOF) and reword the README from "can't serve anyway" to "a dead engine isn't auto-recovered; restart the pod"?

examples/inference/sglang/glm5.2-b300-tp2-dp4/manifests/glm52-deploy.yaml
```diff
@@ -0,0 +1,300 @@
...
+      # keep non-GPU pods off; HyperPod nodes have no such taint. A toleration is
+      # a no-op when the taint is absent, so it's safe to always carry it.
+      tolerations:
+      - key: nvidia.com/gpu
+        operator: Equal
+        value: "true"
+        effect: NoSchedule
+      containers:
+      - name: sglang
+        image: lmsysorg/sglang:dev-glm52-nvfp4
```
KeitaW: **`dev-glm52-nvfp4` is a moving tag — could we pin it to a digest?**

It's a mutable `dev-*` tag (Docker Hub shows it overwritten 2026-07-03), so two `kubectl apply`s a week apart can pull different binaries — which trips the repo's pin-everything rule (CONTRIBUTING + CI floors). Pinning the digest keeps the example reproducible even before NVFP4 lands in a tagged release:

```suggestion
        image: lmsysorg/sglang@sha256:bafcd07dbc511ca4cbae8e3865fecca98a0323b777c83c4d3083008228891f65  # dev-glm52-nvfp4, until GLM-5.2 NVFP4 lands in a tagged release
```

Worth re-resolving with `docker manifest inspect lmsysorg/sglang:dev-glm52-nvfp4` right before merge in case it moves again, and mirroring the pinned form into the two README references.

examples/inference/sglang/glm5.2-b300-tp2-dp4/README.md
```diff
@@ -0,0 +1,63 @@
...
+  node needs no config change. One monolithic engine can only scale by redeploying.
+- **Failure isolation & rolling updates** — one wedged engine restarts alone while the other 3 keep serving; a rolling update keeps 3/4 capacity online
+  (`maxSurge: 0, maxUnavailable: 1`, since all 8 GPUs are taken).
+- **Engine constraint** — with `--enable-dp-attention` SGLang requires `dp` to divide `tp`, so `tp=2, dp=4` is not a valid single-engine config anyway.
+
+## Prerequisites
+
+1. **Provision the B300 node** — on self-managed (eksctl) EKS, create the nodegroup first: see [`../dsv4pro-b300-single-node/NODEGROUP-EKS.md`](../dsv4pro-b300-single-node/NODEGROUP-EKS.md).
+   On HyperPod-on-EKS, nodes are already provisioned.
+2. **Pre-stage the weights** (recommended) — all four replicas share the node's HF cache, so download once instead of having 4 engines race:
```
KeitaW: **"Share the node's HF cache / download once" doesn't hold as written — verified live.**

Two gaps, both seen on a live 4-replica deploy:

1. **No co-location constraint** — on a plain `kubectl apply` the 4 replicas spread across 4 different nodes (2 GPUs each; nothing pins them together). So on any multi-node cluster there's no shared node cache — each node re-downloads all 465 GB (4× the traffic). I had to pin them to one node to get the single-download behavior described here.
2. **Staged path ≠ mounted cache** — `download-model.sh` stages to `/opt/dlami/nvme/<repo-id>` (a plain `snapshot_download` dir), but the pods mount `/opt/dlami/nvme/huggingface → /root/.cache/huggingface` and load by repo-id, so the pre-staged dir is never read. (Shared-helper behavior, not unique to this sample — but this README is the one that leans on it.)

(One thing that *did* hold: when the 4 replicas share one cache, the concurrent download converged fine via HF's file-locks — no corruption. So the "race" is benign; the real issues are the 4× traffic without co-location and the no-op pre-stage.)

Could we add a `podAffinity`/topology constraint so the replicas co-locate, and either mount the staged dir (point `--model-path` at it) or stage into the HF-cache layout — or otherwise soften the claim?

examples/inference/sglang/glm5.2-b300-tp2-dp4/manifests/glm52-deploy.yaml
```diff
@@ -0,0 +1,300 @@
...
+      # The HF cache is shared by all four replicas on the node — the weights
+      # are downloaded (or pre-staged) once and read by every engine.
+      - name: nvme-hf
+        hostPath:
+          path: /opt/dlami/nvme/huggingface  #For EKS: /mnt/k8s-disks/0/huggingface
+          type: DirectoryOrCreate
+      - name: nvme-tmp
+        hostPath:
+          path: /opt/dlami/nvme/tmp         #For EKS: /mnt/k8s-disks/0/tmp
+          type: DirectoryOrCreate
```
KeitaW: **Missing trailing newline (nit).** Repo convention is a newline at EOF; re-stating the last line adds it:

```suggestion
          type: DirectoryOrCreate
```

examples/inference/sglang/dsv4flash-b300-intra-3p1d/README.md
```diff
@@ -0,0 +1,55 @@
...
+The router exposes an OpenAI-compatible endpoint on `dsv4flash-router:30080` (`ClusterIP`) — port-forward to call it:
+
+```bash
+kubectl port-forward svc/dsv4flash-router 30080:30080
+curl http://localhost:30080/v1/completions \
+  -H 'Content-Type: application/json' \
+  -d '{"model": "deepseek-ai/DeepSeek-V4-Flash", "prompt": "The capital of France is", "max_tokens": 32}'
+```
+
+Tear down with `kubectl delete -f dsv4flash-pd-deploy.yaml`.
```
KeitaW: **Missing trailing newline (nit)** — same here:

```suggestion
Tear down with `kubectl delete -f dsv4flash-pd-deploy.yaml`.
```

KeitaW (changes requested):
Thanks for the fast turnaround — the port-contract fix and especially the new **"Why one pod with four engine processes?"** section both landed before this review posted. The CUDA-IPC rationale (shared IPC namespace + mutual GPU visibility, the NVIDIA disagg citation, and the PD decision tree) is exactly the right way to document why the single-pod shape is correct here.

I also deployed **both** samples end-to-end on a B300 node (self-managed EKS, 8× p6-b300.48xlarge): both serve real completions through their routers — glm5.2's Kubernetes service-discovery + RBAC and dsv4flash's intra-node decode→prefill PD both work.

A few things remain, detailed inline:

1. 🟡 **Health gating** (`dsv4flash-pd-deploy.yaml`) — a live kill-test shows a dead background engine leaves the pod `Ready` with no restart, and the single decode is a non-redundant SPOF; the README's "a half-alive PD group can't serve anyway" doesn't hold empirically (it kept serving at reduced capacity).
2. 🟡 **`dev-glm52-nvfp4` moving tag** — pin to a digest for reproducibility.
3. 🟡 **glm5.2 "share the node's HF cache"** — no co-location constraint, so replicas spread across nodes and re-download 465 GB each; and the pre-stage path isn't the mounted cache.
4. 🟢 Trailing newline on two files.

Nothing here is a redesign — (1) is the substantive one. Requesting changes mainly to track (1)–(3); happy to discuss the health-gating trade-off if you see it differently, since you've clearly thought about the failure semantics. Great work overall.

KeitaW (changes requested):
Few comments

ybalbert001:
Thanks for the careful review — deploying both samples end-to-end and kill-testing the health gating goes well beyond the usual pass, and all four findings were real. 

1. Health gating — liveness now ANDs all four engines (exec probe); 
2. Moving tag — pinned to the digest ; 
3. HF cache — root-caused to the pre-stage layout mismatch; the shared daemonset now stages in HF cache layout so pre-staging actually works. I will update kimi2.6 and qwen3.5 to make them consistent in the later PR.
4. Newlines — fixed.

examples/inference/sglang/README.md
```diff
@@ -5,18 +5,20 @@ SPDX-License-Identifier: MIT-0
...
 | [`dsv4pro-b300-single-node`](./dsv4pro-b300-single-node) | 1× B300 (8 GPU) | Unified (non-PD) baseline | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
+| [`dsv4flash-b300-intra-3p1d`](./dsv4flash-b300-intra-3p1d) | 1× B300 (8 GPU) | Intra-node PD — 3 prefill + 1 decode (tp=2 each) in one pod, NIXL, SGLang router sidecar | `lmsysorg/sglang:v0.5.12.post1-cu130`, no inter-node RDMA support |
+| [`glm5.2-b300-tp2-dp4`](./glm5.2-b300-tp2-dp4) | 1× B300 (8 GPU) | 4× independent tp=2 engines behind an SGLang router (cache-aware LB, cluster-level dp=4) | `lmsysorg/sglang@sha256:bafcd0…` (the `dev-glm52-nvfp4` tag pinned by digest — GLM-5.2 NVFP4 support not yet in a tagged release) |
 
-All three serve on the same upstream `lmsysorg/sglang:v0.5.12.post1-cu130` image
-(Kimi adds only an EFA layer on top for inter-node RDMA) — no per-model image.
+> The intra-node PD samples deliberately run several engine processes in one pod — not the usual one-process-per-pod shape. Intra-node KV transfer rides NVLink via CUDA IPC, which requires all engines to share an IPC namespace and see each other's GPUs;
```
KeitaW: ### The intra-node PD summary sentence is truncated

The new overview blockquote sets up the rationale and then stops mid-thought: it ends `…share an IPC namespace and see each other's GPUs; ` — a dangling semicolon with no conclusion (and a trailing space). The full explanation lives in each sample's README ("Why one pod…"), so this table-level summary just needs its clause closed:

```suggestion
> The intra-node PD samples deliberately run several engine processes in one pod — not the usual one-process-per-pod shape. Intra-node KV transfer rides NVLink via CUDA IPC, which requires all engines to share an IPC namespace and see each other's GPUs — impossible across separate pods or containers, so the engines share one container (each sample's README explains the full rationale).
```

examples/inference/sglang/dsv4flash-b300-intra-3p1d/README.md
```diff
@@ -0,0 +1,65 @@
...
+- **PD within one node, no RDMA between engines** — the engines must live in one container that owns all 8 GPUs, as here. The pod is still a single
+  schedulable, restartable unit: the liveness probe ANDs the `/health` of **all four engines**, so if any engine dies the whole group restarts together
+  (`strategy: Recreate`). The router does health-check around a dead prefill, but that leaves the pod serving at reduced capacity with no path back to
+  full — and a dead decode leaves nothing to route to at all — so any dead engine triggers a group restart (at the cost of the ~30 min cold start).
+  Readiness stays on the foreground engine only, so a dying background engine doesn't cut traffic before the restart lands.
+- **No PD at all** — if you just want N independent engines on a node, one-engine-per-pod is the native shape; see [`glm5.2-b300-tp2-dp4`](../glm5.2-b300-tp2-dp4).
+
+## Deploy
+
+**Pre-stage the weights first (recommended)** — this pod runs **four** engine processes that all load the same model from the node's HF cache; without pre-staging, the cold start begins with the decode engine downloading the full weights before anything else can proceed:
```
KeitaW: ### Pre-staging is effectively required, not just recommended

The Deploy section labels pre-staging **(recommended)**, but the manifest's decode wait-loop hard-exits after `240 × 5s = 20 min` (`dsv4flash-pd-deploy.yaml` L482–491), and on a cold cache the decode engine must pull the full ~160 GB before that loop can pass. On a throttled HF link a first `kubectl apply` without pre-staging can exceed the 20-min bound and the pod will restart.

It does self-heal — `nvme-hf` is a node-local `hostPath` so the partial download survives, `strategy: Recreate` + `replicas: 1` relands on the same node, and `snapshot_download` resumes — so it's "restart a few times while the node cache fills," not a permanent CrashLoopBackOff. Still, calling pre-staging **required** for this sample makes the happy path the documented one:

```suggestion
**Pre-stage the weights first (required for this sample)** — this pod runs **four** engine processes that all load the same model from the node's HF cache; the decode engine's startup wait-loop gives up after 20 min, so on a cold cache the full weights must download first or the pod restarts until the node cache fills:
```

KeitaW (approved):
Round 2 — all five round-1 findings resolved; approving. Thanks for the thorough turnaround.

I re-checked each against the current head (`64a1c45`):

1. **Health gating** — liveness now `exec`-ANDs `/health` across all four engine ports (`30000/30100/30200/30300`), so a dead background prefill *or* the single decode now triggers a group restart instead of leaving the pod silently `Ready`. Readiness deliberately stays on the foreground engine, with comments explaining why (avoids flipping the sole replica `NotReady` during the restart window). ✅
2. **`dev-glm52-nvfp4` moving tag** — pinned by digest (`@sha256:bafcd07…`); README and examples table mirror the pinned form. ✅
3. **glm5.2 "share the node's HF cache"** — root-caused: the daemonset now stages into HF cache layout (`cache_dir=.../huggingface/hub`). I traced the path and it lands at `/opt/dlami/nvme/huggingface/hub/models--*`, exactly what the serving pods read via their `/opt/dlami/nvme/huggingface → /root/.cache/huggingface` mount — so the pre-stage is a genuine cache hit now, and the over-claim is gone. ✅
4. **Trailing newlines** (`glm52-deploy.yaml`, `dsv4flash/README.md`) — both fixed. ✅

**Things that look great**
- **The AND-gate liveness reasoning is genuinely good.** `exec` for-loop because `httpGet` can only hit one port; readiness on the foreground engine to avoid a self-inflicted outage during the restart window; the `1800s` initial delay sized against the real cold start. The comments make the trade-off (self-heal to full capacity at the cost of a ~30-min restart) explicit rather than implicit — probe design in examples is usually wrong, and this isn't.
- **Pre-stage root-cause, not band-aid.** Rewriting the daemonset to HF cache layout (and dropping the now-meaningless `LOCAL_DIR_NAME`) fixes it for *every* sample.
- **Consistency pass on qwen3.5** — named `prefill-*`/`decode-*` ports, phantom `30080` removed from the engine container, and the shared "Why one pod" rationale propagated; the note committing to finish kimi2.6/qwen3.5 is good stewardship.
- **`subPathExpr: $(POD_NAME)`** on glm5.2's `/tmp` isolates per-replica compile caches while intentionally sharing `nvme-hf` — the right split.
- **Least-privilege RBAC** on the glm5.2 router (namespaced `Role` + `POD_NAMESPACE` fieldRef feeding `--service-discovery-namespace`).

I left two **optional nits** inline (a truncated summary sentence, and promoting pre-stage from "recommended" to "required" for dsv4flash) — neither blocks. Nice work.

PR #1167, opened 2026-07-07
chore(aws-pcs): clean up deploy-all parameter presentation

DaisukeMiyamoto:
## What

Console-presentation cleanup of `pcs-ml-cluster-deploy-all.yaml`'s parameters.
No behavior change; nested templates untouched.

- **Split parameter group 5 by purpose** — the 9-parameter "Additional Cluster
  Configuration" group is now 5.1 Monitoring / 5.2 Multi-User Directory /
  5.3 Post-Install Script (default: Enroot/Pyxis container runtime).
- **`Advanced:` label prefix** on the pinned monitoring params
  (`MonitoringRepo` / `MonitoringVersion` / `DcgmExporterImage`).
- **Doc-accuracy fix:** scope the placement-group reuse docs to the CPU CNG
  (multiple CPU CNGs, or a customer-owned CPG holding reserved capacity) —
  the P5/P6 GPU templates expose no placement-group option. Applied to
  README §8.6, PARAMETERS.md, and the template description.

Docs (PARAMETERS.md, README, tests) updated to match; `tests/lint-docs.sh` passes.

## Testing

`validate-template` + params ⇄ groups ⇄ labels consistency check pass. e2e on
real hardware (us-east-2, `DirectoryService=OpenLDAP-LoginNode`):
CREATE_COMPLETE, Pyxis container job OK, LDAP user runs `srun` on a compute
node as uid=10001.

### DIFF
--- a/architectures/aws-pcs/README.md
+++ b/architectures/aws-pcs/README.md
@@ -132,18 +132,28 @@ complete reference see [PARAMETERS.md](./docs/PARAMETERS.md).
 | `PseriesInstanceType` | `p5.48xlarge` | Picks the matching template + EFA NIC count automatically. See [GPU compute](#gpu-compute-p5p6) for the accepted types |
 | `CapacityReservationId` | *(empty)* | Capacity **Block** ID for the GPU queue; empty for On-Demand/ODCR |
 
-**5. Additional Cluster Configuration (Monitoring, Multi-User, Container Runtime)**
+**5.1. Additional Cluster Configuration: Monitoring**
 
 | Parameter | Default | Purpose |
 |---|---|---|
 | `MonitoringStack` | `Prometheus-LoginNode` | Prometheus + Grafana on the login node, DCGM Exporter on GPU nodes. `none` disables it. See [§8.2](#82-monitoring) |
 | `GrafanaAccessCidr` | *(empty)* | Open HTTPS/443 (Grafana) on the login node to a trusted CIDR (default: SSM port-forward only) |
+
+(Group 5.1 also has `MonitoringRepo` / `MonitoringVersion` / `DcgmExporterImage` — pinned defaults, rarely changed; see [PARAMETERS.md](./docs/PARAMETERS.md).)
+
+**5.2. Additional Cluster Configuration: Multi-User Directory**
+
+| Parameter | Default | Purpose |
+|---|---|---|
 | `DirectoryService` | `none` | `OpenLDAP-LoginNode` for a multi-user cluster. See [§8.3](#83-user-management) |
-| `PostInstallScriptUrl` | *(empty → auto)* | First-boot script; empty auto-installs Enroot/Pyxis from your templates bucket. Rarely changed. See [PARAMETERS.md](./docs/PARAMETERS.md) |
 
-(Group 5 also has `MonitoringRepo` / `MonitoringVersion` / `DcgmExporterImage` — pinned defaults, rarely changed; see [PARAMETERS.md](./docs/PARAMETERS.md).)
+**5.3. Additional Cluster Configuration: Post-Install Script**
 
-**See [PARAMETERS.md](./docs/PARAMETERS.md) for the complete parameter reference** (all 7
+| Parameter | Default | Purpose |
+|---|---|---|
+| `PostInstallScriptUrl` | *(empty → auto)* | First-boot script on every node; empty (default) auto-installs the Enroot/Pyxis container runtime from your templates bucket. Point at your own script to customize. See [PARAMETERS.md](./docs/PARAMETERS.md) |
+
+**See [PARAMETERS.md](./docs/PARAMETERS.md) for the complete parameter reference** (all
 console parameter groups, with every default). The concept guides below cover the
 choices that need the most thought.
 
--- a/architectures/aws-pcs/README.md
+++ b/architectures/aws-pcs/README.md
@@ -623,8 +623,12 @@ multi-NIC EFA wiring per-family.
 instance type's actual `MaximumEfaInterfaces`, fails at launch.)
 
 **Placement group:** auto-created per-CNG by the template. Override with
-`OnDemandPlacementGroupName=<existing-pg-name>` to share one PG across multiple
-CNGs (e.g. heterogeneous tightly-coupled jobs that span CPU + GPU).
+`OnDemandPlacementGroupName=<existing-pg-name>` to launch the CPU CNG into an
+existing placement group instead — e.g. to share one PG across multiple CPU
+CNGs, or to consume capacity reserved into a customer-owned CPG. (The P5/P6
+GPU templates don't expose an existing-placement-group option: a Capacity
+Block carries its own placement and works as-is; targeting a customer-owned
+CPG from the GPU templates is a future item.)
 
 **Multi-NIC bandwidth needs multiple MPI pairs.** A single MPI pair uses one
 libfabric endpoint and only one NIC. Use `osu_mbw_mr -np 32 -N 16` (or your
--- a/architectures/aws-pcs/README.md
+++ b/architectures/aws-pcs/README.md
@@ -622,11 +622,9 @@ multi-NIC EFA wiring per-family.
 (Setting a count > 0 on a non-EFA type, or mismatching the count with the
 instance type's actual `MaximumEfaInterfaces`, fails at launch.)
 
-**Placement group:** auto-created per-CNG by the template. To reuse an existing
-placement group across multiple CNGs (e.g. heterogeneous tightly-coupled jobs
-that span CPU + GPU), deploy the CNG with the modular
-[`add-cng.yaml`](./assets/add-cng.yaml) and set its `PlacementGroupName`
-parameter — deploy-all always auto-creates.
+**Placement group:** auto-created per-CNG by the template. Override with
+`OnDemandPlacementGroupName=<existing-pg-name>` to share one PG across multiple
+CNGs (e.g. heterogeneous tightly-coupled jobs that span CPU + GPU).
 
 **Multi-NIC bandwidth needs multiple MPI pairs.** A single MPI pair uses one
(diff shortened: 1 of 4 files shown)


KeitaW (commented):
## Review Batch 1/3 — Capacity-Reservation & Placement-Group Strategy

Nice, tight cleanup — 4 files, ~51/33, well within the repo's ≤200-line target, and the three doc surfaces (README, PARAMETERS.md, tests) move together so nothing drifts. The CloudFormation deletion is complete and clean. Two non-blocking items: I'd suggest keeping `OnDemandPlacementGroupName`, plus one doc-accuracy fix on the placement-group note (Batch 2). Neither blocks merge.

### I'd suggest keeping `OnDemandPlacementGroupName` — it's the top-level existing-CPG-targeting affordance for the HPC-EFA path, and removal is a real tradeoff, not pure presentation

**Observation.** `OnDemandPlacementGroupName` did one thing: launch the on-demand nodes into an **existing** cluster placement group instead of an auto-created one. The nodes it served are the EFA-enabled HPC types (`hpc6a`/`hpc7a`/`hpc8a`) — tightly-coupled MPI, i.e. exactly the spine-locality workload CPGs exist for. "Share one PG across CNGs" was the *documented* example, but the underlying capability is the natural **consume-capacity-that-was-reserved-into-a-customer-owned-CPG** path: self-service CR-in-CPG, a RAM-shared CPG, or an assisted CPG-into-CPG (CPG-ACR) delivery.

**Impact — the case for keeping it.** It's an optional parameter, backward-compatible (`Default: ''` → no behavior change for any existing deployment), and costs ~one line of console space. Against that, its removal takes the single top-level affordance for pointing an EFA-HPC queue at a reserved/existing CPG out of `deploy-all`. The "rare, advanced case" framing is debatable here — for tightly-coupled HPC that's the point of a CPG, not an edge case.

**Impact — the case for removing it (why it's non-blocking).** The escape hatch is clean: `deploy-all` gates the CPU CNG behind `DeployOnDemandCNG` (`Condition: DeployOnDemandCNGEnabled`), so a customer who needs an existing/reserved CPG can set `DeployOnDemandCNG=false` and deploy that CNG standalone via the modular `add-cng.yaml`, which still declares `PlacementGroupName`. No capability is lost repo-wide; it moves one deploy surface over.

**Suggestion — keep the parameter.** I'd suggest leaving it in place. One optional, default-empty knob is cheap insurance for a case the modular fallback confirms is real (reserved/existing CPG for tightly-coupled HPC), and its cost is a single console line. The `DeployOnDemandCNG=false` + standalone-`add-cng.yaml` path works, but it's a heavier ask than flipping one field — the customer drops the CPU CNG out of the all-in-one stack and manages a second stack just to set one placement-group name. If you'd still rather remove it for presentation reasons, that's defensible given the escape hatch, but then I'd ask for two things: (a) don't describe this as "no behavior change" for the advanced/HPC user (it removes their one top-level knob), and (b) make the reserved-CPG workaround explicit in the docs (see Batch 2). This is a suggestion, not a blocker — the capability survives via the modular template either way.

**Related, and the bigger gap (a separate follow-up, not this PR).** The GPU path has *never* been able to target an existing CPG at all: `add-cng-p5.yaml` / `add-cng-p6-*.yaml` declare no `PlacementGroupName`, auto-create a fresh placement group only on the on-demand branch (`Condition: UseOnDemand`), and when a `CapacityReservationId` is set they take `InstanceMarketOptions: MarketType: capacity-block` — the Capacity Block path, which conflates "has a reservation" with "is a Capacity Block" (a plain On-Demand CR / CPG-delivered reservation in that field would fail, since `capacity-block` market type is Capacity-Block-only). So capacity reserved into your own CPG via CPG-ACR — which you consume by launching into that CPG — isn't consumable through these templates today. Worth its own issue: add a GPU existing-CPG option and distinguish an ODCR from a Capacity Block.

*Refs: `pcs-ml-cluster-deploy-all.yaml` `DeployOnDemandCNG` gate; `add-cng.yaml` `PlacementGroupName` default; `add-cng-p5.yaml` placement/capacity-reservation logic; [Use Capacity Reservations with cluster placement groups](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/cr-cpg.html).*

architectures/aws-pcs/README.md
```diff
@@ -612,9 +622,11 @@ multi-NIC EFA wiring per-family.
...
 instance type's actual `MaximumEfaInterfaces`, fails at launch.)
 
-**Placement group:** auto-created per-CNG by the template. Override with
-`OnDemandPlacementGroupName=<existing-pg-name>` to share one PG across multiple
-CNGs (e.g. heterogeneous tightly-coupled jobs that span CPU + GPU).
+**Placement group:** auto-created per-CNG by the template. To reuse an existing
+placement group across multiple CNGs (e.g. heterogeneous tightly-coupled jobs
+that span CPU + GPU), deploy the CNG with the modular
+[`add-cng.yaml`](./assets/add-cng.yaml) and set its `PlacementGroupName`
+parameter — deploy-all always auto-creates.
```
KeitaW: ### The README's "share one PG across CPU + GPU" example isn't achievable — the reuse path is CPU-only

**Observation.** The reworded placement-group note keeps the pre-existing example that an existing PG can span "CPU + GPU": it points the reader to "the modular `add-cng.yaml` and set its `PlacementGroupName`." But `add-cng.yaml` provisions **CPU** CNGs only; the GPU node groups use `add-cng-p5.yaml` / `add-cng-p6-*.yaml`, which expose no `PlacementGroupName`. So the documented path shares a PG across CPU CNGs — never across CPU **and** GPU.

**Impact.** A reader chasing a heterogeneous CPU+GPU job follows the instruction, finds no existing-PG option on the GPU template, and is stuck. It's also not how P-instance capacity actually lands: P5/P6 come via a Capacity Block (placement inherited from the block) or CPG-ACR / a reserved CPG (your own customer-owned CPG), so "share one auto-created PG across CPU + GPU" isn't a real pattern to point people at. The claim predates this PR (the old wording promised the same span via `OnDemandPlacementGroupName`, which also only fed the CPU CNG); the rewrite is a good moment to correct it.

**Suggestion:** scope the example to what's actually wireable:

```suggestion
**Placement group:** auto-created per-CNG by the template. To launch the CPU CNG
into an existing placement group instead (e.g. to share one PG across multiple
CPU CNGs), deploy it with the modular
[`add-cng.yaml`](./assets/add-cng.yaml) and set its `PlacementGroupName`
parameter — deploy-all always auto-creates. (The P5/P6 GPU templates don't
expose an existing-placement-group option: a Capacity Block carries its own
placement and works as-is, but capacity reserved into your own CPG via CPG-ACR —
which you consume by launching into that CPG — can't be targeted from these
templates yet.)
```

*Ref: `add-cng-p5.yaml` has no `PlacementGroupName` parameter (verified 2026-07-07).*

KeitaW (commented):

One doc-accuracy fix on the placement-group note (inline below).

KeitaW (commented):
## Review Batch 3/3 — Things That Look Great

- **The CFN deletion is complete and clean.** All four `OnDemandPlacementGroupName` sites go together — the `ParameterGroups` entry, the `ParameterLabels` entry, the `Parameters:` definition, and the sole `!Ref` in the `OnDemandCNGStack` call. No dangling `!Ref`, no `Conditions`/`Outputs` reference left behind, so the template stays valid (a partial deletion here would have been a hard `validate-template` failure).
- **Cross-doc discipline.** README, `PARAMETERS.md`, and `tests/hpc-efa-test.md` are all updated in the same PR, and `PARAMETERS.md`'s "sections match the console's parameter groups exactly" promise still holds — its section 5 is split into 5.1/5.2/5.3 to mirror the new `Metadata` groups. Easy to get half-right; you didn't.
- **Real e2e evidence.** The PR reports a concrete on-hardware run (us-east-2, `DirectoryService=OpenLDAP-LoginNode`, CREATE_COMPLETE, a Pyxis container job, an LDAP `uid=10001` user running `srun`) rather than a bare "tested."
- **The new modular-only wording is more honest than what it replaced** (for the CPU case) — it names the exact template and parameter to use.

### Sources
- [Use Capacity Reservations with cluster placement groups](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/cr-cpg.html)
- [Placement strategies for your placement groups](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/placement-strategies.html)
- Verified live 2026-07-07: `add-cng-p5.yaml`/`add-cng-p6-b300.yaml` placement-group (`Condition: UseOnDemand`) + `CapacityReservationSpecification`/`MarketType: capacity-block` logic; `add-cng.yaml` `PlacementGroupName` `Default: ''`; deletion completeness across all four sites.

KeitaW (approved):
Nit. Feel free to merge once you checked the comments.

DaisukeMiyamoto:
@KeitaW Thanks for the review — updated the PR based on your suggestions:

- **Batch 1 (keep `OnDemandPlacementGroupName`): adopted.** Reverted the removal in c1b6bf04 — agreed it's the top-level affordance for launching the EFA-HPC CPU queue into an existing/reserved CPG, and a GPU-side placement-group option may follow. The parameter, its console label, and the `!Ref` forwarding are back exactly as before.
- **Batch 2 (doc-accuracy on the placement-group note): adopted** in 62b740df, following your inline suggestion — the "share one PG across CPU + GPU" example is now scoped to what's actually wireable (multiple CPU CNGs, or a customer-owned CPG holding reserved capacity), with a note that the P5/P6 templates expose no existing-PG option. Applied consistently to README §8.6, PARAMETERS.md, and the template's parameter description.
- **GPU existing-CPG targeting / ODCR-vs-Capacity-Block distinction:** agreed this is a separate follow-up — it lines up with the existing targeted-ODCR item in `docs/ROADMAP.md`, and I'll fold the CPG-ACR angle into it.

Remaining scope of the PR: group 5.1/5.2/5.3 split, `Advanced:` labels on the pinned monitoring params, and the doc fix above. `lint-docs.sh` + `validate-template` pass on the final state (5b89e710).

Issue #1169, opened 2026-07-08
Placement question: "Jupyter on a compute node" sample — `architectures/aws-pcs/` or `3.test_cases/`?


DaisukeMiyamoto:
There have been requests to run Jupyter notebooks on PCS compute nodes, and I
have a sample ready for a PR: Jupyter Lab launched as a Slurm batch job
(scale-from-zero, `--gres`, `--time` auto-termination all work), reached via
SSM port forwarding through the login node (no inbound ports), with a per-job
token for multi-user safety. Verified on a real PCS cluster on both CPU and
GPU queues — the notebook kernel sees exactly the GPUs allocated by `--gres`.

My current draft puts it under `architectures/aws-pcs/docs/` (+ a verification
procedure in `tests/`). But almost nothing in it is PCS-specific — it's really
a "Jupyter on a Slurm cluster" recipe, so `3.test_cases/` (e.g.
`3.test_cases/jupyter/`) may be the better home, where HyperPod Slurm /
ParallelCluster users would also find it — even though so far it's only been
verified on PCS.

- **A: `architectures/aws-pcs/`** — matches where it was verified and the
  existing IAM/SSM docs, but hides a generic Slurm recipe in one architecture.
- **B: `3.test_cases/`** — discoverable and scheduler-generic, but the
  directory is currently training/inference workloads, and PCS is the only
  verified environment so far. If B, I'd also appreciate a suggestion on the
  subdirectory: the existing top levels are framework-named (`pytorch/`,
  `megatron/`, `jax/`), and an interactive tool doesn't fit any of them —
  a new `3.test_cases/jupyter/`? or something broader like `tools/` or
  `interactive/`?

I lean toward A for now (files written portable, so they can move to B later),
but I'd like maintainers' input before opening the PR, since placement decides
its shape.

PR #1172, opened 2026-07-08
feat(aws-pcs): add Jupyter-on-compute-node guide with verification procedure

DaisukeMiyamoto:
## What

Docs-only PR adding a way to run Jupyter Lab on PCS compute nodes:

- `docs/JUPYTER.md` — user guide. Jupyter runs **as a Slurm batch job**
  (scale-from-zero, `--gres`, `--time` auto-termination), reached from a
  workstation browser via **SSM port forwarding through the login node** — no
  inbound ports, covered by the stock `cluster-user-iam.yaml` policy. Per-job
  token + job-ID-derived port make it multi-user-safe.
- `tests/jupyter-notebook-test.md` — verification procedure (Test 15), with a
  framework-level GPU gate: the notebook kernel must see exactly the `--gres`
  allocation and run a CUDA op.
- `tests/README.md` matrix row + README docs link.

## Placement (#1169)

Per #1169, this starts under `architectures/aws-pcs/` (where it was verified);
the files are written Slurm-generic so they can move to `3.test_cases/` later
without rework.

## Verification (real hardware, fresh `deploy-all` clusters, us-east-2, Slurm 25.11)

| Queue / instance | Result |
|---|---|
| `cpu1` c6i.4xlarge, scale-from-zero | server up as a job, token auth (`403` without / JSON with), `/lab` = `200` through the SSM tunnel |
| `gpu-g6` g6.12xlarge, `--gres=gpu:1` | same gates; kernel sees 1× L4 (`device_count=1`), CUDA matmul OK — from both direct `srun` and a `jupyter execute` notebook run |
| `gpu-p5` p5.48xlarge on a Capacity Block, `--gres=gpu:8` | same gates; kernel sees 8× H100 (`CUDA_VISIBLE_DEVICES=0,…,7`); 32-NIC first-IP binding works unchanged |

Standard E2E gates (monitoring containers, Pyxis container job, DCGM on GPU
nodes) passed on the same clusters. `tests/lint-docs.sh` passes.

**Note:** this PR also updates the `tests/README.md` region-coverage matrix (Ohio row: *GPU verified* → `✅ P5`, date → 2026-07-08) — a byproduct of the p5.48xlarge verification run above, called out here for traceability.

### DIFF
--- a/architectures/aws-pcs/docs/JUPYTER.md
+++ b/architectures/aws-pcs/docs/JUPYTER.md
@@ -88,24 +88,12 @@ openssl rand -hex 24 > "$TOKEN_FILE"
 
 cat <<EOM
 ======================================================================
-Jupyter starting on $(hostname) ($NODE_IP), port $PORT.
-
-1. From your workstation, forward localhost:8888 to the server:
-
-  LOGIN_ID=\$(aws ec2 describe-instances \\
-    --filters "Name=tag:Name,Values=*login" \\
-              "Name=instance-state-name,Values=running" \\
-    --query 'Reservations[0].Instances[0].InstanceId' --output text)
-
-  aws ssm start-session --target \$LOGIN_ID \\
-    --document-name AWS-StartPortForwardingSessionToRemoteHost \\
-    --parameters host=$NODE_IP,portNumber=$PORT,localPortNumber=8888
-
-2. Get the token (on the login node):  cat $TOKEN_FILE
-
-3. Open:  http://localhost:8888/?token=<token>
-
-Stop the server with:  scancel $SLURM_JOB_ID
+Jupyter is running on compute node $NODE_IP, port $PORT (job $SLURM_JOB_ID).
+To connect, follow "Step 3" in JUPYTER.md with these values:
+    NODE_IP = $NODE_IP
+    PORT    = $PORT
+    token   = run  cat $TOKEN_FILE  on the login node
+Stop the server:  scancel $SLURM_JOB_ID
 ======================================================================
 EOM
 
@@ -121,69 +109,41 @@ Enroot/Pyxis install).
 
 ## Step 3 — connect
 
-### 3a. Read the connection details from the job log (on the login node)
+The job log (`~/<user>-jupyter-<jobid>.log`, printed by the running job) gives
+you the compute node's `NODE_IP` and `PORT`. With those two values:
+
+**1. On your workstation** — open the SSM tunnel (fill in your Region and the
+`NODE_IP` / `PORT` from the log):
 
 ```bash
-cat ~/<username>-jupyter-<jobid>.log   # %u = your username, e.g. ubuntu-jupyter-3.log
-```
-
-The log prints the compute node IP and port (e.g. `10.1.8.103`, port `8003`)
-and reminds you where the token file is.
-
-### 3b. Open the SSM port-forward tunnel (on your workstation)
-
-```bash
-# 1. Find the login-node instance ID
-LOGIN_ID=$(aws ec2 describe-instances \
-  --region <region> \
+LOGIN_ID=$(aws ec2 describe-instances --region <region> \
   --filters "Name=tag:Name,Values=*login" \
             "Name=instance-state-name,Values=running" \
   --query 'Reservations[0].Instances[0].InstanceId' --output text)
 
-# If you run more than one cluster in the same account, pin by stack name:
-# --filters "Name=tag:Name,Values=*login" \
-#           "Name=tag:ClusterName,Values=<your-stack-name>" \
-#           "Name=instance-state-name,Values=running"
-
-# 2. Forward localhost:8888 → compute node
-aws ssm start-session \
-  --region <region> \
-  --target "$LOGIN_ID" \
+aws ssm start-session --region <region> --target "$LOGIN_ID" \
   --document-name AWS-StartPortForwardingSessionToRemoteHost \
   --parameters "host=<NODE_IP>,portNumber=<PORT>,localPortNumber=8888"
 ```
 
-The session stays in the foreground; **Ctrl-C closes the tunnel** (the Jupyter
-job keeps running — reconnect any time until the job ends).
+Leave this running — **Ctrl-C closes the tunnel** (the Jupyter job keeps
+running; reconnect any time until the job ends).
 
-### 3c. Get the token and open the browser
-
-On the login node:
+**2. On the login node** — read the token:
 
 ```bash
 cat ~/.jupyter-token-<jobid>
 ```
 
-Then open in your browser: `http://localhost:8888/?token=<token>`
+**3. In your browser** — open `http://localhost:8888/?token=<token>`.
 
-### Required IAM permissions
-
-The two AWS CLI calls above require the following actions, all of which are
-already included in the stock
-[`cluster-user-iam.yaml`](../assets/cluster-user-iam.yaml) policy:
-
-| Call | Action(s) required |
-|---|---|
-| `aws ec2 describe-instances` | `ec2:DescribeInstances`, `ec2:DescribeInstanceStatus`, `ec2:DescribeTags` |
-| `aws ssm start-session` (instance target) | `ssm:StartSession` on `ec2:instance/*` with condition `ssm:resourceTag/Name = PCS-login*` |
-| `aws ssm start-session` (document) | `ssm:StartSession` on `arn:aws:ssm:*:*:document/AWS-StartPortForwardingSessionToRemoteHost` |
-| Session lifecycle | `ssm:DescribeSessions`, `ssm:GetConnectionStatus`, `ssm:DescribeInstanceProperties`, `ssm:TerminateSession` |
-
-If you are using the cluster-admin policy (`cluster-admin-iam.yaml`) or an
-account with broad permissions, these are implicitly covered. If you are
-operating under a restricted IAM role, attach the `cluster-user-iam.yaml`
-stack output policy (or an equivalent inline policy) before running the
-commands above.
+> **Required IAM.** The two CLI calls above are already permitted by the stock
+> [`cluster-user-iam.yaml`](../assets/cluster-user-iam.yaml) policy (and by any
+> broader admin credentials). Under a restricted role, ensure it grants:
+> `ec2:DescribeInstances` (find the login node); `ssm:StartSession` on the
+> login instance (`ssm:resourceTag/Name = PCS-login*`) and on the
+> `AWS-StartPortForwardingSessionToRemoteHost` document; and
+> `ssm:TerminateSession` / `ssm:DescribeSessions` to end the tunnel.
 
 ## Stopping
 
@@ -229,9 +189,7 @@ How the GPU allocation behaves:
   plenty for most exploration) and keep `--time` tight — an idle notebook
   holds its GPUs until the job ends. For multi-hour *training*, prefer a
   batch job over a notebook so the GPUs free up when the run finishes.
-- **Set `HF_HOME=/fsx/$USER/.hf-cache`** (e.g. in the sbatch script before
-  starting Jupyter, or per notebook) — model downloads must not go to NFS
-  `/home`, and a per-user path avoids cache-ownership clashes between users.
+  (Remember `HF_HOME=/fsx/$USER/.hf-cache` from Step 1 for model downloads.)
 - **Multi-NIC GPU nodes (P5/P6) work as-is.** On these instances `hostname -I`
   returns dozens of addresses (one per EFA NIC). The script binds Jupyter to
   the first entry, which in practice is the primary interface's IP (verified
--- a/architectures/aws-pcs/docs/JUPYTER.md
+++ b/architectures/aws-pcs/docs/JUPYTER.md
@@ -121,17 +121,69 @@ Enroot/Pyxis install).
 
 ## Step 3 — connect
 
+### 3a. Read the connection details from the job log (on the login node)
+
 ```bash
-# On the login node: connection instructions are in the job log
-cat ~/ubuntu-jupyter-<jobid>.log      # (%u = your username)
+cat ~/<username>-jupyter-<jobid>.log   # %u = your username, e.g. ubuntu-jupyter-3.log
 ```
 
-Run the `aws ssm start-session …` command from the log on your **local
-workstation**, read the token (`cat ~/.jupyter-token-<jobid>` on the login
-node), then open `http://localhost:8888/?token=<token>` in your browser.
+The log prints the compute node IP and port (e.g. `10.1.8.103`, port `8003`)
+and reminds you where the token file is.
 
-The SSM session stays in the foreground; Ctrl-C closes the tunnel (the
-Jupyter job keeps running — reconnect any time until the job ends).
+### 3b. Open the SSM port-forward tunnel (on your workstation)
+
+```bash
+# 1. Find the login-node instance ID
+LOGIN_ID=$(aws ec2 describe-instances \
+  --region <region> \
+  --filters "Name=tag:Name,Values=*login" \
+            "Name=instance-state-name,Values=running" \
+  --query 'Reservations[0].Instances[0].InstanceId' --output text)
+
+# If you run more than one cluster in the same account, pin by stack name:
+# --filters "Name=tag:Name,Values=*login" \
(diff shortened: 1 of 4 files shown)


architectures/aws-pcs/docs/JUPYTER.md
```diff
@@ -0,0 +1,207 @@
...
+openssl rand -hex 24 > "$TOKEN_FILE"
+
+cat <<EOM
+======================================================================
+Jupyter starting on $(hostname) ($NODE_IP), port $PORT.
+
+1. From your workstation, forward localhost:8888 to the server:
+
+  LOGIN_ID=\$(aws ec2 describe-instances \\
+    --filters "Name=tag:aws:pcs:compute-node-group-name,Values=login" \\
```
KeitaW: ### `docs/JUPYTER.md`: the login-node discovery `--filters` tag returns no instances on real PCS clusters, so the connect step stalls

**Observation:** the sbatch heredoc tells the user to find the login instance with `LOGIN_ID=$(aws ec2 describe-instances --filters "Name=tag:aws:pcs:compute-node-group-name,Values=login" …)`. I ran this against live PCS clusters and it returns **nothing**: no running PCS instance carries an `aws:pcs:compute-node-group-name` tag. PCS emits `aws:pcs:cluster-id` and `aws:pcs:compute-node-group-id` (the group *ID*, e.g. `pcs_e6u3zq02ol`), plus `Name` — but not `…-group-name`. Since `aws:`-prefixed tags are AWS-applied only, a cluster can't add it either.

**Impact:** `LOGIN_ID` comes back empty, so the very next `aws ssm start-session --target $LOGIN_ID …` has no target and the connect step — the whole point of the guide — fails. I confirmed the tag is absent on **all five PCS clusters in this account across three regions**, including one whose nodes are named `PCS-login`/`PCS-cpu1` (i.e. it looks `deploy-all`-provisioned), so this isn't a one-off. The same filter is used elsewhere in the aws-pcs reference (README, `IAM.md`, `USER-MANAGEMENT.md`, `setup-directory.sh`), so this is broader than this PR — but the PR adds a fresh copy of it here.

**Suggestion:** key on a tag that PCS actually emits — the `Name` convention (`<cluster>-login`, `PCS-login`) or `aws:pcs:compute-node-group-id`. The line below already carries the `instance-state-name` filter, so only the tag line needs to change:

```suggestion
    --filters "Name=tag:Name,Values=*login" \\
```

Could you double-check what your verified clusters tag the login node with? If your `deploy-all` templates were adding `aws:pcs:compute-node-group-name` themselves it would explain the mismatch — but it isn't present on any live cluster I can see, so a `Name`-based (or `aws:pcs:compute-node-group-id`) filter is the portable choice. *Verified live 2026-07-09.*

architectures/aws-pcs/tests/jupyter-notebook-test.md
```diff
@@ -0,0 +1,227 @@
...
+### 2b. Browser path from the workstation
+
+Via SSM (AWS; `LOGIN_ID` = login-node instance ID) — run locally:
+
+```bash
+aws ssm start-session --target $LOGIN_ID \
+  --document-name AWS-StartPortForwardingSessionToRemoteHost \
+  --parameters host=$NODE_IP,portNumber=$PORT,localPortNumber=8888
+# other terminal:
+curl -s -o /dev/null -w '%{http_code}\n' "http://localhost:8888/lab?token=$(TOKEN)"
```
KeitaW: ### `tests/jupyter-notebook-test.md` Step 2b: `$(TOKEN)` is command substitution, so the "expect 200" check never passes

**Observation:** the browser-path check runs `curl … "http://localhost:8888/lab?token=$(TOKEN)"`. `$(TOKEN)` is a command substitution — bash tries to run a command named `TOKEN`, finds none, and substitutes the empty string, so the request goes out as `…/lab?token=` (no token).

**Impact:** JupyterLab won't authenticate a tokenless `/lab` request, so the step returns `403`/redirect, not the documented `200` — a tester following the runbook literally sees the check "fail" and may conclude the feature is broken when it isn't. (I ran it live with a *valid* token and `/lab?token=<valid>` does return `200`, so the expectation is right — only the `$(TOKEN)` literal is wrong.)

**Suggestion:** the token lives on the cluster's shared `$HOME` (read on the login node in Step 2a) — but this step runs on the **workstation**, where that file isn't present, so `$(cat $HOME/.jupyter-token-$JOBID)` wouldn't resolve here either. Reference a variable the tester exports from the token they already read:

```suggestion
  curl -s -o /dev/null -w '%{http_code}\n' "http://localhost:8888/lab?token=$TOKEN"   # TOKEN=<value from `cat ~/.jupyter-token-$JOBID` on the login node>
```

architectures/aws-pcs/tests/jupyter-notebook-test.md
```diff
@@ -0,0 +1,227 @@
...
+executed by the Jupyter machinery (`jupyter execute` uses the same kernel the
+browser would):
+
+```bash
+# direct
+srun --jobid=<gpu-jobid> --overlap $HOME/jupyter-env/bin/python $HOME/gpu_check.py
+
+# as a notebook: one code cell containing  exec(open('/home/.../gpu_check.py').read())
+srun --jobid=<gpu-jobid> --overlap $HOME/jupyter-env/bin/jupyter execute \
+  --output=gpu_check_out.ipynb $HOME/gpu_check.ipynb
```
KeitaW, line 215: ### `tests/jupyter-notebook-test.md` Step 3a: the `jupyter execute` check references `gpu_check.ipynb`, but no step creates it

**Observation:** Step 3a saves `gpu_check.py`, then runs `jupyter execute … $HOME/gpu_check.ipynb` — but the `.ipynb` only exists as a description in a comment ("one code cell containing `exec(open('/home/.../gpu_check.py').read())`"). No command materializes it, and a `.ipynb` is nbformat JSON, not something you can produce by echoing Python source. The inner path `/home/.../gpu_check.py` is also a placeholder that won't match `$HOME/gpu_check.py` if copied.

**Impact:** the notebook-executed run is the more valuable half of the GPU gate — it proves the GPU is visible *through the Jupyter kernel machinery the browser would use*, not just inside the raw allocation (which the direct `srun … python gpu_check.py` already covers). As written, that half is unrunnable, so the end-to-end "kernel sees the GPU" claim isn't reproducible from the doc alone. (The direct-`srun` path still works, so this doesn't block the whole test.)

**Suggestion:** show how to create the one-cell notebook so the step is runnable — e.g. `jupytext --to notebook` from a `# %%`-tagged `gpu_check.py`, or a small `python -c` that writes the nbformat JSON, or `jupyter nbconvert --to notebook --execute`. Whatever you pick, make the cell's `open(...)` path match where `gpu_check.py` was saved (`$HOME/gpu_check.py`).

KeitaW (commented):
## Review Batch 1/2 — Operational Correctness & Verification Procedure

Thanks for this — it's a polished, genuinely useful guide, and the design (Jupyter as a Slurm job, browser access via SSM port-forward through the login node, token-as-user-boundary) is sound. I verified the load-bearing claim that the stock `cluster-user-iam.yaml` already permits the flow — the port-forward documents sit in a separate *untagged* statement, so they don't fail closed. I also **ran the CPU procedure end-to-end on a live PCS cluster** (Slurm 25.11) using your sbatch script verbatim, and every documented gate passed: job runs as a Slurm job, the job-ID-derived port resolves, the token file is created mode `600` and kept out of the job log, `403` without token, JSON + `200` with token, and `/lab?token=<valid>` = `200`.

Three should-fixes follow. The most impactful is **F1** — the login-node discovery command returned no instances on every PCS cluster I checked, which would stall a real user at the connect step. The other two are in the verification runbook.

KeitaW (commented):
## Review Batch 2/2 — Documentation Accuracy & Multi-User Notes

One more should-fix (inline) plus a few minor notes, then what stood out.

### Minor notes (nits — take or leave)

- **`HF_HOME=/fsx/.hf-cache` is a single shared path on a multi-user cluster** (`docs/JUPYTER.md`). The first user to create it owns it; later users may hit permission-denied writes or share a cache. Since the guide targets multi-user clusters, consider a per-user default like `/fsx/$USER/.hf-cache` (or note the shared-cache ownership assumption).
- **`NODE_IP=$(hostname -I | awk '{print $1}')` — "the first entry is always the primary interface's IP"** (`docs/JUPYTER.md`). `hostname -I` order isn't guaranteed primary-first across kernels/NIC layouts; you verified it holds on the 32-NIC p5.48xlarge (nice), and even an EFA-IP-first case stays same-VPC-reachable, so it works in practice — I'd just soften "always" or pick the default-route IP for robustness.
- **`device_count() == --gres count` assumes Slurm device cgroups are on** (`tests/jupyter-notebook-test.md`). The gate holds only under `ConstrainDevices=yes` (your g6 run empirically confirms it — `device_count=1` on a 4-GPU node). Since the runbook advertises portability to "any Slurm cluster," one line stating that assumption would keep it from silently over-counting on a cluster without device constraints.
- **The sbatch script is duplicated** in `docs/JUPYTER.md` and `tests/jupyter-notebook-test.md` (the test's copy is a trimmed variant). Any later fix to the shared logic — the port comment, for one — needs to land in both; a one-line "same script as the guide, keep in sync" note (the test already hints at this) or single-sourcing avoids drift.
- **`tests/README.md` Ohio row** flips to `✅ P5` / `2026-07-08` — a legitimate byproduct of this PR's p5.48xlarge verification, but it's a region-compatibility-matrix change distinct from the Jupyter feature and isn't mentioned in the PR description. Worth a one-line callout there so the matrix edit is traceable.

## Things That Look Great

- **The load-bearing IAM claim is correct, and correctly reflects the tag-scoping fix.** I verified `cluster-user-iam.yaml`: the SSM session documents (`AWS-StartSSHSession`, `AWS-StartPortForwardingSession`, `AWS-StartPortForwardingSessionToRemoteHost`, `SSM-SessionManagerRunShell`) sit in their own *untagged* statement, while the login-node instance is tag-gated (`PCS-login*`) separately — exactly the split that keeps the port-forward from failing closed. The "already permitted by the stock policy, no inbound ports" claim holds.
- **The security model is sound and the ordering is right — confirmed on hardware.** `umask 077` precedes the `openssl > $TOKEN_FILE` write, so the token really is created mode `600`; it's deliberately kept out of the job log; and the guide is honest that the token — not the SSM/IAM layer — is the per-user boundary. I ran it live and the token file came out `600` with the token absent from the log, exactly as documented.
- **The core procedure reproduces on an independent cluster.** I ran the CPU path end-to-end on a different PCS cluster than the ones in your verification table (Slurm 25.11), using your sbatch script verbatim on the `cpu1` queue. Job ran as a Slurm job, the job-ID-derived port resolved (`8001` for job `1`), and all auth gates matched the doc — `403` without token, `{"connections":0,"kernels":0,…}` + `200` with token, `/lab?token=<valid>` = `200`. The guide travels well beyond its home deployment (F1 aside).
- **The GPU gate is a real evaluation, not enumeration.** Asserting `device_count == --gres count`, checking `CUDA_VISIBLE_DEVICES`, *and* running an actual matmul — from both a direct `srun` and a notebook-executed run — verifies compute executes on the allocated device, which is exactly the right bar for a GPU-visibility claim.
- **Verification is on real hardware with raw specifics** — dated runs across `c6i.4xlarge`, `g6.12xlarge` (4× L4), and `p5.48xlarge` (8× H100, 32 NICs) on fresh `deploy-all` clusters, with exact versions and the standard E2E gates noted as also passing.
- **Nice real-world details:** the `/fsx` `noexec` catch (keep the sbatch under `$HOME`), the multi-user `pam_mkhomedir` prerequisite, `--overlap` to attach without consuming the job's resources, and the portability framing with explicit prerequisites.

## Sources

- `AWS-StartPortForwardingSessionToRemoteHost` forwards from the target instance to a remote host inside its network — https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html
- Jupyter Server token auth (`/api/status` 403 without token; `--ServerApp.token`) — https://jupyter-server.readthedocs.io/en/latest/operators/security.html
- `jupyter execute` operates on nbformat `.ipynb` files (nbclient) — https://nbclient.readthedocs.io/
- **Verified live 2026-07-09**: ran the CPU procedure end-to-end on a PCS cluster (Slurm 25.11) — all gates green as above; and confirmed `aws:pcs:compute-node-group-name` is absent on all live PCS clusters checked across three regions (F1).

architectures/aws-pcs/docs/JUPYTER.md
```diff
@@ -0,0 +1,207 @@
...
+#!/bin/bash
+#SBATCH --job-name=jupyter
+#SBATCH --partition=cpu1          # or your GPU queue; then add e.g. --gres=gpu:1
+#SBATCH --nodes=1
+#SBATCH --time=8:00:00            # auto-terminate after 8 h — adjust to taste
+#SBATCH --output=%u-jupyter-%j.log
+
+umask 077   # everything this job writes is owner-only
+
+# Port derived from the job ID → no collisions between concurrent servers
```
KeitaW: ### `docs/JUPYTER.md`: the job-ID port derivation aliases every 1000 IDs, so "no collisions between concurrent servers" is stronger than it holds

**Observation:** `PORT=$((8000 + SLURM_JOB_ID % 1000))` maps job IDs into `8000–8999`, so IDs congruent mod 1000 (e.g. 1500 and 2500) get the same port. The comment promises "no collisions between concurrent servers."

**Impact:** two servers only actually collide when they land on the **same node** *and* compute the same port (same port on different nodes is harmless — different IPs). But the guide explicitly enables that exact case a few lines down — "on multi-GPU nodes Slurm can schedule other jobs (another user's Jupyter…) onto the remaining GPUs." So on a busy shared node the guarantee can break: the second `jupyter lab` fails to bind. Low probability, but the doc invites the scenario.

**Suggestion:** soften the claim (the arithmetic itself is fine — `%` binds tighter than `+`, range is 8000–8999 as intended):

```suggestion
# Port derived from the job ID (range 8000-8999). Collisions are unlikely but
# possible if two servers share a node with job IDs differing by a multiple of 1000.
```

KeitaW (changes requested):
Few comments

DaisukeMiyamoto:
Thanks for the thorough review — especially for running the CPU procedure
end-to-end on an independent cluster and for catching F1 live. All four
should-fixes plus the nits are addressed in 444fa130.

**F1 (login-node discovery tag):** you're right — `aws:pcs:compute-node-group-name`
is never emitted by PCS. I checked a live cluster deployed from these templates
and confirmed the actual tag set (`aws:pcs:cluster-id`,
`aws:pcs:compute-node-group-id`, `Name`, `ClusterName`, …) has no
`…-group-name`. Switched to your suggestion, `"Name=tag:Name,Values=*login"`,
and re-ran the discovery from a workstation — it now returns the login
instance ID. Also added a note that on accounts with several clusters the
filter should be pinned with `"Name=tag:ClusterName,Values=<stack>"`
(I hit exactly that during re-verification). The same stale filter exists in
README.md (3 places) — as you noted that predates this PR, so I'll fix it in a
follow-up PR rather than widen this one.

**F2 (`$(TOKEN)`):** fixed to `$TOKEN` with an inline comment that the value
comes from `cat ~/.jupyter-token-<jobid>` on the login node.

**F3 (missing `gpu_check.ipynb` creation):** added a heredoc that writes the
one-cell nbformat JSON (cell path matches `$HOME/gpu_check.py`), plus a
command to print the executed cell's output from `gpu_check_out.ipynb`.

**F4 (port-collision comment):** took your suggested wording in both copies of
the script.

**Nits:** all taken — per-user `HF_HOME=/fsx/$USER/.hf-cache`; softened the
`hostname -I` "always" claim (kept the p5 empirical note, added the
same-VPC-reachable fallback point); documented the `ConstrainDevices=yes`
assumption behind the `device_count == --gres` gate; added a keep-in-sync note
on the test file's sbatch variant; and called out the Ohio matrix change in
the PR description.

**Re-verification:** after the fixes I did a clean-room pass on a fresh
cluster — every command block in both files executed verbatim in its
documented role (login node vs. workstation), including the new discovery
filter, the workstation `$TOKEN` curl, and the notebook-creation heredoc.
All gates green (CPU + `gpu-g6` `--gres=gpu:1`); recorded as the 2026-07-09
row in the verified-configurations table.

KeitaW (approved):

PR #1173, opened 2026-07-09
Align Kimi example with shared HF-cache pre-staging; harden qwen intra-PD liveness

ybalbert001:
## Purpose

Align Kimi and qwen examples, to be consistent with previous PR

## Changes

- kimi2.6-h200-1p1d: load by repo id (moonshotai/Kimi-K2.6) from the node's HF cache mounted at /root/.cache/huggingface, matching the layout the shared download-model.sh stages — the old /nvme/moonshotai-Kimi-K2.5 path never matched what the downloader produced. hostPath now points at <nvme>/huggingface with DirectoryOrCreate, consistent with the sibling B300 examples; README curl and pre-stage steps updated to match.
- Rename remaining Kimi-K2.5 references to Kimi-K2.6 across the shared downloader docs, finally matching the kimi2.6-* directory name.
- qwen3.5-27b-b300-intra-pd: liveness now ANDs /health of all eight engines via exec probe (same shape as dsv4flash) — a dead background engine previously left the pod Ready at reduced capacity forever; readiness deliberately stays on the foreground engine only. curl -m 3 keeps the worst case (8x3s) under timeoutSeconds. Pre-staging promoted to a required Deploy step (eight engines share the node cache).
- glm5.2: pre-staging is required, not recommended.

## Test Plan

**Environment:**
- AWS Service:
- Instance type:
- Number of nodes:

**Test commands:**
```bash

```

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
+++ b/examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
@@ -205,14 +205,14 @@ spec:
       #   HyperPod-on-EKS (DLAMI):           /opt/dlami/nvme/...
       #   self-managed EKS (RAID0 instance): /mnt/k8s-disks/0/...   (setup-local-disks raid0)
       # This is the node's HF cache (mounted at /root/.cache/huggingface; the
-      # engine loads by repo id) — pre-stage it with ../download-model.sh.
-      # Picking the WRONG path is not a hard error: type=DirectoryOrCreate will
-      # silently create an empty dir on the small root disk, and the 1TB+
-      # weights then download there and can fill it up.
+      # engine loads by repo id) — pre-stage it with ../download-model.sh,
+      # which creates this dir. type=Directory makes a wrong path (or a skipped
+      # pre-stage) fail loudly with FailedMount instead of silently creating an
+      # empty dir on the small root disk and downloading the 1TB+ weights there.
       - name: nvme-hf
         hostPath:
           path: /opt/dlami/nvme/huggingface  # For EKS: /mnt/k8s-disks/0/huggingface
-          type: DirectoryOrCreate
+          type: Directory
       - name: gdrdrv
         hostPath:
           path: /dev/gdrdrv
@@ -367,14 +367,14 @@ spec:
       #   HyperPod-on-EKS (DLAMI):           /opt/dlami/nvme/...
       #   self-managed EKS (RAID0 instance): /mnt/k8s-disks/0/...   (setup-local-disks raid0)
       # This is the node's HF cache (mounted at /root/.cache/huggingface; the
-      # engine loads by repo id) — pre-stage it with ../download-model.sh.
-      # Picking the WRONG path is not a hard error: type=DirectoryOrCreate will
-      # silently create an empty dir on the small root disk, and the 1TB+
-      # weights then download there and can fill it up.
+      # engine loads by repo id) — pre-stage it with ../download-model.sh,
+      # which creates this dir. type=Directory makes a wrong path (or a skipped
+      # pre-stage) fail loudly with FailedMount instead of silently creating an
+      # empty dir on the small root disk and downloading the 1TB+ weights there.
       - name: nvme-hf
         hostPath:
           path: /opt/dlami/nvme/huggingface  # For EKS: /mnt/k8s-disks/0/huggingface
-          type: DirectoryOrCreate
+          type: Directory
       - name: gdrdrv
         hostPath:
           path: /dev/gdrdrv
--- a/examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
+++ b/examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
@@ -143,7 +143,7 @@ spec:
         - -c
         - |
           python3 -m sglang.launch_server \
-            --model-path /nvme/moonshotai-Kimi-K2.5 \
+            --model-path moonshotai/Kimi-K2.6 \
             --host 0.0.0.0 \
             --port 30000 \
             --tp-size 8 \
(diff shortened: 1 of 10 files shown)


examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
```diff
@@ -202,15 +202,17 @@ spec:
...
+      # engine loads by repo id) — pre-stage it with ../download-model.sh.
+      # Picking the WRONG path is not a hard error: type=DirectoryOrCreate will
+      # silently create an empty dir on the small root disk, and the 1TB+
+      # weights then download there and can fill it up.
+      - name: nvme-hf
         hostPath:
-          path: /opt/dlami/nvme  #For EKS: /mnt/k8s-disks/0
-          type: Directory
+          path: /opt/dlami/nvme/huggingface  # For EKS: /mnt/k8s-disks/0/huggingface
+          type: DirectoryOrCreate
```
KeitaW: ### `type: DirectoryOrCreate` converts an operator path mistake into a node-wide disk-fill

The model-cache `hostPath` volume switches from `path: /opt/dlami/nvme` + `type: Directory` to `path: /opt/dlami/nvme/huggingface` + `type: DirectoryOrCreate` (here, and identically in the decode volume below). Your own comment above captures the downside precisely: with `DirectoryOrCreate`, a wrong per-cluster path is *not* a hard error — the kubelet silently creates an empty dir, and since the engine now loads `moonshotai/Kimi-K2.6` by repo id, it misses the cache and downloads the 1TB+ weights into whatever filesystem that path landed on.

**Why it matters:** the old `type: Directory` failed *loud and pod-local* — a wrong path (e.g. the DLAMI default left in place on a self-managed EKS node whose NVMe is actually at `/mnt/k8s-disks/0`) kept the pod `ContainerCreating` with a `FailedMount`, harming nothing else. `DirectoryOrCreate` fails *silent and node-wide*: the mistaken path resolves onto the small root disk, the multi-hundred-GB download fills it, and the node hits `DiskPressure` → eviction cascade to every other pod scheduled there. Same operator error, strictly worse blast radius — and a comment is documentation, not enforcement.

**Suggestion:** since this PR promotes pre-staging to a **required** step, and the `download-model.sh` DaemonSet *creates* `/opt/dlami/nvme/huggingface` on the happy path, `type: Directory` still succeeds whenever pre-staging was done — and fails loudly precisely when someone skips the now-required pre-stage or points at the wrong NVMe path. Consider keeping `type: Directory` in both the prefill and decode volumes for that guard. If `DirectoryOrCreate` is a deliberate choice for byte-consistency with the sibling B300 examples, could you confirm that tradeoff is intended? — those examples predate the "pre-staging required" wording, so the safer primitive may now be the better default across all of them.

Ref: Kubernetes `hostPath` volume-type semantics — `Directory` requires the path to pre-exist, `DirectoryOrCreate` creates an empty dir on miss (https://kubernetes.io/docs/concepts/storage/volumes/#hostpath).

KeitaW (commented):
## Review Batch 1/1 — Deployment Pipeline & Operational Correctness

This is a clean, well-reasoned follow-up that (a) migrates the Kimi PD example to the same load-by-repo-id / shared-HF-cache pattern as the sibling B300 examples and (b) hardens the qwen intra-PD liveness to AND all eight engines. I verified the load-bearing claims — the 8-port topology, the probe timing budget, and the `/health` endpoint choice — and they all hold. One should-fix worth a look before merge (inline on the Kimi manifest): the switch to `DirectoryOrCreate` trades a fail-loud mount guard for a fail-silent root-disk-fill, and with pre-staging now "required" the stricter primitive is essentially free.

### Things That Look Great

- **The qwen liveness rework correctly closes the multi-engine health-gating gap.** A single-container pod running eight engine processes can only `httpGet`-probe one port, so a dead background engine previously left the pod `Ready` at reduced capacity indefinitely. ANDing all eight `/health` ports in the exec probe means any dead engine now restarts the whole PD group — and keeping **readiness** on the foreground prefill only (so one dead background engine doesn't instantly flip a `replicas: 1` pod `NotReady` and stop *all* traffic during the restart window) is exactly the right split. The rationale comments on both probes are excellent — they name the tradeoff instead of hiding it.
- **The probe timing budget checks out.** `curl -sf -m 3` hard-caps each call at 3s via `--max-time` and `|| exit 1` bails on the first failure, so the worst case is 8 × 3s = 24s, comfortably under `timeoutSeconds: 30`. Using `127.0.0.1` also sidesteps any DNS in the probe path.
- **Port topology matches the documented engine layout.** The probe's `30000 30001 30002 30003 30004 30005 30010 30011` is exactly the README's "prefill `30000-30005`, decode `30010-30011`" — 6 + 2 = 8 engines against `nvidia.com/gpu: 8`. No off-by-one.
- **Right `/health` endpoint choice.** Probing `/health` (a pure process-liveness check that returns 200 whenever the server is up) rather than SGLang's load-sensitive `/health_generate` avoids restarting a healthy-but-busy engine under backpressure.
- **Cross-doc consistency is tight.** The `Kimi-K2.5 → Kimi-K2.6` rename lands consistently across the top-level README, the daemonset comment, `download-model.sh`, the per-example README, and the manifest `--model-path`; the curl example's `"model": "moonshotai/Kimi-K2.6"` matches the new repo-id model-path (SGLang defaults the served name to the model-path). The qwen README's new "required" pre-stage step and the updated Notes section reference each other correctly.
- **Honest documentation of the footgun.** Even though the inline note argues for reverting it, the rewritten comment that spells out the `DirectoryOrCreate` disk-fill failure mode is the right instinct — it's exactly the kind of "here's what bites you" note the repo's best examples carry.

### Sources

- Kubernetes `hostPath` volume types (`Directory` vs `DirectoryOrCreate`): https://kubernetes.io/docs/concepts/storage/volumes/#hostpath
- `curl(1)` `--max-time` / `-f` / `-s` semantics: https://curl.se/docs/manpage.html
- SGLang server `/health` vs `/health_generate` endpoints: https://docs.sglang.ai/backend/native_api.html

examples/inference/sglang/kimi2.6-h200-1p1d/manifests/kimi-pd-deploy.yaml
ybalbert001: make sense, updated.

PR #1175, opened 2026-07-09
fix(aws-pcs): make ldap-add-user.sh's ssh-key argument actually work

DaisukeMiyamoto:
## What

The optional 4th argument (SSH public key) of `ldap-add-user.sh` was broken:
it wrote `objectClass: ldapPublicKey` into the LDIF, but `setup-directory.sh`
never loads the openssh-lpk schema, so `ldapadd` failed with `Invalid syntax
(21)` — **the user wasn't created at all**. Even a stored key would have been
unused (no `AuthorizedKeysCommand`/SSSD wiring on the ssh side).

Fix: install the key into the user's `authorized_keys` on shared `/home`
instead. Same one-command admin flow, works cluster-wide immediately, no LDAP
schema or sshd changes:

```bash
sudo LDAP_ADMIN_PASSWORD="$ADMIN_PW" ldap-add-user.sh alice 10001 3000 "ssh-ed25519 AAAA..."
```

The home directory is created on the spot from `/etc/skel` (pre-empting
`pam_mkhomedir`) with correct ownership and modes (750 home / 700 `.ssh` /
600 `authorized_keys`). Without the key argument, behavior is unchanged.
`username`/`uid`/`gid` are validated up front, so a malformed value (e.g. a
`..`-containing name from the batch-add loop) errors cleanly instead of
letting `chown -R`/`cp` act on the wrong path.

Script + docs only — no CloudFormation template changes. `ldap-add-user.sh`
is copied from the templates S3 bucket to `/usr/local/bin/` by the login
node's first-boot directory setup, so the fix reaches a cluster when its
login node boots **after** the updated script is published to the bucket —
new deployments and login-node replacements get it automatically; a login
node that is already running keeps the old copy until it is replaced (no
stack update required either way).

## Testing

On fresh multi-user clusters (`DirectoryService=OpenLDAP-LoginNode`, us-east-2),
including one deployed with the fixed script from a private bucket so the
boot-time install path itself was exercised:

- one command creates the LDAP user **and** installs the key (exit 0)
- key login works end-to-end: `ssh -i <key> user@login` → correct uid/gid;
  skel files present; modes/ownership as above
- `srun -p cpu1 id <user>` resolves on a compute node
- input validation: `../../etc` and a non-numeric uid are rejected; a valid
  user still creates successfully
- `tests/lint-docs.sh` passes


architectures/aws-pcs/assets/scripts/ldap-add-user.sh
```diff
@@ -48,25 +48,42 @@ userPassword: {SSHA}placeholder
...
 ldappasswd -x -H ldap://localhost -D "${LDAP_ADMIN_DN}" -w "${LDAP_ADMIN_PASSWORD}" \
     -s "${INITIAL_PW}" "uid=${USERNAME},ou=People,${LDAP_DOMAIN_SUFFIX}"
 
+# SSH public key (optional 4th arg): install into the user's authorized_keys
+# on shared /home (visible to every node). This intentionally does NOT use the
+# LDAP openssh-lpk schema — slapd here doesn't load it, and authorized_keys on
+# the shared filesystem gives the same result with zero sshd/SSSD wiring.
+# Requires root (this script already needs root to be useful).
+if [ -n "${SSH_PUBKEY}" ]; then
+    HOME_DIR="/home/${USERNAME}"
```
KeitaW, line 82: ### An unvalidated `USERNAME` flows into `chown -R`/`cp` filesystem paths; a malformed value (e.g. from the batch-add loop) can act on the wrong tree

**Observation:** `USERNAME` (positional `$1`) is used verbatim to build `HOME_DIR="/home/${USERNAME}"`, which then drives `cp -rT /etc/skel "${HOME_DIR}"`, `mkdir -p`, and `chown -R "${USER_UID}:${USER_GID}" "${HOME_DIR}"`. There's no check that it's a plain POSIX name.

**Impact:** for a well-formed name this is fine (I confirmed the happy path live). But a value containing `..` resolves the path elsewhere — `USERNAME=../../etc` makes `HOME_DIR=/home/../../etc` = `/etc`, and `chown -R 20099:3000 /etc` (as root) would break the node. This is a root-run tool, so it's not a privilege-escalation hole — but the docs add a **batch-add loop** that feeds usernames from a here-doc/CSV (`while read user uid gid …`), and one malformed row would run the destructive `chown -R` against the wrong tree with no warning. A one-line guard is cheap, legitimate input validation, and turns a silent foot-gun into a clean error.

**Suggestion:** validate before the name reaches a path (guarding the LDAP DN too if you move it up near the `USERNAME=` read):

```suggestion
    [[ "${USERNAME}" =~ ^[a-z_][a-z0-9_-]*$ ]] || { echo "Invalid username: '${USERNAME}'" >&2; exit 1; }
    HOME_DIR="/home/${USERNAME}"
```

Reasonable to decline if you'd rather trust the operator — your call — but it's near-free.

KeitaW (commented):
## Review — Script Robustness & Documentation

Thanks for this fix — it's clean and well-diagnosed, and I **validated it end-to-end on a live OpenLDAP PCS cluster** (Slurm 25.11): the patched script created the LDAP user, installed the key into `/home/<user>/.ssh/authorized_keys` with correct modes/ownership (750/700/600, owned by the numeric uid:gid), copied `/etc/skel`, resolved via SSSD, and an **actual SSH key login succeeded** — the exact flow the old `objectClass: ldapPublicKey` path broke (it made `ldapadd` fail with Invalid syntax and created no user at all). The root-cause analysis in the description is spot-on. No blocking issues — one should-consider inline, plus a few doc nits.

### Minor notes (nits — take or leave)

- **`ssh-rsa` / `ssh-ed25519` examples are now mixed.** The PR modernizes the two primary examples to `ssh-ed25519` (the table row and the "With an SSH key" block), but two spots still say `ssh-rsa`: the script's own header comment (`ldap-add-user.sh` line 11, `"ssh-rsa AAAA..."`) and the batch-add CSV block in `USER-MANAGEMENT.md` (lines 266–267). Both are on lines this PR doesn't touch, so purely informational — but if the intent was to standardize on ed25519, those two are worth the same swap.
- **The fix depends on `/home` not root-squashing.** The whole key-install path relies on `chown -R` running as root against the shared `/home`. I verified live that it works on the reference OpenZFS `/home` (the files ended up owned by the target uid, not squashed to `nobody`). Worth one line in `USER-MANAGEMENT.md` noting the dependency, so an operator who points the directory at a root-squashed NFS export understands why the key install would then `EPERM` mid-run.
- **Partial-state on a mid-run failure is fail-loud but leaves cleanup to the operator.** With `set -euo pipefail`, if `ldappasswd` fails the script aborts *after* the LDAP entry exists but *before* the key is installed — the user is in LDAP with the `{SSHA}placeholder` password and no `authorized_keys` (locked out), and a re-run then aborts at `ldapadd` with "Already exists (68)". That's acceptable for a root-run admin helper (please **don't** add rollback/traps); a one-line "on failure, delete the partial `uid=…` entry before re-running" note in the docs would just save an operator a confused minute.

## Things That Look Great

- **The root-cause fix is correct and I confirmed it end-to-end.** On a live OpenLDAP cluster the patched script created the user, installed the key (`authorized_keys` mode 600, owned by the uid), and an actual SSH key login succeeded — the flow the old `objectClass: ldapPublicKey` path broke entirely. Moving to `authorized_keys` on shared `/home` is the right, schema-free approach.
- **Numeric `chown "${USER_UID}:${USER_GID}"` is the right call, not a smell.** The new LDAP user isn't in NSS yet when the script runs, so a name-based `chown ${USERNAME}` would fail under `set -e`; using the numeric ids sidesteps the lookup race. (Verified live: SSSD only resolved the name *after* creation.)
- **Permissions are StrictModes-correct.** `750` home / `700` `.ssh` / `600` `authorized_keys`, all owned by the user after the `chown` — which is exactly why the loopback SSH login I ran was accepted rather than refused by sshd for loose modes.
- **The success message was made honest.** The `Home: …` line is now conditional — with a key it prints "(created; SSH public key installed)", without it keeps the `pam_mkhomedir` wording. Live output confirmed the right branch printed.
- **The script already reflects earlier review feedback** — `set -euo pipefail`, the required-`uid` guard with the RANDOM-collision rationale, and the `sed -n 's/^ldap_search_base…//p'` suffix detection (a single-delimiter parse, not `awk -F=`) are all in place.
- **`cp -rT /etc/skel` is the correct idiom** (copies skel *contents* into the home, not a nested `skel/` dir), guarded by `[ ! -d "$HOME_DIR" ]` so it won't clobber an existing home. Verified the `.bashrc` landed.

## Sources

- OpenSSH `sshd` `StrictModes` / `AuthorizedKeysFile` (why 750/700/600 + user ownership is required) — https://man.openbsd.org/sshd_config#StrictModes
- OpenLDAP `ldapadd` / `ldappasswd` — https://www.openldap.org/software/man.cgi?query=ldapadd
- **Verified live 2026-07-09**: ran the patched `ldap-add-user.sh <user> 20099 3000 "<ed25519 key>"` on a live OpenLDAP PCS cluster (Slurm 25.11) as root → exit 0; LDAP entry created; `/home/<user>` 750 + `.ssh` 700 + `authorized_keys` 600 all owned by the target uid (chown on shared `/home` succeeded — no root_squash); key present; `/etc/skel` `.bashrc` copied; `id <user>` resolved via SSSD; end-to-end SSH key login (loopback) succeeded; throwaway user + home fully cleaned up. sshd config: `AuthorizedKeysFile`/`AuthorizedKeysCommand` at defaults (reads `~/.ssh/authorized_keys`), `StrictModes yes`.

KeitaW (approved):
Approving with nit

PR #1176, opened 2026-07-10
fix(aws-pcs): correct EC2 tag filter for locating the login node in docs

DaisukeMiyamoto:
## What

README §6 (Accessing the Cluster) and §8.2 (Monitoring — Option A/B) tell
users to locate the login node by filtering on

```
Name=tag:aws:pcs:compute-node-group-name,Values=login
```

but **PCS does not emit that tag**. What PCS emits is
`aws:pcs:compute-node-group-id` (value is the internal `pcs_xxx` handle,
not the human-readable CngName). The documented filter matches nothing,
so `describe-instances ... --query 'Reservations[0].Instances[0].InstanceId'`
returns `None` and every follow-up (`aws ssm start-session --target None`,
port-forward, public-IP lookup) fails at the very first copy-paste step.

Fix: switch the three filters (and the console hint) to
`Name=tag:Name,Values=PCS-login`, which uses the stable `Name` tag PCS puts
on the login CNG. Verified on a live cluster:

```
LOGIN=$(aws ec2 describe-instances \
  --filters "Name=tag:Name,Values=PCS-login" "Name=instance-state-name,Values=running" \
  --query 'Reservations[0].Instances[0].InstanceId' --output text)
# → i-0a35438c5c04e8b14 (correct)
```

Also adds a `tests/lint-docs.sh` banned-pattern entry so this tag-key typo
can't reappear silently. The pattern matches the bare key (no `tag:`
prefix), so **both** forms of the typo — the CLI filter
`Name=tag:aws:pcs:compute-node-group-name,...` and the plain console hint
`by aws:pcs:compute-node-group-name = login` — are caught.

## Testing

- Lint passes on the fix.
- Negative tests, both re-injecting the broken **CLI filter** form and the
  bare **console-hint** form, make the lint FAIL with the affected README
  locations.
- Filter resolves the login instance on a running cluster (us-east-2).
- `JUPYTER.md` already used the correct form, no change needed.

### DIFF
--- a/architectures/aws-pcs/tests/lint-docs.sh
+++ b/architectures/aws-pcs/tests/lint-docs.sh
@@ -37,6 +37,11 @@ BANNED=(
   $'DeployMonitoring=true\tMonitoringStack|internally|nested\tDeployMonitoring (bool) was replaced by MonitoringStack at the deploy-all layer'
   $'S3 (public )?hosting is not allowed\tNEVERMATCH\tPostInstallScriptUrl now accepts s3:// URLs'
   $'architectures/aws-pcs/iam/\tNEVERMATCH\tthe iam/ directory was removed; use docs/IAM.md + assets/cluster-*-iam.yaml'
+  # PCS emits aws:pcs:compute-node-group-*id* (the internal pcs_xxx handle), NOT
+  # aws:pcs:compute-node-group-name — a copy-paste filter with the -name key
+  # returns nothing (silently), so the docs example fails at first use.
+  # The stable, human-readable filter is Name=tag:Name,Values=PCS-login.
+  $'tag:aws:pcs:compute-node-group-name\tNEVERMATCH\ttag key aws:pcs:compute-node-group-name does not exist — use Name=tag:Name,Values=PCS-login'
 )
 for entry in "${BANNED[@]}"; do
   IFS=$'\t' read -r pat allow msg <<<"$entry"
--- a/architectures/aws-pcs/tests/lint-docs.sh
+++ b/architectures/aws-pcs/tests/lint-docs.sh
@@ -40,8 +40,11 @@ BANNED=(
   # PCS emits aws:pcs:compute-node-group-*id* (the internal pcs_xxx handle), NOT
   # aws:pcs:compute-node-group-name — a copy-paste filter with the -name key
   # returns nothing (silently), so the docs example fails at first use.
-  # The stable, human-readable filter is Name=tag:Name,Values=PCS-login.
-  $'tag:aws:pcs:compute-node-group-name\tNEVERMATCH\ttag key aws:pcs:compute-node-group-name does not exist — use Name=tag:Name,Values=PCS-login'
+  # Match the bare key (no `tag:` prefix) so both forms of the typo — the CLI
+  # `Name=tag:aws:pcs:compute-node-group-name,...` filter AND the plain
+  # console-hint `by aws:pcs:compute-node-group-name = login` — are caught.
+  # The real tag ends in `-id`, so `-name` never appears legitimately.
+  $'aws:pcs:compute-node-group-name\tNEVERMATCH\ttag key aws:pcs:compute-node-group-name does not exist — use Name=tag:Name,Values=PCS-login'
 )
 for entry in "${BANNED[@]}"; do
   IFS=$'\t' read -r pat allow msg <<<"$entry"
--- a/architectures/aws-pcs/README.md
+++ b/architectures/aws-pcs/README.md
@@ -287,13 +287,13 @@ want direct SSH, set `SSHAccessCidr` to a trusted CIDR at deploy time to open SS
 on the login node to that CIDR.
 
 **Console:** [EC2 Console](https://console.aws.amazon.com/ec2/home#Instances:) → filter
-by `aws:pcs:compute-node-group-name = login` → **Connect** → **Session Manager**.
+by `Name = PCS-login` → **Connect** → **Session Manager**.
 
 **CLI** (AWS CloudShell has the required permissions):
 
 ```bash
 INSTANCE_ID=$(aws ec2 describe-instances \
-  --filters "Name=tag:aws:pcs:compute-node-group-name,Values=login" \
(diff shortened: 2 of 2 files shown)


KeitaW (commented):
## Review Batch 1/2 — Regression-Guard Completeness

architectures/aws-pcs/tests/lint-docs.sh
```diff
@@ -37,6 +37,11 @@ BANNED=(
   $'DeployMonitoring=true\tMonitoringStack|internally|nested\tDeployMonitoring (bool) was replaced by MonitoringStack at the deploy-all layer'
   $'S3 (public )?hosting is not allowed\tNEVERMATCH\tPostInstallScriptUrl now accepts s3:// URLs'
   $'architectures/aws-pcs/iam/\tNEVERMATCH\tthe iam/ directory was removed; use docs/IAM.md + assets/cluster-*-iam.yaml'
+  # PCS emits aws:pcs:compute-node-group-*id* (the internal pcs_xxx handle), NOT
+  # aws:pcs:compute-node-group-name — a copy-paste filter with the -name key
+  # returns nothing (silently), so the docs example fails at first use.
+  # The stable, human-readable filter is Name=tag:Name,Values=PCS-login.
+  $'tag:aws:pcs:compute-node-group-name\tNEVERMATCH\ttag key aws:pcs:compute-node-group-name does not exist — use Name=tag:Name,Values=PCS-login'
```
KeitaW: ### The new lint guard misses the bare console-hint form of the very typo it prevents — 🟡 [confirmed]
**Observation:** The banned pattern is `tag:aws:pcs:compute-node-group-name` (with the `tag:` prefix), and the lint runs `grep -rnE "$pat"` over the docs. But the typo occurs in **two forms** in this very file: the CLI filter `Name=tag:aws:pcs:compute-node-group-name,...` (has `tag:`) *and* the console hint you just fixed on README line 290, which was written **without** the prefix — `by \`aws:pcs:compute-node-group-name = login\``. The regex requires `tag:` immediately before the key, so a reintroduced console-hint typo would slip through lint silently.
**Impact:** The guard only covers the CLI-filter half of the typo class it's named for; the exact console-hint form this PR is removing could come back unnoticed.
**Suggestion:** Drop the `tag:` prefix so the pattern matches both forms. It's safe from false positives — the real tag is `aws:pcs:compute-node-group-**id**`, so `...-name` never appears legitimately, and `:`/`-` are literal in ERE:
```suggestion
  $'aws:pcs:compute-node-group-name\tNEVERMATCH\ttag key aws:pcs:compute-node-group-name does not exist — use Name=tag:Name,Values=PCS-login'
```
(The lint scans `README.md docs/*.md tests/*.md` only, so the broadened pattern still won't self-trigger on this `.sh` source.)

KeitaW (commented):
## Review Batch 2/2 — Pre-existing Notes, Positives & Sources



### Pre-existing (informational — not introduced by this PR, no action needed)
These predate the diff and are neither introduced nor worsened; noting only to confirm they were checked:
- `README.md` login-node lookups use `Reservations[0].Instances[0]` and pass no `--region`. With the `Name=PCS-login` filter and the deploy-all single-login-CNG topology, `[0]` resolving one instance is fine, and the commands rely on the caller's default region. Fine as concrete examples.
- `Values=PCS-login` is specific to the deploy-all path (`CngName: login`). A user who renamed the login CNG would adjust the value — inherent to a concrete example, and consistent with `docs/IAM.md` and the IAM policy scope.

---

### Things That Look Great
- **Correct root-cause, not a guess.** The diagnosis (PCS emits `aws:pcs:compute-node-group-id` with an internal `pcs_xxx` value, not `...-group-name`) is accurate, and the PR body shows it verified live on a real cluster (`i-0a35438c5c04e8b14`). This is the kind of copy-paste-first-step bug that silently returns `None` and breaks every downstream command, so fixing it is high-value.
- **The replacement value is verified-consistent, not arbitrary.** `Name=PCS-login` is exactly what the repo's own CloudFormation sets (`Name=PCS-${CngName}`, `CngName: login`), and it aligns README with `docs/IAM.md:61` and the IAM `ssm:resourceTag/Name = PCS-login*` scope that was already in the tree — so the docs, the IAM policy, and the templates now agree.
- **All four occurrences fixed** (three CLI filters + the console hint), not just the obvious ones — I confirmed no stragglers remain in the aws-pcs docs.
- **Ships a regression guard with the fix.** Adding a `tests/lint-docs.sh` banned-pattern entry (with a clear rationale comment) so the typo can't reappear is exactly the right instinct — the one nit above just makes that guard cover both forms of the typo.

---

### Sources
- Repo files (verified live 2026-07-10 on `origin/main`): `architectures/aws-pcs/assets/add-cng.yaml` / `add-cng-p5|p6-b200|p6-b300.yaml` (`Name: !Sub 'PCS-${CngName}'`), `assets/pcs-ml-cluster-deploy-all.yaml` (`CngName: login`), `assets/cluster-user-iam.yaml` (`ssm:resourceTag/Name: PCS-login*`), `docs/IAM.md:61` (already uses `Name=tag:Name,Values=PCS-login`), `tests/lint-docs.sh` (`DOC_GLOBS=(README.md docs/*.md tests/*.md)`, `grep -rnE`).
- AWS PCS tag behavior corroborated by the reviewer's prior live check (2026-07-09): `aws:pcs:compute-node-group-name` absent on all PCS clusters in the test account; `PCS-login`-named deploy works.

DaisukeMiyamoto:
Thanks for the review — nit adopted in 6f96ef34:

- Dropped the `tag:` anchor from the banned pattern so both forms of the typo (CLI filter with `tag:` prefix and the bare console-hint form) are caught by lint.
- Re-verified: injecting the console-hint form now produces the expected FAIL; the CLI-filter form still fails as before; a clean tree passes.

DaisukeMiyamoto:
Superseded by a follow-up that (a) makes docs derive LOGIN_INSTANCE_ID from the CFN stack via the PCS API + `aws:pcs:compute-node-group-id` (tag-naming independent), and (b) makes the Name tag cluster-scoped (`${ClusterName}-${CngName}`) for EC2 console readability. Closing this in favour of that PR.

PR #1177, opened 2026-07-10
fix(aws-pcs): work around AWS CLI cross-region S3 redirect failure on multi-NIC instances

DaisukeMiyamoto:
## What

The UserData snippets that fetch `install-enroot-pyxis.sh` (post-install)
and `setup-directory.sh` (multi-user) call `aws s3 cp` without an explicit
`--region`, expecting the CLI's normal behaviour: when the bucket is in
a different region than the caller, the S3 HeadObject probe returns 301
with an `x-amz-bucket-region` hint, the CLI follows it, and the copy
succeeds. This works as intended on login nodes and on ordinary compute.

On **multi-NIC EFA instances (P5 / P5e / P5en / P6-B200 / P6-B300 —
8–32 default routes)** the CLI does not recover from that 301: the CRT
S3 client aborts with `AWS_ERROR_S3_INVALID_RESPONSE_STATUS` (error 14343)
before the redirect is applied.

```
[S3MetaRequest] error 14343 (aws-c-s3 AWS_ERROR_S3_INVALID_RESPONSE_STATUS)
[S3MetaRequest] Meta request cannot recover from error 14343
download failed: … An error occurred (301) when calling the HeadObject operation
bash: /tmp/pcs-post-install.sh: No such file or directory
Post-install script completed (exit 127)
```

The compute node ends up with no Pyxis, and every `srun --container-image=…`
fails — the failure is compute-only and easy to miss until the first
container job. The common trigger is the stock public bucket
(`awsome-distributed-ai`, us-east-1) consumed from a non-us-east-1
cluster; anyone deploying P5/P6 from Tokyo hits it immediately.

The upstream CLI bug is being tracked separately; this PR is a defensive
workaround at the deploy layer so PCS clusters don't have to wait for
that fix.

## Fix

Resolve the bucket's real region via an anonymous HTTPS HEAD
(`x-amz-bucket-region` response header — no IAM/credentials needed) and
pass it explicitly as `--region` to `aws s3 cp`. If the HEAD probe
returns nothing (private / non-standard endpoint), fall back to the
original `aws s3 cp` invocation so **same-region deployments behave
exactly as before** and no working setup regresses.

Applied in lock-step to the four hand-maintained CNG UserData copies
(`add-cng.yaml` + `add-cng-p5.yaml` + `add-cng-p6-b200.yaml` +
`add-cng-p6-b300.yaml`), on both the post-install fetch and the
setup-directory fetch, and to `scripts/setup-directory.sh`'s
`ldap-add-user.sh` fetch on the login node.

## Testing

- **Symptom reproduced** on a fresh Tokyo cluster with
  `DirectoryService=OpenLDAP-LoginNode` and `p5.48xlarge` on a Capacity
  Block: the P5 compute produced the 301 / exit 127 above; login was fine.
- **Fix verified live** on a new Tokyo cluster with the same P5 CB using
  the fixed templates. Both login and P5 compute now log:

  ```
  Downloading post-install script from s3://…/install-enroot-pyxis.sh...
  Resolved bucket region: us-east-1
  Completed 12.7 KiB/12.7 KiB … download: … to tmp/pcs-post-install.sh
  Executing post-install script...
  ```

  Post-install runs to `exit 0`; `spank_pyxis.so` installed; monitoring
  stack up (6 containers) on login.
- The region-resolve block runs POSIX-clean under `dash` (cloud-init
  `runcmd` shell): real public bucket → `us-east-1` resolved; nonexistent
  bucket → empty, falls through to the original call.
- `aws cloudformation validate-template` passes on all four CNG templates;
  `tests/lint-docs.sh` passes.

### DIFF
--- a/architectures/aws-pcs/assets/scripts/setup-directory.sh
+++ b/architectures/aws-pcs/assets/scripts/setup-directory.sh
@@ -210,8 +210,17 @@ EOF
     # location this script came from (S3_BUCKET/S3_KEY_PREFIX passed by UserData);
     # falls back to a no-op with a hint if those are unset.
     if [ -n "${S3_BUCKET:-}" ]; then
+        # Resolve the bucket's region via HTTPS HEAD and pass it to --region.
+        # Without this, cross-region `aws s3 cp` gets a 301 on the HeadObject
+        # probe that CRT S3 client cannot recover from — most visibly on
+        # multi-NIC instances (P5/P6). Silent fallback keeps single-region
+        # deployments working exactly as before.
+        BUCKET_REGION=$(curl -sI "https://${S3_BUCKET}.s3.amazonaws.com/" 2>/dev/null \
+          | tr -d '\r' | awk -F': ' 'tolower($1)=="x-amz-bucket-region"{print tolower($2)}')
+        REGION_ARG=""
+        [ -n "${BUCKET_REGION}" ] && REGION_ARG="--region ${BUCKET_REGION}"
         if aws s3 cp "s3://${S3_BUCKET}/${S3_KEY_PREFIX:-templates/aws-pcs/}scripts/ldap-add-user.sh" \
-             /usr/local/bin/ldap-add-user.sh 2>/dev/null; then
+             /usr/local/bin/ldap-add-user.sh ${REGION_ARG} 2>/dev/null; then
             chmod 755 /usr/local/bin/ldap-add-user.sh
             echo "[directory-server] Installed helper: /usr/local/bin/ldap-add-user.sh"
         else
--- a/architectures/aws-pcs/assets/add-cng-p5.yaml
+++ b/architectures/aws-pcs/assets/add-cng-p5.yaml
@@ -418,7 +418,27 @@ Resources:
               if [ -n "${PostInstallScriptUrl}" ]; then
                 echo "Downloading post-install script from ${PostInstallScriptUrl}..." | tee /var/log/pcs-post-install.log
                 case "${PostInstallScriptUrl}" in
-                  s3://*) aws s3 cp "${PostInstallScriptUrl}" /tmp/pcs-post-install.sh >> /var/log/pcs-post-install.log 2>&1 ;;
+                  s3://*)
+                    # Resolve the bucket's real region via HTTPS HEAD and pass it as
+                    # --region. Without this, `aws s3 cp` uses the instance's IMDS
+                    # region for signing; if that differs from the bucket's region,
+                    # the HeadObject probe gets 301 Moved Permanently and the CLI's
+                    # CRT S3 client fails to recover (aws-c-s3 AWS_ERROR_S3_INVALID_
+                    # RESPONSE_STATUS 14343). Symptom: post-install exit 127 on GPU
+                    # CNGs when the templates bucket sits in a different region
+                    # than the cluster.
+                    S3_URL="${PostInstallScriptUrl}"
+                    S3_BUCKET_FOR_HEAD=$(printf '%s' "$S3_URL" | sed -E 's#^s3://([^/]+).*#\1#')
+                    S3_BUCKET_REGION=$(curl -sI "https://$S3_BUCKET_FOR_HEAD.s3.amazonaws.com/" \
+                      | tr -d '\r' | awk -F': ' 'tolower($1)=="x-amz-bucket-region"{print tolower($2)}')
+                    if [ -n "$S3_BUCKET_REGION" ]; then
+                      echo "Resolved bucket region: $S3_BUCKET_REGION" | tee -a /var/log/pcs-post-install.log
+                      aws s3 cp "$S3_URL" /tmp/pcs-post-install.sh --region "$S3_BUCKET_REGION" >> /var/log/pcs-post-install.log 2>&1
+                    else
+                      echo "Bucket region lookup failed; falling back to CLI default" | tee -a /var/log/pcs-post-install.log
+                      aws s3 cp "$S3_URL" /tmp/pcs-post-install.sh >> /var/log/pcs-post-install.log 2>&1
+                    fi
+                    ;;
                   *)      curl -fsSL "${PostInstallScriptUrl}" -o /tmp/pcs-post-install.sh ;;
                 esac
                 chmod +x /tmp/pcs-post-install.sh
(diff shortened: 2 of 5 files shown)


KeitaW (approved):
Merging now, am reviewing the PR on actual environment in parallel. Follow up PR will be created if any issue discovered.

architectures/aws-pcs/assets/scripts/setup-directory.sh
```diff
@@ -210,8 +210,17 @@ EOF
     # location this script came from (S3_BUCKET/S3_KEY_PREFIX passed by UserData);
     # falls back to a no-op with a hint if those are unset.
     if [ -n "${S3_BUCKET:-}" ]; then
+        # Resolve the bucket's region via HTTPS HEAD and pass it to --region.
+        # Without this, cross-region `aws s3 cp` gets a 301 on the HeadObject
+        # probe that CRT S3 client cannot recover from — most visibly on
+        # multi-NIC instances (P5/P6). Silent fallback keeps single-region
+        # deployments working exactly as before.
+        BUCKET_REGION=$(curl -sI "https://${S3_BUCKET}.s3.amazonaws.com/" 2>/dev/null \
```
KeitaW, line 218: ### The region-probe `curl` is unbounded — a probe added *for* multi-NIC boot paths should cap its own time — 🟢
**Observation:** The `curl -sI "https://${bucket}.s3.amazonaws.com/"` region probe (this line, and the same probe in the 4 `add-cng*.yaml` UserData copies) has no `--connect-timeout` / `--max-time`. It runs early in first boot on the very multi-NIC instances the fix targets — where the box has 8–32 default routes and only the primary NIC reaches the internet. If curl ever binds a route without egress, an unbounded probe stalls the whole post-install (and thus Pyxis install) until curl's default connect timeout (~2 min/attempt).
**Impact:** Low in practice — on the real P5 I tested, the probe returned in **0.48s** (curl picked a working route), so it does not currently hang. But an unbounded network call on a multi-NIC boot path is exactly the class of latent hang this PR is otherwise hardening against.
**Suggestion:** Bound the probe so a bad route fails fast into the existing fallback:
```suggestion
        BUCKET_REGION=$(curl -sI --connect-timeout 5 --max-time 10 "https://${S3_BUCKET}.s3.amazonaws.com/" 2>/dev/null \
```
(Apply the same `--connect-timeout 5 --max-time 10` to the four `add-cng*.yaml` probe lines.)

KeitaW (commented):
## Review Batch 1/2 — Probe Robustness (follow-up PR)

> Note: this PR is already merged (to unblock users), so the items below are small defensive improvements for a **separate follow-up PR**, not blockers. The fix as shipped works — I confirmed it end-to-end on the exact failing hardware (see Positives).

### Bucket names containing dots silently fall back (TLS wildcard) — 🟢 worth a one-line comment
**Observation:** The probe uses virtual-hosted style (`https://<bucket>.s3.amazonaws.com/`). For a bucket whose name contains dots (`my.dotted.bucket`), the TLS SNI/cert `*.s3.amazonaws.com` wildcard only matches one label, so curl's cert validation fails, `x-amz-bucket-region` comes back empty, and the code falls back to the original no-`--region` `aws s3 cp` — which is the buggy path for a cross-region dotted bucket.
**Impact:** Very low — dotted bucket names are discouraged and the stock bucket isn't dotted, and the fallback is safe (same behavior as before the PR). But it's a silent degradation worth a comment so a future reader knows dotted buckets aren't covered.
**Suggestion:** Either note it in the existing comment, or use path-style for the probe (`https://s3.<guess>.amazonaws.com/<bucket>` won't help without a region; simplest is to document the limitation). Not worth code churn on its own — fold into the timeout follow-up.

KeitaW (commented):
## Review Batch 2/2 — Maintainability, Positives & Sources



### The 4-way UserData duplication keeps forcing lock-step edits — 🟢 (observation)
**Observation:** This fix had to be applied identically to four hand-maintained CNG UserData copies (`add-cng.yaml` + `add-cng-p5|p6-b200|p6-b300.yaml`). To the PR's credit, all four added blocks are **byte-identical** (I diffed them: same 30-line block, same hash), so there's no drift *this* time. But this is now another load-bearing edit that had to land in four places — the same duplication that has bitten this file set before.
**Impact:** None today; it's a standing fragility. The next edit that lands in three of four copies is where a real divergence bug appears.
**Suggestion:** Not for this PR — but a good candidate to factor the shared UserData into one source (a nested snippet or a fetched-and-sourced fragment), or add a CI assertion that the block is byte-identical across the four templates.

---

### Things That Look Great
- **The root-cause is correct and I reproduced it live.** On a real p5.48xlarge in ap-northeast-1 (Tokyo), `aws s3 cp` of the us-east-1 stock script **without** `--region` fails with `An error occurred (301) when calling the HeadObject operation: Moved Permanently`. The inline comment's mechanism is accurate: without `--region` the request signs for the instance's IMDS region, S3 answers the cross-region HeadObject with `301 + x-amz-bucket-region`, and the CRT S3 client aborts (`AWS_ERROR_S3_INVALID_RESPONSE_STATUS`/14343) instead of following it. One precision worth a line in the description: the true differentiator is that **AWS CLI v2 auto-enables the CRT client on P4d/P5/P6/Trn-class instances** ([AWS docs](https://docs.aws.amazon.com/cli/latest/topic/s3-config.html)) — the classic client (e.g. on the m6i login node in the same cluster/region) follows the 301 fine. "Multi-NIC" is a correlate of those instance types, not the mechanism; the fix is right either way because `--region` removes the 301 entirely.
- **The fix works end-to-end on that hardware.** The merged post-install log on the P5 shows `Resolved bucket region: us-east-1` → download success → `Post-install script completed (exit 0)`, and Enroot/Pyxis installed. The raw `aws s3 cp … --region us-east-1` returns exit 0.
- **The probe's core assumption holds where it matters.** The key risk was whether `curl` also breaks on the 8–32-default-route condition that defeats the CLI's CRT client — it doesn't: the `x-amz-bucket-region` probe resolved `us-east-1` in **0.48s** on the 32-NIC P5. Choosing an anonymous `curl` HEAD over the CLI for region discovery is the right call.
- **The fallback is genuinely conservative.** When the probe returns nothing, it runs the original `aws s3 cp` unchanged, so same-region deployments behave exactly as before — no regression surface. The comments explaining *why* (the 14343 / 301 mechanism) are excellent and will save the next reader a long debugging session.
- **Applied consistently** to all four CNG copies plus `setup-directory.sh`, byte-identical.

---

### Sources
- e2e (verified live 2026-07-10, acct 159553542841): PCS deploy-all in ap-northeast-1 on capacity block `cr-06685e167af91ddca`, P5 node `i-084678dc80a75aa21` (32 default routes); bug repro `301 Moved Permanently` (no `--region`), probe `us-east-1` in 0.48s, fix `--region` exit 0, post-install exit 0. Stack torn down after validation.
- aws-cli on the PCS-Ready DLAMI: `aws-cli/2.35.16` (the v2/CRT client that exhibits the redirect failure).
- The `x-amz-bucket-region` response header is the documented way to discover a bucket's region without credentials (S3 returns it on a HEAD even for a 301/403).

PR #1180, opened 2026-07-11
Fix unquoted OS_VERSION in FSDP create_venv.sh

Zhenye-Na:
## Purpose

Relates to a shell-robustness issue in the FSDP Slurm setup script. The
`create_venv.sh` script does an unquoted string comparison on `$OS_VERSION`,
which breaks when the variable is empty.

## Changes

- Quote `$OS_VERSION` in the `[ ]` string test in
  `3.test_cases/pytorch/FSDP/slurm/create_venv.sh`.

`OS_VERSION` is parsed from `/etc/os-release`. If that file is missing or the
`awk` parse yields an empty string, `if [ $OS_VERSION = "20.04" ]` expands to
`if [ = "20.04" ]`, which is a syntax error (`unary operator expected` /
`too many arguments`). Because the script runs under `set -ex`, this aborts
venv creation. Quoting the variable (`[ "$OS_VERSION" = "20.04" ]`) makes the
comparison evaluate to a well-defined false instead of erroring out.
Addresses ShellCheck SC2086.

## Test Plan

**Environment:**
- AWS Service: N/A (shell syntax fix, filesystem-independent)
- Instance type: N/A
- Number of nodes: N/A

**Test commands:**
```bash
# Syntax check
bash -n 3.test_cases/pytorch/FSDP/slurm/create_venv.sh

# Behavior with an empty OS_VERSION (reproduces the original failure)
OS_VERSION=""
[ $OS_VERSION = "20.04" ]   && echo ok   # before: syntax error
[ "$OS_VERSION" = "20.04" ] && echo ok || echo "false, no error"  # after
```

## Test Results

- `bash -n` passes with no syntax errors.
- With `OS_VERSION=""`, the unquoted form raises
  `bash: [: =: unary operator expected`; the quoted form cleanly evaluates to
  false and the script proceeds to the `else` branch.

## Directory Structure

N/A — one-line fix to an existing script, no test case added or restructured.

## Checklist

- [x] External dependencies are pinned to a specific version or tag (no `latest`). — N/A, no dependency changes
- [x] A README is included or updated with prerequisites, instructions, and known issues. — N/A, no behavior/usage change
- [x] New test cases follow the [expected directory structure](#directory-structure). — N/A, no new test case

### DIFF
--- a/3.test_cases/pytorch/FSDP/slurm/create_venv.sh
+++ b/3.test_cases/pytorch/FSDP/slurm/create_venv.sh
@@ -8,7 +8,7 @@ PYTHON_V=python3
 OS_VERSION=$(cat /etc/os-release | grep VERSION_ID | awk -F '=' '{print $2}')
 OS_VERSION=${OS_VERSION//\"/}
 
-if [ $OS_VERSION = "20.04" ]; then
+if [ "$OS_VERSION" = "20.04" ]; then
 
    PYTHON_VERSION=$(python3.9 --version | awk '{print $2}' | awk -F'.' '{print $1"."$2}')
    PYTHON_V=python3.9


KeitaW (commented):
## Review Batch 1/2 — Shell Robustness

### The quoting fix is correct — but the description overstates the failure mode it prevents

Quoting `"$OS_VERSION"` in the `[ ]` test is the right hardening — it's exactly the SC2086 fix and protects against an empty value, embedded whitespace, and glob expansion in the operand. The change itself is correct and complete: `create_venv.sh` contains only this one `[ ]` test, so nothing of this class is left behind.

One small thing on the rationale, so it isn't quoted as fact in the commit history later: the description says the unquoted form is a *syntax error that aborts venv creation under `set -ex`*. It doesn't actually abort. A command that fails while being used as an `if` condition is exempt from `set -e`, so the original emitted a spurious `[: =: unary operator expected` to stderr and then fell through to the (correct) `else` branch — the script continued and exited 0. So the real benefit here is silencing the noisy diagnostic and being robust to a value with whitespace/globs, not preventing an abort. The fix is still worth landing; only the "aborts" framing is off.

Verified locally 2026-07-12: `set -ex; OS_VERSION=""; if [ $OS_VERSION = "20.04" ]; ...` prints the `unary operator expected` diagnostic on stderr but reaches the end of the script with exit 0.

## Review Batch 2/2 — Positives & Sources

### Things That Look Great
- Minimal, surgical one-line change that traces exactly to the stated purpose — no drive-by edits.
- The quoted form is genuinely more robust than the description claims: it also handles an `OS_VERSION` that parses to a value containing spaces or glob characters, both of which would misbehave unquoted independent of the `set -e` question.
- MIT-0 header and shebang already present and untouched; trailing newline preserved.

### Sources
- Bash `set -e` semantics — a command failing as part of an `if`/`while`/`until` condition (or on the left of `&&`/`||`) does not trigger exit (GNU Bash manual, "The Set Builtin", `-e`): https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
- ShellCheck SC2086 (quote to prevent word splitting/globbing): https://www.shellcheck.net/wiki/SC2086

KeitaW (approved):

PR #1189, opened 2026-07-14
feat: kimi-k2 DeepEP v1 + NVSHMEM-over-EFA arm (32-node results) + ep-backend scale limits

KeitaW:
## Summary

- Adds the **DeepEP v1 + NVSHMEM-over-EFA dispatcher arm** to the `kimi-k2` Megatron-Bridge test case (third arm, mirroring the `qwen3-235b` 3-way pattern) with **measured 32-node results** on literal Kimi-K2 (256× B300, TP8/PP8/EP32) — the parent README's "not yet validated at 16–32 nodes" follow-up is now closed.
- Extends `micro-benchmarks/expert-parallelism/ep-backend-comparison` with a **16/32-node (128/256-rank) scale run**: new NCCL reference numbers plus source-cited documentation of the hard scale limits every DeepEP-class kernel hits past 64–160 ranks.
- Launcher (`run-ab-rawpods.sh`) gains two env-gated, default-off capabilities needed to run on FSx-less local-zone capacity blocks (`STORAGE=hostpath`, `GDRCOPY_DEV=on`) and two bug fixes.

### kimi-k2: DeepEP+NVSHMEM arm (`3.test_cases/megatron/megatron-bridge/`)

Measured 2026-07-14 on a 32× `p6-b300.48xlarge` Capacity Block (us-east-1-atl-2a local zone), campaign `20260714T0515Z-k2-nvshmem-pp8-32n-ue1`, image `megatron-bridge-uccl:nemo-26.04.01-deepep-nvshmem-567632d-cu13` (NVSHMEM v3.7.0-0, DeepEP `567632d`, the [`deepep-benchmark`](https://github.com/awslabs/awsome-distributed-ai/tree/main/micro-benchmarks/expert-parallelism/deepep-benchmark) NVSHMEM-over-libfabric/EFA build):

| cell | DeepEP+NVSHMEM (2026-07-14) | UCCL deepep (2026-06-04) | NCCL alltoall (2026-06-04) |
|---|---|---|---|
| mb=4, overlap=on | **3.742 s/iter · 224.6 TF/GPU · 280.2k tok/s** | 3.834 · 219.1 · 273.5k | 5.793 · 144.8 · 181.0k |
| mb=4, overlap=off | 6.632 s/iter · 126.4 TF/GPU | **6.041 · 138.8** | 9.303 · 90.1 |
| mb=1, overlap=off | 13.538 s/iter · 61.9 TF/GPU | 14.98 · 56.0 | **12.88 · 65.1** |

Cross-campaign comparison (different cluster/date/storage than the June UCCL/NCCL A/B) — caveat documented in RESULTS.md. Validity per cell: `Selected provider is efa` on all 32 nodes, 0 stalls, `n_steady` 20/20, `parse-runs.py` transport = nvshmem, iteration-1 loss matches the June arms to 5 significant figures.

- `kimi-k2/README.md`: 3-arm table, NVSHMEM run instructions, `GDRCOPY_DEV` section, "running without FSx (`STORAGE=hostpath`)" section (per-node staging incl. the offline `deepseek-ai/DeepSeek-V3` HF cache the DSV3 recipe scaffolding needs, per-cell harvest loop)
- `kimi-k2/benchmarks/RESULTS.md`: new NVSHMEM section + fixed reproduce block (`ARM` is positional, the old example passed it as env and would not run)
- `README.md` (parent): recommendation bullet updated with the measured 32-node outcome; kimi-k2 models-table row now 3-way

### Launcher (`run-ab-rawpods.sh`) — env-gated, defaults unchanged

- `STORAGE=pvc|hostpath` (+ `FSX_PVC`, `HOSTPATH_ROOT`): the `/fsx` volume source was hard-coded to the `fsx-kimi-k2` PVC; `hostpath` backs it with node-local NVMe for clusters without FSx
- `GDRCOPY_DEV=on`: hostPath-mounts `/dev/gdrdrv` (privileged). NVSHMEM grows its symmetric heap dynamically (CUDA-VMM) once buffers outgrow the 1 GiB init chunk, and registering grown chunks over libfabric/EFA requires GDRCopy **in-container**; on clusters whose nvidia container toolkit ignores `NVIDIA_GDRCOPY=enabled`, every rank dies at `mem_heap.cpp:1361 register_mem_handle failed` without this
- Fix: `set -u` crash on direct invocation (`HF_HOME` default referenced `STAGE` before it was set)
- Fix: `HF_HOME`/`HF_HUB_OFFLINE` were computed but never threaded into the pod env (dead variables)

### ep-backend-comparison: 16/32-node scale run (`micro-benchmarks/expert-parallelism/`)

- `RESULTS.md`: new "32 / 16 nodes — 256 / 128 ranks" section. NCCL alltoall matched-size busbw 73.8 GB/s (128 ranks) / 54.7 GB/s (256 ranks); NVSHMEM HT dispatch tunes to 74.7 GB/s RDMA at 16 nodes. Hard limits (each reproduced ≥2×, source-cited):
  - DeepEP HT: 160-rank cap (`NUM_MAX_NVL_PEERS 8 × NUM_MAX_RDMA_PEERS 20`, `kernels/configs.cuh`); per-node-count template instantiation table; stock combine tuning aborts at 16 RDMA peers (`internode.cu:2363`)
  - UCCL HT: `int32` `num_rdma_bytes` overflow above 64 ranks (`uccl_ep.cc:431`)
  - UCCL LL: `kAtomicBufferSize` signaling-arena overflow at ≥128 ranks (`ep_config.hpp:279`)
  - DeepEP LL over NVSHMEM-libfabric: host-proxy `FI_EAGAIN` retry exhaustion at ≥128 PEs with different victim ranks each run (statistical fan-out limit, not a bad node — verified across 4 runs)
- `README.md`: "Scaling beyond 8 nodes" section (operational guidance incl. the gdrdrv workaround); `env_vars.example` annotated
- Toleration fix: `nccl-alltoall.yaml` and the three `uccl-ep-benchmark` manifests were missing the `workload=bench` toleration the deepep manifests already carry — pods stay `Pending` on tainted B300 pools without it (launcher + worker both patched)

## Test plan

- All three kimi-k2 NVSHMEM cells run end-to-end on real hardware (32× p6-b300) and pass every `parse-runs.py` gate; numbers above are from `index.csv`
- All ep-backend numbers from launcher logs collated with `collect_results.py` and cross-checked against raw logs
- Each documented limit reproduced at least twice; DeepEP/UCCL assert lines quoted verbatim from the shipped sources
- `bash -n run-ab-rawpods.sh` clean; default-path pod rendering unchanged (PVC volume, no privileged, no gdrdrv mount)
- All manifests re-validated as YAML after the toleration patch; relative links in all modified markdown verified to resolve

### DIFF
--- a/3.test_cases/megatron/megatron-bridge/README.md
+++ b/3.test_cases/megatron/megatron-bridge/README.md
@@ -117,7 +117,7 @@ identical directory layout.
 
 | Model | Directory | Workloads |
 |-------|-----------|--------|
-| [Kimi K2](https://huggingface.co/moonshotai/Kimi-K2-Base) (1.04T MoE, 384 experts) | [`kimi-k2/`](kimi-k2/) | Full-parameter SFT + UCCL-EP vs NCCL dispatcher A/B (32× p6-b300) |
+| [Kimi K2](https://huggingface.co/moonshotai/Kimi-K2-Base) (1.04T MoE, 384 experts) | [`kimi-k2/`](kimi-k2/) | Full-parameter SFT + 3-way dispatcher comparison: NCCL vs DeepEP+UCCL (same-campaign A/B) vs DeepEP+NVSHMEM (32× p6-b300) |
 | [DeepSeek-V3](https://github.com/deepseek-ai/DeepSeek-V3) (671B MoE, 256 experts) | [`dsv3/`](dsv3/) | Full-parameter SFT + UCCL-EP vs NCCL dispatcher A/B (32× p6-b300) |
 | [Qwen3-235B-A22B](https://huggingface.co/Qwen/Qwen3-235B-A22B) (235B MoE, 128 experts) | [`qwen3-235b/`](qwen3-235b/) | **3-way** dispatcher comparison: NCCL vs DeepEP+UCCL vs DeepEP+NVSHMEM, EP16/EP32 (8× p6-b300 / B300) |
 | [Qwen3-30B-A3B](https://huggingface.co/Qwen/Qwen3-30B-A3B) (30B MoE, 128 experts) | [`qwen3-30b/`](qwen3-30b/) | Same **3-way** comparison on **8× p5.48xlarge / H100** (the size that fits H100-80GB), EP16/EP32 |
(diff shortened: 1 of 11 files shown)


whn09:
Nice work — the cross-campaign caveat is handled honestly and the source-cited scale limits (verified the 8×20=160 HT peer cap) are a great touch. One ask: the NVSHMEM win is really only at mb=4+overlap=on (and within cross-campaign noise), while at overlap=off it trails UCCL by ~10% — could we soften the parent README's "slightly ahead of UCCL" to "on par / competitive" so the ~2.4% isn't read as a ranking?

KeitaW:
Thanks @whn09 — agreed, "slightly ahead" invited a ranking the data doesn't support. Fixed in ae364b15:

- Parent README now reads "**on par with the June UCCL number** (within cross-campaign noise; at mb=4 overlap=off it trails UCCL by ~10% — read the two DeepEP transports as competitive, not ranked)".
- The RESULTS.md takeaway now states both cells explicitly (~2.4% under UCCL at mb=4+overlap, ~10% over at mb=4 no-overlap) and says "no ranking to read here".

Worth adding for context: the HT scaling sweep pushed since your comment (b94efa0b) includes an 8-node re-measure on this cluster that reproduces the June us-west-2 numbers within ~1% — so the cross-campaign noise floor for these benchmarks is small, but the mb4-overlap delta (2.4%) still sits inside it.

PR #1210, opened 2026-07-22
Slurm HMA CloudWatch bridge + widen EventBridge rule

mayankgupta14:
### What

- Adds a **Slurm-only CloudWatch Logs subscription filter → tiny Lambda → `events:PutEvents`** path that synthesizes a `SageMaker HyperPod Cluster Event Warn` from HMA detection records. Downstream pipeline (bridge Lambda → DevOps Agent → RCA → email) unchanged.
- **Widens** the `HyperPodEventRule` EventBridge pattern to catch the Slurm-drift and Deep-Health-Check event families that were previously silently dropped (hybrid `prefix: "SageMaker HyperPod"` + explicit list).
- Adds **dedicated bridge Lambda handlers** for `DeepHealthCheckCompleted`/`Failed`, `SlurmConfigurationDriftDetected`, `SlurmReconfigurationCompleted`, `SlurmTopologyUpdated`, `MungeKeyCreated`. Improves the **generic-fallback branch** to inline compact JSON of `detail` for any unrecognized detail-type.
- Documentation: `README.md` (new HMA-bridge Slurm-only section), `IMPLEMENTATION.md` (short design rationale + design-boundary note on Monitor re-checks), `IDEAS.md` (marks the CloudWatch-subscription-filter idea as shipped).

### Why

- **Silent-drop bug 1 (Slurm HMA)**: On Slurm-orchestrated HyperPod clusters, the control plane does **not** currently emit the `Warn`-level HMA-attribution `Cluster Event` on EventBridge that EKS emits natively. HMA still writes its detection JSON to CloudWatch Logs and HyperPod still initiates node replacement, but no operator-visible EventBridge event is produced, so the DevOps Agent pipeline never runs for HMA-origin faults on Slurm. Live-verified on 2026-07-20: Slurm cluster HMA detected Xid 79 within 4 s of injection, wrote to CloudWatch stream `SagemakerHealthMonitoringAgent/worker-gpu/i-…`, but no operator-visible EventBridge event fired, and no email was produced.

- **Silent-drop bug 2 (EventBridge rule too narrow)**: The current rule enumerates 3 detail-types. Newer HyperPod detail-types (`SlurmConfigurationDriftDetected`, `SlurmReconfigurationCompleted`, `SlurmTopologyUpdated`, `MungeKeyCreated`, `DeepHealthCheckCompleted`/`Failed`) don't match the rule and are silently dropped before the bridge Lambda sees them.

Both are behavior bugs in the shipped solution, not new capabilities.

## Changes

-

## Test Plan

**Environment:**
- AWS Service:
- Instance type:
- Number of nodes:

**Test commands:**
```bash

```

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/tools/devops-agent/deploy/lambda/hma_cw_bridge.py
+++ b/1.architectures/5.sagemaker-hyperpod/tools/devops-agent/deploy/lambda/hma_cw_bridge.py
@@ -1,0 +1,429 @@
+"""Bridge HMA CloudWatch Logs detection events onto EventBridge (Slurm only).
+
+Why this exists
+---------------
+On EKS-orchestrated HyperPod clusters, the control plane emits a Warn-level
+`SageMaker HyperPod Cluster Event` on EventBridge whenever HMA detects a GPU
+fault (NvidiaGPUUnhealthy, ECC DBE, etc.). That event flows through the
+webhook bridge Lambda into DevOps Agent and produces an investigation email.
+
+On Slurm-orchestrated HyperPod clusters today, the HyperPod control plane
+does NOT emit that Warn-level event for HMA-attributed faults. HMA still
+writes its detection JSON to the cluster's CloudWatch log group
+(SagemakerHealthMonitoringAgent/<group>/<instance>), and HyperPod still
+initiates node replacement, but no operator-visible EventBridge event is
+produced -- the DevOps Agent pipeline never runs for HMA-origin faults on
+Slurm.
+
+This Lambda closes that gap. A CloudWatch Logs subscription filter on the
+cluster's log group (with FilterPattern
+`{ $.HealthMonitoringAgentDetectionEvent = "HealthEvent" }`) invokes this
+Lambda for every HMA detection record. The Lambda extracts fault type +
+repair action + instance ID from the HMA record and calls `events:PutEvents`
+with a synthetic `SageMaker HyperPod Cluster Event` shaped identically to
+what HyperPod emits on EKS -- so the existing webhook bridge picks it up
+unchanged, no downstream code changes required.
+
+This is a Slurm-only workaround. Enabled by the CFN parameter
+`EnableHmaCloudWatchBridge`, which `deploy.sh` auto-resolves to `true` for
+Slurm and `false` for EKS. On EKS the subscription filter and this Lambda
+are simply not deployed -- HyperPod's native event still flows.
+
+Environment variables
+---------------------
+  CLUSTER_NAME     HyperPod cluster name (goes into the synthetic event).
+  EVENT_BUS_NAME   EventBridge bus to publish onto. Default: "default".
+  LOG_LEVEL        Structured-log level (INFO or DEBUG). Default: INFO.
+
+CloudWatch Logs invocation contract
+-----------------------------------
+CloudWatch delivers a gzipped, base64-encoded payload:
+
+  {
+    "awslogs": {
+      "data": "<base64-encoded gzip of the actual JSON>"
+    }
+  }
+
+The decoded JSON has this shape:
+
+  {
+    "messageType": "DATA_MESSAGE",
+    "owner": "<account-id>",
+    "logGroup": "/aws/sagemaker/Clusters/<name>/<id>",
+    "logStream": "SagemakerHealthMonitoringAgent/<group>/<instance>",
+    "subscriptionFilters": [...],
+    "logEvents": [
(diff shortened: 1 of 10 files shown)


1.architectures/5.sagemaker-hyperpod/tools/devops-agent/deploy/lambda/webhook_bridge.py
```diff
@@ -345,10 +345,24 @@ def _pick_failure_message() -> str:
             description += f" EventMetadata: {extras_str}."
         return title, description
 
+    # -------------------- Generic fallback (improved) --------------------
```
shimomut: This generic fallback handles future unknown event formats, but it performs no EventLevel filtering. It forwards every matched event unconditionally. Because the prefix rule (SageMaker HyperPod *) now routes all current and future detail-types here, and each forwarded event triggers a DevOps Agent triage and may be RCA, a new high-frequency event type from HyperPod could fan out to one paid investigation per emission. This is a cost concern on the DevOps Agent side and I'd suggest gating the catch-all by severity before merge.

"SageMaker HyperPod Cluster Node Health Event" and "SageMaker HyperPod Cluster State Change" are also not checking the EventLevel (actually there is no EventLevel for those events). But with Continuous Provisioning, these two event types don't come actually. So handling for "SageMaker HyperPod Cluster Node Health Event" and "SageMaker HyperPod Cluster State Change" is a dead code we can remove (out of scope of this PR).

shimomut (changes requested):
There is a cost concern in the generic fallback because unknown new event detail type may cause a lot of Triage and RTA. Not an immediate concern, but essentially we are changing our stance of "The Lambda POSTs a webhook event to DevOps Agent only when a real issue is found. On a healthy cluster, nothing is POSTed".

1.architectures/5.sagemaker-hyperpod/tools/devops-agent/deploy/lambda/hma_cw_bridge.py
```diff
@@ -0,0 +1,429 @@
...
+        f"HyperPod Health Monitoring Agent (HMA) has detected fault type "
+        f"{fault_display} on this node and is unhealthy. "
+        f"Repair action: {action_display}."
+    )
+    if original_message and original_message not in description:
+        description += f" HMA detail: {original_message}"
+
+    detail = {
+        "EventDetails": {
+            "EventId": "synthetic-hma-cw-bridge",  # not a real cluster-event ID
```
shimomut: This static event ID is causing an error in webhook_bridge Lambda.

```
describe_cluster_event failed cluster='slurm-2' eventId='synthetic-hma-cw-bridge' err=ClientError("An error occurred (ValidationException) when calling the DescribeClusterEvent operation: 1 validation error detected: Value 'synthetic-hma-cw-bridge' at 'eventId' failed to satisfy constraint: Member must satisfy regular expression pattern: [a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}")
```

It looks the webhook lambda is working fine even with this error, so this is not a blocking feedback, but we could easily fix this issue by setting empty string for EventID.

shimomut (commented):
Non blocking comment.

1.architectures/5.sagemaker-hyperpod/tools/devops-agent/deploy/lambda/webhook_bridge.py
mayankgupta14: Agree. Pushing a fix that adds a severity gate to the generic-fallback path: any detail-type that isn't one of the three explicit SageMaker HyperPod Cluster {State Change, Node Health Event, Event} types is dropped unless EventLevel is Warn, Error, or Critical. The three known detail-types keep their exact current behavior, no regression. Preserves the future-proofing goal of the prefix rule (a new high-severity detail-type still catches) while restoring the "POST only on real issues" cost model.

On the out-of-scope note about Cluster State Change / Cluster Node Health Event handlers being dead code with Continuous Provisioning - makes sense, will file a follow-up to clean those up, keeping it out of this PR as suggested.

1.architectures/5.sagemaker-hyperpod/tools/devops-agent/deploy/lambda/hma_cw_bridge.py
mayankgupta14: Thanks for catching this, I missed it. It's a small fix, will push the fix shortly.

shimomut (approved):
LGTM!

Issue #427, opened 2024-09-12
HyperPod Default Enroot Path is Root Volume


nghtm:
The HyperPod default enroot path uses `/opt/sagemaker` due to the first if statement defined [here](https://github.com/aws-samples/awsome-distributed-training/blob/cb28283a0cf11688173fab128e141c4f5e89b3b2/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_enroot_pyxis.sh#L64). This is ussually approx 500 GB of root volume, depending on user configuration. 

For larger models, including Nemotron 340b, a larger volume is required, to avoid running out of enroot space as seen in error log below:

```
slurmstepd: error: pyxis: child 1528235 failed with error code: 1
slurmstepd: error: pyxis: failed to create container filesystem
slurmstepd: error: pyxis: printing enroot log file:
slurmstepd: error: pyxis:     [INFO] Extracting squashfs filesystem...
slurmstepd: error: pyxis:     Write on output file failed because No space left on device
slurmstepd: error: pyxis:     FATAL ERROR:writer: failed to write file /opt/sagemaker/tmp/enroot/data/user-1000/pyxis_167.0/usr/local/tensorrt/targets/x86_64-linux-gnu/lib/libnvinfer_static.a
slurmstepd: error: pyxis:     Parallel unsquashfs: Using 96 processors
slurmstepd: error: pyxis:     433207 inodes (820890 blocks) to write
```

This can be fixed by changing the order of the if/elif statement to default to`/opt/dlami/nvme` (28TB on p5s) instead, which will make enroot use NVME instead of root volume space [here](https://github.com/aws-samples/awsome-distributed-training/blob/main/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_enroot_pyxis.sh#L81).A PR is required to modify the order if/elif in the lifecycle script

nghtm:
When run on a newly deployed HyperPod cluster (4x p5s), I see the following, which indicates that the enroot path is indeed being set to `/opt/dlami/nvme` ...

```
srun -N 4 cat /etc/enroot/enroot.conf | grep -E "ENROOT_CONFIG_PATH|ENROOT_CACHE_PATH|ENROOT_RUNTIME_PATH|ENROOT_DATA_PATH|ENROOT_TEMP_PATH"
ENROOT_RUNTIME_PATH        /opt/dlami/nvme/tmp/enroot/user-$(id -u)
[...]

PR #1218, opened 2026-07-28
fix(hyperpod-slurm): skip pyxis/enroot install + mount-aware tuning when HyperPod DLAMI already provisions them

mayankgupta14:
## Symptom

On a HyperPod Slurm cluster created with **custom lifecycle scripts from this repo** on the current HyperPod DLAMI, every `srun`/`sbatch` job launch prints 14 duplicate-option warnings:

```
$ srun
srun: spank: option "container-image" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-mounts" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-workdir" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-name" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-save" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-mount-home" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "no-container-mount-home" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-remap-root" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "no-container-remap-root" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-entrypoint" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "no-container-entrypoint" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-writable" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-readonly" provided by both spank_pyxis.so and spank_pyxis.so
srun: spank: option "container-env" provided by both spank_pyxis.so and spank_pyxis.so
```

Clusters using **None** (AMI-based lifecycle) don't hit this because `install_enroot_pyxis.sh` never runs.

## Root cause

The current HyperPod DLAMI ships both pyxis and enroot pre-installed and pre-configured:

- **pyxis:** SPANK plugin at `/opt/slurm/lib/slurm/spank_pyxis.so`, pre-registered in `plugstack.conf` with a direct `required` line.
- **enroot:** installed as an apt package.
- **enroot.conf mount-aware tuning:** the DLAMI's on-boot HostAgent tunes `ENROOT_RUNTIME_PATH`, `ENROOT_CACHE_PATH`, `ENROOT_DATA_PATH`, and `ENROOT_TEMP_PATH` based on which of NVMe / `/opt/sagemaker` / `/fsx` is mounted (verified: with FSx attached, HostAgent sets `ENROOT_CACHE_PATH=/fsx/enroot`).

Meanwhile `install_enroot_pyxis.sh`:

1. Rebuilds pyxis from source into `/usr/local/lib/slurm/spank_pyxis.so` and appends `include /opt/slurm/etc/plugstack.conf.d/pyxis.conf` to `plugstack.conf` on top of the DLAMI's `required` line - producing two SPANK registrations from two different `.so` files.
2. Re-downloads and re-installs the same enroot version via `.deb`.
3. Re-runs the mount-aware `sed` on `enroot.conf`.

With current DLAMIs (1) causes the visible bug; (2) and (3) are silent waste.

## Fix

`install_enroot_pyxis.sh` now detects each DLAMI-provided component **independently** and skips only what's redundant. Detections:

- `DLAMI_PYXIS_INSTALLED=1` if and only if `/opt/slurm/lib/slurm/spank_pyxis.so` exists **and** `plugstack.conf` contains a `required …spank_pyxis.so` line.
- `DLAMI_ENROOT_INSTALLED=1` if and only if `command -v enroot` succeeds.

## Behavior matrix

| Environment | pyxis install | enroot deb install | enroot.conf overwrite | mount-aware sed | slurm restart |
|---|---|---|---|---|---|
| **DLAMI + LCS** (this bug) | **skip** | **skip** | run | **skip** | run |
| **DLAMI, no LCS (None)** | n/a | n/a | n/a | HostAgent | HostAgent |
| **Older AMI + LCS** | run | run | run | run | run |
| **LCS re-run on any** | skip (guard idempotent) | skip (already installed) | run (safe) | skip (already enroot present) | run |

## Why LCS `enroot.conf` is still copied on top of the DLAMI's

The DLAMI ships a conservative default `enroot.conf`. LCS `enroot.conf` deliberately differs in five settings:

| Setting | LCS | DLAMI (effective) | Effect if we let DLAMI's win |
|---|---|---|---|
| `ENROOT_CONFIG_PATH` | `${HOME}/enroot` | `${HOME}/.config/enroot` (default) | Users with existing per-user credentials/mount/env files under `~/enroot/` silently stop having them read; enroot looks in `~/.config/enroot/` instead. Low impact but breakage for power users. |
| `ENROOT_SQUASH_OPTIONS` | `-comp lzo -noI -noD -noF -noX -no-duplicates` | `-noI -noD -noF -noX -no-duplicates` (no compression) | Container image size on disk ~1.5–2× larger (no LZO compression). Slightly faster `enroot import`/`enroot start`. Storage impact on FSx cache. |
| `ENROOT_ROOTFS_WRITABLE` | `yes` | `no` (default) | **HIGH.** Container rootfs becomes read-only by default. `apt install`, `pip install`, `mkdir`, writes to `/root/…`, notebook writes - all fail unless users add `--container-writable` to every `srun`. Breaking change for typical HyperPod training workflows. |
| `ENROOT_MOUNT_HOME` | `no` | `yes` (LC-inverted default) | Container's `$HOME` becomes the host's `$HOME`. Container gains ambient access to `~/.aws/credentials`, `~/.ssh/id_rsa`, etc. Files written by root-in-container end up owned by root on host FS. Security + permissions impact. |
| `ENROOT_RESTRICT_DEV` | `no` | `yes` (default) | `/dev` inside container is restricted to a minimal safe set. GPU access still works via `--nvidia`. Breaks in-container Docker, raw NVMe, some MPI stacks. Low-to-medium impact. |

Everything else in the two configs is either identical or purely commented documentation.

Two of these (`ENROOT_ROOTFS_WRITABLE`, `ENROOT_MOUNT_HOME`) are behavior-visible flags customers have very likely built training workflows around today. Silently switching them by dropping the `cp enroot.conf` step would be a breaking behavior change for anyone upgrading to a new version of LC.

For that reason this PR **preserves LC's `enroot.conf` overwrite as-is**. Aligning LC and DLAMI defaults is a separate, deliberate product decision - if it happens, it should be a distinct PR with a clear `BREAKING CHANGE:` note, an upgrade guide, and customer communication about the flags that flipped. It does not belong inside a bug fix for a duplicate SPANK registration.
## Changes

-

## Test Plan

**Environment:**
- AWS Service:
- Instance type:
- Number of nodes:

**Test commands:**
```bash

```

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_enroot_pyxis.sh
+++ b/1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_enroot_pyxis.sh
@@ -73,23 +73,25 @@ fi
 #   - enroot as an installed apt package
 # DLAMI also handles the mount-aware enroot.conf tuning for NVMe / /opt/sagemaker
 # / /fsx
-# Running `install_enroot_pyxis.sh` on top of that produces:
+#
+# Running `install_enroot_pyxis.sh` on top of a full DLAMI baseline produces:
 #   - a duplicate SPANK registration ("srun: spank: option ... provided by both
 #     spank_pyxis.so and spank_pyxis.so" on every job launch), because LCS
 #     installs a second pyxis under /usr/local/lib/slurm/ and adds an
 #     `include .../pyxis.conf` line on top of the DLAMI's direct `required` line;
-#   - redundant apt install of the same enroot version;
-#   - redundant idempotent sed of enroot.conf paths already tuned by HostAgent.
+#   - redundant apt install of the same enroot version.
 #
 # Detect each independently so partial-baseline DLAMIs (e.g. pyxis present but
 # not enroot, or vice versa) still get the piece they are missing, and so LCS
 # behavior on older Packer-built AMIs / bare Ubuntu is unchanged.
 #
 # What is NOT skipped, regardless of detection:
-#   - `cp $BIN_DIR/enroot.conf /etc/enroot/enroot.conf` (LCS copies a different,
-#     broader scoped enroot config that differs from the DLAMI's defaults in ways
-#     containers depend on: ENROOT_ROOTFS_WRITABLE=yes, ENROOT_MOUNT_HOME=no,
-#     ENROOT_RESTRICT_DEV=no, `-comp lzo` squash options, ENROOT_CONFIG_PATH).
+#   - `cp $BIN_DIR/enroot.conf /etc/enroot/enroot.conf` (LCS ships broader flags
+#     that customer container workloads depend on: ENROOT_ROOTFS_WRITABLE=yes,
+#     ENROOT_MOUNT_HOME=no, ENROOT_RESTRICT_DEV=no, `-comp lzo` squash options,
+#     ENROOT_CONFIG_PATH).
+#   - Mount-aware sed for /opt/dlami/nvme / /opt/sagemaker / /fsx (paired with
+#     the cp above - cp brings the flags, sed rescues the paths).
 #   - slurmctld / slurmd restart at the bottom of the script.
 # ------------------------------------------------------------------------------
 DLAMI_PYXIS_INSTALLED=0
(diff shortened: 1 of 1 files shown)


1.architectures/5.sagemaker-hyperpod/LifecycleScripts/base-config/utils/install_enroot_pyxis.sh
```diff
@@ -64,113 +64,189 @@ if [ ! -d $SLURM_INSTALL_DIR ]; then
...
-    mkdir -p /fsx/enroot
-    chmod 1777 /fsx/enroot
+    # Configure FSX for enroot cache if available
+    if [[ $(mount | grep /fsx) ]]; then
+        sed -i -e 's|^\(ENROOT_CACHE_PATH  *\).*$|\1/fsx/enroot|' /etc/enroot/enroot.conf
+        mkdir -p /fsx/enroot
+        chmod 1777 /fsx/enroot
+    fi
+else
+    echo "[INFO] Skipping mount-aware enroot.conf tuning (DLAMI HostAgent handles NVMe / /opt/sagemaker / /fsx)."
```
svnisar: The cp above always overwrites enroot.conf with static paths (`ENROOT_CACHE_PATH=/opt/enroot`, `/tmp`). LCS runs after the DLAMI HostAgent, so this wipes the HostAgent's mount-aware paths and skipping the sed means nothing restores them. pre-PR, cp + sed are a pair; but post-PR this would drop the sed half.

As a result on every cluster, enroot cache lands on the small root volume instead of FSx/NVMe. First large training container fills root → node degrades/drops (issue #427). 

We should keep the enroot deb-install skip, but un-guard the sed so it always runs after the cp.

mayankgupta14: You're right, thank you for catching this. `cp` and `sed` are a pair by design; `cp` brings LCS opinionated flags, `sed` rescues the paths from the `cp`'s static `/opt/enroot` defaults. Guarding the sed alongside the cp would leave `ENROOT_CACHE_PATH=/opt/enroot` on cluster after LCS ran, filling the root volume on the first large container pull (issue #427).

Pushing an update that removes the `DLAMI_ENROOT_INSTALLED` gate around the mount-aware sed block. The enroot deb-install skip is kept (it's independent, no dependency). The pyxis skip and plugstack.conf idempotency work are unchanged.

My live test happened to show `ENROOT_CACHE_PATH=/fsx/enroot` because HostAgent re-tuned the file after LCS copy, but that's a race window, not a real fix. The corrected version keeps `cp` and `sed` together so LCS leaves the file in the correct state on exit regardless of HostAgent behavior.

svnisar (approved):

PR #41, opened 2023-11-13
Update .gitignore

KeitaW:
Avoid accidental commit of `sqsh` files.

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #15, opened 2023-10-06
Add a markdown on setting-up AWS ML GPU stacks


verdimrc:
what goes to ami, what goes to container.

PR #1222, opened 2026-08-04
docs(aws-pcs): improve JUPYTER + USER-MANAGEMENT + IAM walkthroughs (verbatim-verified)

DaisukeMiyamoto:
## Summary

Refresh of the customer-facing docs (`USER-MANAGEMENT.md`, `JUPYTER.md`, `IAM.md`) after end-to-end verbatim verification on a live Tokyo cluster, plus a related script bug fix, a cluster.yaml default tweak, regression tests, and a new PCS-Ready DLAMI version-history doc.

- `USER-MANAGEMENT.md` opens with the single-user default and treats multi-user as opt-in. Adds a §2 walkthrough that takes a first-time admin from recovering the admin password through adding a user, logging in, and running an `srun` on a compute node, then reorganises the remaining operations and Slurm accounting for day-to-day use.
- `JUPYTER.md` fixes the friction in Steps 1–3, adds a notebook cell that verifies GPU visibility, and documents a token-less mode for single-user clusters only.
- `IAM.md` is trimmed to a role table plus a short Considerations section.
- `PCS-READY-DLAMI.md` is new: version history of the AWS-published PCS-Ready DLAMI (Ubuntu 24.04).

## Test plan

Deployed a fresh cluster in `ap-northeast-1` with `DirectoryService=OpenLDAP-LoginNode`, `ManagedAccounting=enabled`, `AccountingPolicyEnforcement=associations,limits,safe`, and `MonitoringStack=Prometheus-LoginNode`. Ran the fenced blocks in README §6/§7/§8.2, USER-MANAGEMENT §2/§3/§4, JUPYTER Step 1–3, and the load-bearing blocks in `multi-user-test.md` verbatim. `lint-docs.sh` PASS.

One gap surfaced during verification and fixed in this PR: on clusters with `AccountingPolicyEnforcement` set, the README §7 `srun` example is rejected because the submitter must be registered in Slurm accounting first. Added a short callout in §7 that points to USER-MANAGEMENT §4 for the details.

### DIFF
--- a/architectures/aws-pcs/assets/scripts/ldap-add-user.sh
+++ b/architectures/aws-pcs/assets/scripts/ldap-add-user.sh
@@ -49,6 +49,27 @@ if [ -z "${LDAP_ADMIN_PASSWORD:-}" ]; then
     echo
 fi
 
+# Refuse a uidNumber that another entry already owns. LDAP itself won't reject
+# this — uidNumber is not part of the DN — so two users would silently share
+# the same POSIX UID and, on the shared /home + /fsx, become the same principal.
+# Same check for the username (its DN would be unique, but shadowing an
+# existing account is still a mistake). ldapsearch here is unauthenticated
+# and read-only; a lookup failure (LDAP down) still lets us continue and
+# surface the real error at ldapadd time.
+if command -v ldapsearch >/dev/null 2>&1; then
+    if OWNER=$(ldapsearch -x -LLL -H ldap://localhost -b "${LDAP_DOMAIN_SUFFIX}" \
+                 "(uidNumber=${USER_UID})" uid 2>/dev/null | sed -n 's/^uid: //p' | head -1) \
+       && [ -n "${OWNER}" ]; then
+        echo "uidNumber ${USER_UID} is already used by '${OWNER}'. Pick a different uid." >&2
+        exit 1
+    fi
+    if ldapsearch -x -LLL -H ldap://localhost -b "${LDAP_DOMAIN_SUFFIX}" \
+         "(uid=${USERNAME})" dn 2>/dev/null | grep -q .; then
+        echo "Username '${USERNAME}' already exists in the directory." >&2
+        exit 1
+    fi
+fi
+
 # Create user entry
 LDIF=$(cat <<EOF
 dn: uid=${USERNAME},ou=People,${LDAP_DOMAIN_SUFFIX}
--- a/architectures/aws-pcs/assets/scripts/ldap-add-user.sh
+++ b/architectures/aws-pcs/assets/scripts/ldap-add-user.sh
@@ -35,6 +35,32 @@ if [[ ! "${USER_GID}" =~ ^[0-9]+$ ]]; then
     exit 1
 fi
 
+# Refuse UIDs outside the range this cluster reserves for LDAP users
+# (10001-59999). The DLAMI's local accounts sit at 1000 (ubuntu) and other
+# system UIDs live below 10000; a shadowing LDAP entry at those UIDs would
+# `chown -R "${USER_UID}:${USER_GID}" /home/<local-name>` and destroy the
+# local user's HOME on the shared /home (documented lower bound in
+# USER-MANAGEMENT.md §3.1).
+if [ "${USER_UID}" -lt 10001 ] || [ "${USER_UID}" -gt 59999 ]; then
+    echo "uid '${USER_UID}' is outside the LDAP-user range 10001-59999 (see USER-MANAGEMENT.md §3.1)." >&2
+    exit 1
+fi
+
+# Reject any username/UID that already resolves via NSS (local /etc/passwd,
+# SSSD/LDAP, or any other configured source). LDAP-only checks below miss
+# local shadowing: e.g. `ubuntu` has UID 1000 in /etc/passwd, and adding an
+# LDAP `ubuntu` would append the pubkey to /home/ubuntu/.ssh/authorized_keys
+# and chown /home/ubuntu to the new numeric UID — breaking the DLAMI account.
+if getent passwd "${USERNAME}" >/dev/null; then
+    echo "Username '${USERNAME}' already exists (local passwd or LDAP via NSS)." >&2
+    exit 1
+fi
+if getent passwd "${USER_UID}" >/dev/null; then
+    EXISTING=$(getent passwd "${USER_UID}" | cut -d: -f1)
+    echo "uid ${USER_UID} is already used by '${EXISTING}' (local passwd or LDAP via NSS)." >&2
+    exit 1
+fi
+
 # Auto-detect LDAP config from sssd.conf or environment
 LDAP_DOMAIN_SUFFIX="${LDAP_DOMAIN_SUFFIX:-$(sed -n 's/^ldap_search_base[[:space:]]*=[[:space:]]*//p' /etc/sssd/sssd.conf 2>/dev/null || echo 'dc=cluster,dc=internal')}"
 LDAP_DOMAIN_SUFFIX="${LDAP_DOMAIN_SUFFIX:-dc=cluster,dc=internal}"
--- a/architectures/aws-pcs/docs/IAM.md
+++ b/architectures/aws-pcs/docs/IAM.md
@@ -30,17 +30,6 @@ plus the extra permissions the all-in-one template needs because it provisions
 VPC + FSx + IAM roles itself (the AWS reference policy assumes those already
 exist). Review and tighten before production use.
 
-**Login-node access is scoped to one stack.** The user policy conditions
-`ssm:StartSession` on `ssm:resourceTag/Name` equalling
-`<ClusterStackName>-login` (exact match, no wildcards) — the Name tag every
-login node carries by default. The tag is operator-mutable; if you re-tag
-the login node, update the policy condition to match.
-
-**The admin policy is split into core + Image Builder** because the combined
-document exceeds IAM's 6,144-character per-policy limit. The core covers a
-normal deploy; the Image Builder add-on is only needed for the standalone
-DLAMI builder.
-
 **Pairing with AWS-managed policies.** For a smaller customer-managed
 surface, attach AWS-managed policies for parts of the stack and trim the
 matching statements: `AWSCloudFormationFullAccess`, `AmazonFSxFullAccess`,
@@ -49,11 +38,6 @@ matching statements: `AWSCloudFormationFullAccess`, `AmazonFSxFullAccess`,
 public-share); prefer the customer-managed EC2 statements in the template.
 There is no `AmazonPCSFullAccess`, so the PCS portion has to stay
 customer-managed.
-
-**Private template bucket.** The admin policy grants no `s3:GetObject`
-because the production templates live in the public `awsome-distributed-ai`
-bucket (CFN fetches `--template-url` anonymously). If you host the templates
-in a private bucket, grant `s3:GetObject` on that bucket separately.
 
 **Not covered by these policies:**
 - The compute instance role (passed to EC2 by `cluster.yaml`) — use the
--- a/architectures/aws-pcs/docs/IAM.md
+++ b/architectures/aws-pcs/docs/IAM.md
@@ -8,7 +8,7 @@ adding existing IAM users at deploy time).
 | Role | Launch |
 |---|---|
 | **Cluster admin** — deploys/updates/deletes clusters. Full CRUD on CloudFormation, PCS, EC2 (VPC/SG/launch templates/placement groups/NAT/EIP), FSx, scoped IAM, SSM Parameter Store, KMS, Secrets Manager, and (optionally) Image Builder. **Broad — do not hand to every engineer.** | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-admin-iam.yaml&stackName=pcs-cluster-admins) |
-| **Cluster user** — engineers running jobs on an existing cluster. SSM session to the **login node only**, port-forward Grafana, read the Grafana password, read PCS cluster/queue status. Cannot create, modify, or delete anything, and cannot shell into compute nodes. **Safe to hand out widely.** | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-user-iam.yaml&stackName=pcs-cluster-users) |
+| **Cluster user** — engineers running jobs on an existing cluster. SSM session to the **login node only**, port-forward Grafana, read the Grafana password, read PCS cluster/queue status. **No AWS-API mutations** (cannot create, modify, or delete cluster resources) and **no compute-node sessions**. The login-node SSM shell it grants runs as `ssm-user` with passwordless sudo (SSM default), so on the login node it is effectively root — meaning read access to shared `/home`, Slurm accounting, and the instance role's `/pcs/<id>/ldap/*` SSM secret; treat it as trusted-operator-scope, not arbitrary-viewer-scope. | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-user-iam.yaml&stackName=pcs-cluster-users) |
 
 Both templates take an optional `AttachUsers` parameter (comma-separated
 existing IAM user names) so you can wire up group membership at deploy time,
@@ -17,7 +17,12 @@ or add users later via the IAM console. The cluster-admin template's
 admin will also deploy the standalone DLAMI builder
 (`pcs-ready-dlami-with-enroot-pyxis.yaml`). The cluster-user template
 requires `ClusterStackName` and scopes SSM session access to that one
-cluster's login node — deploy one stack of it per cluster.
+cluster's login node — deploy one stack of it per cluster. That scoping
+keys on the EC2 `Name` tag (`ssm:resourceTag/Name = <ClusterStackName>-login`),
+which is operator-mutable: re-tagging a login node changes who can reach it,
+and if another instance in the account happens to carry a matching `Name`
+tag it would also satisfy the condition. Treat `Name` as part of this
+policy's trust surface.
 
 ---
 
--- a/architectures/aws-pcs/docs/IAM.md
+++ b/architectures/aws-pcs/docs/IAM.md
@@ -5,68 +5,19 @@ responsibilities. Each ships a CloudFormation template that creates the
 customer-managed IAM policy and an IAM group with it attached (optionally
 adding existing IAM users at deploy time).
 
-| Role | Can do | Template · Launch |
-|---|---|---|
-| **Cluster admin** — deploys/updates/deletes clusters | Full CRUD on CloudFormation, PCS, EC2 (VPC/SG/launch templates/placement groups/NAT/EIP), FSx, scoped IAM, SSM Parameter Store, KMS, Secrets Manager, and (optionally) Image Builder. **Broad — do not hand to every engineer.** | [`cluster-admin-iam.yaml`](../assets/cluster-admin-iam.yaml) · [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-admin-iam.yaml&stackName=pcs-cluster-admins) |
-| **Cluster user** — engineers running jobs on an existing cluster | SSM session to the **login node only**, port-forward Grafana, read the Grafana password, read PCS cluster/queue status. Cannot create, modify, or delete anything, and cannot shell into compute nodes. **Safe to hand out widely.** | [`cluster-user-iam.yaml`](../assets/cluster-user-iam.yaml) · [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-user-iam.yaml&stackName=pcs-cluster-users) |
+| Role | Launch |
+|---|---|
+| **Cluster admin** — deploys/updates/deletes clusters. Full CRUD on CloudFormation, PCS, EC2 (VPC/SG/launch templates/placement groups/NAT/EIP), FSx, scoped IAM, SSM Parameter Store, KMS, Secrets Manager, and (optionally) Image Builder. **Broad — do not hand to every engineer.** | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-admin-iam.yaml&stackName=pcs-cluster-admins) |
+| **Cluster user** — engineers running jobs on an existing cluster. SSM session to the **login node only**, port-forward Grafana, read the Grafana password, read PCS cluster/queue status. Cannot create, modify, or delete anything, and cannot shell into compute nodes. **Safe to hand out widely.** | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-user-iam.yaml&stackName=pcs-cluster-users) |
 
----
-
-## Deploying the policies
-
-From the CLI (use `--template-body` against a local checkout for a pre-merge
-sandbox test):
-
-```bash
-# Admin: create the policies + group, attach existing users, include Image Builder perms
-aws cloudformation create-stack \
-  --stack-name pcs-cluster-admins \
-  --template-body file://architectures/aws-pcs/assets/cluster-admin-iam.yaml \
-  --parameters ParameterKey=AttachUsers,ParameterValue=alice,bob \
-               ParameterKey=AttachImageBuilderPolicy,ParameterValue=true \
-  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM
-
-# User: create the policy + group, attach existing users
-# ClusterStackName scopes SSM session access to that one cluster's login node —
-# deploy one stack of this template per cluster.
-aws cloudformation create-stack \
-  --stack-name pcs-cluster-users-pcs-ml-cluster \
-  --template-body file://architectures/aws-pcs/assets/cluster-user-iam.yaml \
-  --parameters ParameterKey=ClusterStackName,ParameterValue=pcs-ml-cluster \
-               ParameterKey=AttachUsers,ParameterValue=carol,dave \
-  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM
-```
-
-`AttachUsers` is optional — leave it empty and add users to the group later
-via IAM console/CLI. `AttachImageBuilderPolicy` defaults to `false`; enable
-it only if the admin will also deploy the standalone DLAMI builder
-(`pcs-ready-dlami-with-enroot-pyxis.yaml`).
-
-### What the cluster user can do once attached
-
-```bash
-STACK_NAME=pcs-ml-cluster        # your CloudFormation stack name
-AWS_REGION=us-east-1             # your region
-
-# Find the login node's instance ID via the PCS API — no dependency on tag naming
-CLUSTER_ID=$(aws cloudformation describe-stacks --stack-name "$STACK_NAME" --region "$AWS_REGION" --query 'Stacks[0].Outputs[?OutputKey==`ClusterId`].OutputValue' --output text)
-[ -n "$CLUSTER_ID" ] && [ "$CLUSTER_ID" != "None" ] || { echo "No ClusterId — check STACK_NAME/AWS_REGION"; return 1; }
-
-LOGIN_CNG_ID=$(aws pcs list-compute-node-groups --cluster-identifier "$CLUSTER_ID" --region "$AWS_REGION" --query 'computeNodeGroups[?name==`login`].id' --output text)
-LOGIN_INSTANCE_ID=$(aws ec2 describe-instances --region "$AWS_REGION" --filters "Name=tag:aws:pcs:compute-node-group-id,Values=$LOGIN_CNG_ID" "Name=instance-state-name,Values=running" --query 'Reservations[0].Instances[0].InstanceId' --output text)
-
-# Open a session
-aws ssm start-session --target "$LOGIN_INSTANCE_ID" --region "$AWS_REGION"
-
-# Port-forward Grafana (443 -> 8443), then open https://localhost:8443/grafana/
-aws ssm start-session --target "$LOGIN_INSTANCE_ID" --region "$AWS_REGION" \
-  --document-name AWS-StartPortForwardingSession \
-  --parameters '{"portNumber":["443"],"localPortNumber":["8443"]}'
-
-# Read the Grafana admin password
-aws ssm get-parameter --name "/pcs/${CLUSTER_ID}/grafana/admin-password" --region "$AWS_REGION" \
-  --with-decryption --query 'Parameter.Value' --output text
-```
+Both templates take an optional `AttachUsers` parameter (comma-separated
+existing IAM user names) so you can wire up group membership at deploy time,
+or add users later via the IAM console. The cluster-admin template's
+`AttachImageBuilderPolicy` defaults to `false`; set it `true` only if the
+admin will also deploy the standalone DLAMI builder
+(`pcs-ready-dlami-with-enroot-pyxis.yaml`). The cluster-user template
+requires `ClusterStackName` and scopes SSM session access to that one
+cluster's login node — deploy one stack of it per cluster.
 
 ---
 
@@ -85,13 +36,6 @@ exist). Review and tighten before production use.
 login node carries by default. The tag is operator-mutable; if you re-tag
 the login node, update the policy condition to match.
 
-**Combined CRUD is intentional, not a mistake.** The admin policy covers
-create + update + delete in one policy because (1) CFN rollback on a failed
-Create requires Delete actions, (2) UpdateStack is operationally a superset of
-Create (it may replace resources), and (3) drift detection during Update calls
-Describe across every service. If you want a read-only variant, reduce the same
-actions to `*:Describe*` / `*:Get*` / `*:List*` for an auditor role.
-
 **The admin policy is split into core + Image Builder** because the combined
 document exceeds IAM's 6,144-character per-policy limit. The core covers a
 normal deploy; the Image Builder add-on is only needed for the standalone
@@ -106,44 +50,17 @@ public-share); prefer the customer-managed EC2 statements in the template.
 There is no `AmazonPCSFullAccess`, so the PCS portion has to stay
 customer-managed.
 
-### Not covered by these policies
+**Private template bucket.** The admin policy grants no `s3:GetObject`
+because the production templates live in the public `awsome-distributed-ai`
+bucket (CFN fetches `--template-url` anonymously). If you host the templates
+in a private bucket, grant `s3:GetObject` on that bucket separately.
 
-- **The compute instance role** (passed to EC2 by `cluster.yaml`) — provisioned
-  by the templates themselves; use the AWS-managed `AWSPCSComputeNodePolicy`.
-- **The Image Builder build instance role** — use the AWS-managed
+**Not covered by these policies:**
+- The compute instance role (passed to EC2 by `cluster.yaml`) — use the
+  AWS-managed `AWSPCSComputeNodePolicy`.
+- The Image Builder build instance role — use the AWS-managed
   `EC2InstanceProfileForImageBuilder` /
   `EC2InstanceProfileForImageBuilderECRContainerBuilds`.
-- **Fine-grained per-cluster scoping** — both policies use `Resource: "*"` for
-  many EC2/VPC actions because resource-level scoping there is limited. This is
-  a deliberate sample-grade choice.
-
-### Refining to least-privilege via CloudTrail
-
-To generate a tighter policy from real usage:
-
-1. Deploy a representative cluster in a sandbox account with broad permissions on
-   the deploying principal (so nothing fails for spurious IAM reasons).
-2. Let the full lifecycle run — deploy, then `delete-stack` — so CloudTrail
-   captures every API call.
-3. Generate a policy from CloudTrail with IAM Access Analyzer
-   (`aws accessanalyzer start-policy-generation` → `get-generated-policy`), then
-   diff against the template's statements. Access Analyzer output is usually
-   *narrower* on actions but leaves `Resource: "*"`; the template's resource ARNs
-   are usually the keepers.
-
-The same approach works for the user policy — exercise the user workflows
-(start a session, port-forward Grafana, terminate it) in a sandbox, then narrow.
-
----
-
-## Verifying the policies
-
-To confirm the admin policy can deploy a cluster end-to-end and the user
-policy is correctly constrained (login-only SSM, no LDAP-password access),
-see the reproducible procedure in [tests/iam-test.md](../tests/iam-test.md).
-
-> **Private template bucket:** the admin policy grants no `s3:GetObject`
-> because the production templates live in the public
-> `awsome-distributed-ai` bucket (CFN fetches `--template-url`
-> anonymously). If you host the templates in a private bucket, grant
-> `s3:GetObject` on that bucket separately.
+- Fine-grained per-cluster scoping — both policies use `Resource: "*"` for
+  many EC2/VPC actions where resource-level scoping is limited. Sample-grade,
+  deliberate.
--- a/architectures/aws-pcs/docs/IAM.md
+++ b/architectures/aws-pcs/docs/IAM.md
@@ -1,32 +1,20 @@
 # IAM Permissions Guide
 
 The cluster distinguishes **two human roles** with very different
-responsibilities, and ships a ready-to-deploy IAM policy stack for each:
+responsibilities. Each ships a CloudFormation template that creates the
+customer-managed IAM policy and an IAM group with it attached (optionally
+adding existing IAM users at deploy time).
 
-| Role | Who | What they can do | Template |
-|---|---|---|---|
-| **Cluster admin** | The person who deploys/updates/deletes the cluster | Full CRUD on the infrastructure: CloudFormation, PCS, EC2 (VPC/SG/launch templates/placement groups/NAT/EIP), FSx, scoped IAM, SSM Parameter Store, KMS, Secrets Manager, and (optionally) Image Builder | [`cluster-admin-iam.yaml`](../assets/cluster-admin-iam.yaml) |
-| **Cluster user** | Engineers who just run jobs on an existing cluster | SSM session **to the login node only**, port-forward Grafana, read the Grafana password, read PCS cluster/queue status. **Cannot create, modify, or delete anything**, and cannot open shells on compute nodes | [`cluster-user-iam.yaml`](../assets/cluster-user-iam.yaml) |
-
-Splitting the roles means the deployer's broad permissions never have to be
-handed to every engineer who just wants to `srun`, and the user role can be
-given out widely — it can't accidentally delete the cluster or get a shell on a
-compute node.
+| Role | Can do | Template · Launch |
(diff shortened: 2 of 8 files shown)


KeitaW (commented):
## Review Batch 1/5 — Slurm Accounting Runbooks (USER-MANAGEMENT §4 + multi-user-test Part F)

This is a strong docs refresh overall — the §2 first-user walkthrough and the JUPYTER step fixes are exactly the kind of friction removal these docs needed, and the live-verifiable claims I could check all held up (details in the final batch). The one area that doesn't meet the PR's own "verbatim-verified" bar is the new accounting material: Part F and USER-MANAGEMENT §4.2 each contain commands whose documented expected output is unreachable as committed — details in the 7 inline comments of this batch. The common thread is that Part F cannot have completed as written, so I'd suggest re-running it once these are fixed and noting the run in the test plan.

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -488,3 +596,102 @@ sudo su - charlie -c 'export PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:$PATH;
...
+**Prerequisites**: same as the accounting test (`ManagedAccounting=enabled`
++ `DirectoryService=OpenLDAP-LoginNode`); alice / bob / carol as LDAP
+users (add carol here if only alice/bob exist).
+
+### F1. Project accounts and quota
+
+```bash
+sudo LDAP_ADMIN_PASSWORD="$ADMIN_PW" ldap-add-user.sh carol 10003 3000
+
+sudo sacctmgr -i add account proj_physics   Description="Physics group"
```
KeitaW: ### Part F's `sudo sacctmgr` invocations fail under `secure_path` — the section's own sibling doc explains why (blocking)

F1, F3, and F5 invoke bare `sudo sacctmgr` (lines 620–626, 654, 667, 692–696 — including cleanup). USER-MANAGEMENT §4 — added in this same PR — explains precisely why that form fails: Ubuntu's `sudo` resets `PATH` via `secure_path`, so "`sudo sacctmgr` from a shell where you only `export PATH=…` still fails to find the binary", and §4.1 accordingly uses `sudo $S -i …` with the absolute path throughout. As committed, every accounting mutation in Part F — including the F5 cleanup — dies with `sudo: sacctmgr: command not found`, which also means Part F can't have run verbatim in this form. I'd suggest opening Part F by binding `S=/opt/aws/pcs/scheduler/slurm-25.11/bin/sacctmgr` (as §4 does) and using `sudo $S -i …` in F1/F3/F5. Could you re-run Part F after the fix and confirm?

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -488,3 +596,102 @@ sudo su - charlie -c 'export PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:$PATH;
...
+```
+
+Expected: three users listed under the two project accounts; alice's
+`GrpTRESRunMins` column shows `cpu=6000`.
+
+### F2. Blog-style job submission with `--account=`
+
+```bash
+sudo su - alice -c 'export PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:$PATH; \
+  echo "hostname; id" > /home/alice/smalljob.sh && chmod +x /home/alice/smalljob.sh; \
```
KeitaW: ### F2's batch script has no shebang — `sbatch` rejects it before submission (blocking)

`echo "hostname; id" > /home/alice/smalljob.sh` produces a script whose first line is not `#!…`. `sbatch` hard-rejects such scripts at submit time — [`src/sbatch/sbatch.c`](https://github.com/SchedMD/slurm/blob/master/src/sbatch/sbatch.c) in the Slurm source: *"This does not look like a batch script. The first line must start with #! followed by the path to an interpreter."* So F2's expected `Submitted batch job <n>` / `COMPLETED` output is unreachable. The existing C3 sidesteps this with `--wrap` — either form works:

```suggestion
  printf '#!/bin/bash\nhostname; id\n' > /home/alice/smalljob.sh && chmod +x /home/alice/smalljob.sh; \
```

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -488,3 +596,102 @@ sudo su - charlie -c 'export PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:$PATH;
...
+[USER-MANAGEMENT.md §4.1](../docs/USER-MANAGEMENT.md#41-worked-example-project-accounts--per-user-quota--reports).
+
+**Prerequisites**: same as the accounting test (`ManagedAccounting=enabled`
++ `DirectoryService=OpenLDAP-LoginNode`); alice / bob / carol as LDAP
+users (add carol here if only alice/bob exist).
+
+### F1. Project accounts and quota
+
+```bash
+sudo LDAP_ADMIN_PASSWORD="$ADMIN_PW" ldap-add-user.sh carol 10003 3000
```
KeitaW: ### F1 creates `carol` with UID 10003 — already taken by `charlie`, so the PR's own new guard aborts the scenario (blocking)

Part E (line 580) creates `charlie 10003 3000`. F1 then unconditionally runs `ldap-add-user.sh carol 10003 3000` — and the duplicate-`uidNumber` guard this very PR adds to `ldap-add-user.sh` now (correctly!) refuses it: `uidNumber 10003 is already used by 'charlie'`. On a directory where Part E ran, Part F aborts at its first command. The prerequisites line ("add carol here if only alice/bob exist") also conflicts with the unconditional add — if carol exists, the username guard trips instead. I'd pick the next free UID and make the add conditional:

```suggestion
getent passwd carol >/dev/null || sudo LDAP_ADMIN_PASSWORD="$ADMIN_PW" ldap-add-user.sh carol 10004 3000
```

architectures/aws-pcs/docs/USER-MANAGEMENT.md
```diff
@@ -382,184 +366,229 @@ EOF
...
+With `AccountingPolicyEnforcement=associations,limits,safe` a job
+that would put alice over her running-minutes budget is **accepted
+at submit time but held pending** with reason
+`AssocGrpCPURunMinutesLimit`; it starts only once earlier work drains
+the budget. A job that fits proceeds normally:
+
 ```bash
-srun -N 1 -n 1 -p cpu1 bash -c 'sudo sss_cache -E; sleep 2; getent passwd alice'
+# As alice
+sbatch --account=proj_physics --nodes=1 --ntasks-per-node=8 --time=30:00 myjob.sh
```
KeitaW: ### §4.2's hold demo can't trigger: the scripts don't exist, `-p` is missing, and 240 < 6000 (blocking)

Three independent problems make the documented `squeue` output (`PD AssocGrpCPURunMinutesLimit`) unreachable as written:

1. Neither `myjob.sh` nor `smalljob.sh` is created by any prior step — the first `sbatch` fails with `Unable to open file myjob.sh`.
2. Neither `sbatch` passes `--partition`, and this repo's own docs (JUPYTER.md Step 2, README §7) are emphatic that PCS clusters have no default partition — a bare `sbatch` is rejected with *"invalid partition specified"*. (The expected `squeue` output even shows `cpu1`.)
3. The quota math: the limit just set is `cpu=6000` CPU-minutes, but the "over-budget" job requests 8 CPUs × 30 min = **240** CPU-minutes — it fits with room to spare, so no hold occurs. The AWS blog this section reproduces triggers the hold only because ~95 CPU-hours had already been consumed in its narrative; that precondition doesn't hold here. Part F's own F3 gets this right by tightening the limit to `cpu=60` first.

I'd suggest mirroring F3: `--wrap='sleep 1000'`, `-p cpu1`, and either tighten the limit below the request or state the prior-usage precondition explicitly.

architectures/aws-pcs/docs/USER-MANAGEMENT.md
```diff
@@ -382,184 +366,229 @@ EOF
...
+[Introducing managed accounting for AWS Parallel Computing Service](https://aws.amazon.com/blogs/hpc/introducing-managed-accounting-for-aws-parallel-computing-service/):
+register users under project accounts, cap their usage, and read
+back utilization. Extra managed-accounting charges (an hourly fee
+tied to the controller size + per-GB-month storage governed by the
+cluster's Default Purge Time) apply on top of the base cluster —
+see the blog for pricing detail.
+
+Registering users in accounting is optional unless the cluster was
+deployed with `AccountingPolicyEnforcement=associations,limits,safe`
+(or the stricter `associations,limits`) — with either enforcement on,
```
KeitaW: ### §4's "the stricter `associations,limits`" mode isn't deployable from these templates (should fix)

§4 intro and the §4.2 callout both describe `AccountingPolicyEnforcement=associations,limits` (no `safe`) as the stricter alternative that rejects at submit time, and multi-user-test F3 repeats it. But both templates only allow `none` and `'associations,limits,safe'` — see the `AccountingPolicyEnforcement` AllowedValues in `assets/cluster.yaml` and `assets/pcs-ml-cluster-deploy-all.yaml` — so a reader following this doc into a deploy hits CloudFormation parameter validation failure, and the claimed submit-time-rejection behavior of the strict mode can't have been verified on a cluster from these templates. I'd either scope the text to what the templates offer ("the strict variant exists in Slurm but isn't offered by these templates") or add the AllowedValue and verify the behavior before documenting it.

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -488,3 +596,102 @@ sudo su - charlie -c 'export PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:$PATH;
...
+
+sudo sacctmgr -i add account proj_physics   Description="Physics group"
+sudo sacctmgr -i add account proj_chemistry Description="Chemistry group"
+sudo sacctmgr -i add user alice Account=proj_physics
+sudo sacctmgr -i add user bob   Account=proj_physics
+sudo sacctmgr -i add user carol Account=proj_chemistry
+
+sudo sacctmgr -i modify user alice set GrpTRESRunMins=cpu=6000
+
+sacctmgr show user alice bob carol format=User,Account,DefaultAccount,GrpTRESRunMins
```
KeitaW: ### F1's `sacctmgr show` omits `WithAssoc` — the very flag §4.1 explains is required (nit)

`format=…Account,DefaultAccount,GrpTRESRunMins` requests association-level fields, and USER-MANAGEMENT §4.1 (this PR) adds the sentence "`WithAssoc` joins the user's association row so per-account attributes actually appear in the output" — but F1's own show command drops the flag, so the expected verification ("alice's `GrpTRESRunMins` column shows `cpu=6000`") won't render:

```suggestion
sacctmgr show user alice bob carol WithAssoc format=User,Account,DefaultAccount,GrpTRESRunMins
```

architectures/aws-pcs/docs/USER-MANAGEMENT.md
```diff
@@ -382,184 +366,229 @@ EOF
...
-# User 'alice' logs in via SSH and runs:
-srun -p cpu1 -N 1 -n 1 bash -c 'whoami; hostname'
-sbatch --partition=cpu1 my-training.sbatch
+# Weekly all-user job listing
+sacct --starttime=$(date -d "7 days ago" +%Y-%m-%d) \
+  --format="JobID,User,JobName,Partition,Account,AllocCPUS,State,ExitCode"
+
+# Monthly cluster utilization by account/user
+sreport cluster AccountUtilizationByUser \
+  start=2026-04-01 end=2026-04-30 -t percent \
```
KeitaW: ### `sreport` end dates exclude the final day — "monthly" reports drop April 30 / March 31 (nit)

`sreport` date-only times default to 00:00:00 of that day (https://slurm.schedmd.com/sreport.html), so `end=2026-04-30` cuts off at midnight *entering* Apr 30 — the last day of the month is excluded (same for `end=2026-03-31` on line 472, and for `end=$(date +%Y-%m-%d)` in multi-user-test F4, which excludes today's just-submitted jobs — precisely the ones the test wants to see). Use the first day of the next month / `end=now`:

```suggestion
  start=2026-04-01 end=2026-05-01 -t percent \
```

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -136,6 +159,91 @@ Verify on compute (after cache expires or forced refresh):
...
+cat > /tmp/m.ldif <<EOF
+dn: cn=ml-team,ou=Groups,dc=cluster,dc=internal
+changetype: modify
+add: memberUid
+memberUid: testuser2
+EOF
+ldapmodify -x -H ldap://localhost -D "cn=admin,dc=cluster,dc=internal" -W -f /tmp/m.ldif
+
+sudo sss_cache -E && sleep 2
+id testuser2
```
KeitaW, line 191: ### B8 verifies membership for `testuser2` — deleted two sections earlier in B6 (should fix)

B6 (line ~149) `ldapdelete`s `testuser2` and verifies `getent` returns nothing. New B8 then adds `memberUid: testuser2` to `ml-team` (LDAP accepts the dangling string — `memberUid` is unconstrained) and expects `id testuser2` to print `uid=10002(testuser2) … 3001(ml-team)` — impossible for a deleted user, so B8's expected output is unreachable in document order. I'd use `testuser1` for the member-add half, or move B6's deletion after B8.

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -136,6 +159,91 @@ Verify on compute (after cache expires or forced refresh):
...
+
+From an operator laptop that has AWS CLI + IAM permissions but no direct
+SSH access to the VPC. Uses the PCS API to derive the login instance ID
+(no reliance on the mutable `Name` tag).
+
+```bash
+# Generate a keypair and install its public key on user creation:
+ssh-keygen -t ed25519 -N "" -f ~/.ssh/pcs-testuser1 -C "testuser1@laptop"
+sudo -E ldap-add-user.sh testuser1 10001 3000 "$(cat ~/.ssh/pcs-testuser1.pub)"
+# (Skip if B1 already created testuser1 without a key — add it with:
```
KeitaW: ### B9's opening block mixes laptop and login-node commands, always fails under the new duplicate guard, and its `ldapmodify` fallback can't work (should fix)

Three issues in the setup preamble:

1. The section header says "From an operator laptop", but `sudo -E ldap-add-user.sh …` must run on the login node (the helper lives at `/usr/local/bin` there and binds `ldap://localhost`).
2. In document order B1 has always already created `testuser1`, so this add now *always* exits 1 under the duplicate-username guard added by this same PR — the "(Skip if B1 already created testuser1…)" parenthetical is the mainline case, not the exception.
3. The suggested fallback "add it with: ldapmodify" can't work: the helper's own comment (`assets/scripts/ldap-add-user.sh`, SSH-key block) states keys are intentionally *not* stored in LDAP — they go to `/home/<user>/.ssh/authorized_keys`. The working login-node fallback is appending the pubkey there, e.g. `echo "<pubkey>" | sudo tee -a /home/testuser1/.ssh/authorized_keys` (plus the `chmod`/`chown` the helper does).

I'd restructure: generate the keypair on the laptop, install the key on the login node (helper for a fresh user, `authorized_keys` append for an existing one), then run the SSH-over-SSM block from the laptop.

architectures/aws-pcs/docs/USER-MANAGEMENT.md
```diff
@@ -382,184 +366,229 @@ EOF
...
+> accounts"*. Read-only forms (`sacctmgr show …`, `sacct`, `sreport`)
+> work as any user.
+
+Ubuntu's `sudo` resets `PATH` (`secure_path`), so `sudo sacctmgr` from
+a shell where you only `export PATH=…` still fails to find the binary.
+Bind the absolute path to a variable and use it throughout:
 
 ```bash
-SACCTMGR=/opt/aws/pcs/scheduler/slurm-25.11/bin/sacctmgr
+S=/opt/aws/pcs/scheduler/slurm-25.11/bin/sacctmgr   # 25.05 if SlurmVersion=25.05
```
KeitaW: ### `slurm-25.11` paths are now hard-coded with only a trailing comment — the old version-switch note was deleted (should fix)

The previous doc opened with a prominent blockquote ("Slurm path matches `SlurmVersion` … replace `slurm-25.11` with `slurm-25.05` in every path"). This PR removes it; what remains is one `# 25.05 if SlurmVersion=25.05` comment on this line — but §3.3's `sudo /opt/aws/pcs/scheduler/slurm-25.11/bin/sacctmgr` and the `export PATH=…slurm-25.11…` line carry no caveat at all, so a 25.05 deployment gets `No such file or directory` with no pointer to why. Cheapest fix that covers every occurrence: derive it once, e.g. `SLURM_BIN=/opt/aws/pcs/scheduler/$(ls /opt/aws/pcs/scheduler | grep '^slurm-')/bin`, or restore a one-line note at the top of §3.

KeitaW (commented):
## Review Batch 2/5 — LDAP Test Sequencing & Version Paths

Three sequencing/consistency issues in the new test sections and the operations doc — each breaks a runbook when followed top-to-bottom. Details inline.

architectures/aws-pcs/docs/USER-MANAGEMENT.md
```diff
@@ -1,363 +1,349 @@
...
-│                                                          │
-│  ┌──────────┐     ┌──────────────┐                      │
-│  │  SSSD    │────►│  NSS / PAM   │                      │
-│  │  (client)│     │  (getent,    │                      │
-│  │          │     │   srun user) │                      │
-│  └──────────┘     └──────────────┘                      │
-└─────────────────────────────────────────────────────────┘
-```
+1. Open the deploy-all template's **Launch Stack** link in the
+   [README](../README.md#quick-start) (or navigate to CloudFormation
```
KeitaW: ### `../README.md#quick-start` — README's heading is "3. Quick Start", so the anchor doesn't resolve (nit)

README.md's heading is `## 3. Quick Start` → anchor `#3-quick-start`; `#quick-start` opens the README at the top without navigating. (`tests/lint-docs.sh`'s anchor check only covers README-internal anchors, which is why this passed lint — extending it to cross-file `../` anchors would catch both this and the Part F one.)

```suggestion
   [README](../README.md#3-quick-start) (or navigate to CloudFormation
```

architectures/aws-pcs/tests/multi-user-test.md
```diff
@@ -488,3 +596,102 @@ sudo su - charlie -c 'export PATH=/opt/aws/pcs/scheduler/slurm-25.11/bin:$PATH;
...
+
+## Part F — Blog scenario end-to-end
+
+Reproduces [Introducing managed accounting for AWS Parallel Computing
+Service](https://aws.amazon.com/blogs/hpc/introducing-managed-accounting-for-aws-parallel-computing-service/)
+on top of the existing LDAP users, exercising the customer-facing flow
+the blog demonstrates: two project accounts, a per-user cap, `--account=`
+job attribution, quota-based rejection, and the blog's own reporting
+commands. Regression guard for the walkthrough in
+[USER-MANAGEMENT.md §4.1](../docs/USER-MANAGEMENT.md#41-worked-example-project-accounts--per-user-quota--reports).
```
KeitaW: ### Part F links a USER-MANAGEMENT §4.1 heading that doesn't exist (nit)

The anchor `#41-worked-example-project-accounts--per-user-quota--reports` matches no heading in this PR's USER-MANAGEMENT.md — §4.1 is "Register users to accounts" (`#41-register-users-to-accounts`). Looks like a leftover from an earlier draft of the section title:

```suggestion
[USER-MANAGEMENT.md §4.1](../docs/USER-MANAGEMENT.md#41-register-users-to-accounts).
```

architectures/aws-pcs/docs/JUPYTER.md
```diff
@@ -56,6 +59,14 @@ python3 -m venv $HOME/jupyter-env
 $HOME/jupyter-env/bin/pip install --upgrade pip jupyterlab
 ```
 
+**To run the GPU-visibility verify cell below**, also install PyTorch here so
+the kernel can `import torch` without another install roundtrip. The
+PCS-Ready DLAMI ships CUDA 12.x, so pin the matching wheel:
+
+```bash
+$HOME/jupyter-env/bin/pip install torch --index-url https://download.pytorch.org/whl/cu124
```
KeitaW: ### "Ships CUDA 12.x → pin cu124" contradicts this PR's own DLAMI table (CUDA 13.2) — the tested wheel is cu130 (should fix)

PCS-READY-DLAMI.md (added in this same PR) lists Default CUDA **13.2** with a 12.8/12.9/13.0/13.2 stack — 12.4 isn't on the image at all — and the sibling `tests/jupyter-notebook-test.md` records the verified environment as `torch 2.12.1+cu130, driver 595.71.05 / CUDA 13.2`. Two things worth fixing: (a) the premise — a venv-installed torch wheel bundles its own CUDA runtime, so what matters is the NVIDIA driver (595.71.05 supports all current wheels), not the image's CUDA toolkit; (b) the pick — since the sibling test already validated cu130, pin that (all of cu124/cu126/cu128/cu129/cu130 exist on download.pytorch.org — verified live, 2026-08-04):

```suggestion
$HOME/jupyter-env/bin/pip install torch --index-url https://download.pytorch.org/whl/cu130
```

I'd also reword the preceding sentence — "The PCS-Ready DLAMI's driver (595.xx) supports current CUDA wheels; the cu130 build matches the image's CUDA 13.2 stack" — rather than "ships CUDA 12.x, so pin the matching wheel".

architectures/aws-pcs/docs/JUPYTER.md
```diff
@@ -174,54 +224,54 @@ costs nothing.
...
+    print(f"  [{i}] {p.name}  {p.total_memory/1024**3:.1f} GiB  SM {p.major}.{p.minor}")
+
+# 3. Prove compute runs on the allocated GPU (raises if the kernel can't launch)
+if torch.cuda.is_available():
+    x = torch.randn(4096, 4096, device="cuda:0")
+    print("matmul on cuda:0 -> sum =", float((x @ x).sum()))
+```
+
+Expected: `torch.cuda.device_count()` **equals the `--gres=gpu:N` you
+requested**. If `nvidia-smi` shows more devices than
```
KeitaW: ### "File a bug" advice labels the documented-normal GPU visibility as broken (should fix)

`tests/jupyter-notebook-test.md` documents the verified behavior on these clusters: *"the full GPU table for the node (all physical GPUs listed … allocation limits are enforced per-process via `CUDA_VISIBLE_DEVICES`, not in `nvidia-smi`)"*. The new text tells readers that exact state means "Slurm's cgroup isolation is not restricting them — file a bug against the CNG configuration". As written, every reader on the documented configuration is told to file a bug. The reliable gate is the one the cell already checks — `torch.cuda.device_count()` equals the `--gres` count; I'd replace the `nvidia-smi` sentence with "seeing all physical GPUs in `nvidia-smi` is expected — allocation is enforced per-process via `CUDA_VISIBLE_DEVICES`".

architectures/aws-pcs/docs/JUPYTER.md
```diff
@@ -153,14 +175,42 @@ cat ~/.jupyter-token-<jobid>
...
+```bash
+exec jupyter lab --no-browser --ip="$NODE_IP" --port="$PORT" \
+  --ServerApp.token='' --ServerApp.password='' \
+  --ServerApp.disable_check_xsrf=True \
+  --notebook-dir="$HOME"
+```
+
+The token file (and the `openssl rand` line above it) is no longer needed
+and can be dropped. Connect to `http://localhost:8888/lab` — no token
+required. Nothing else in Step 3 changes.
```
KeitaW: ### Token-less mode: "Nothing else in Step 3 changes" isn't quite true (nit)

Step 3's sub-step 2 is "read the token" (`cat ~/.jupyter-token-<jobid>`) — gone in this mode — and the sbatch banner's `token   = run  cat $TOKEN_FILE …` line goes stale/empty once the token lines are dropped. Tiny fix: "Steps 1 and 3 of Step 3 are unchanged; skip sub-step 2, and drop the `token` line from the banner heredoc.

KeitaW (commented):
## Review Batch 3/5 — Cross-Doc Consistency (anchors & sibling drift)

Five inline comments: two broken anchors, two places where new text contradicts a sibling doc, and one wording nit. Plus one finding in a file this PR doesn't touch:

### Complementary fix: `tests/jupyter-notebook-test.md` still teaches the old `Values=*login` discovery this PR removes from JUPYTER.md (should fix)

JUPYTER.md Step 3 now resolves the login node via the PCS API — great. Its regression companion `tests/jupyter-notebook-test.md:119` still uses `--filters "Name=tag:Name,Values=*login"`, so after this PR the test doc no longer matches the doc it tests (and `*login` matches any cluster's login node in multi-cluster accounts — the problem the PCS-API recipe exists to avoid). Note `tests/lint-docs.sh`'s banned-pattern only matches the `Values=PCS-` spelling, so this form passes lint. One hunk to port the same PCS-API snippet there would finish the migration.

KeitaW (commented):
## Review Batch 4/5 — Security Posture

Four inline comments: one on the token-less Jupyter mode (with a live-verified one-line fix), two on IAM.md claims vs what the policy enforces, and one hardening gap in the new duplicate guard.

architectures/aws-pcs/docs/JUPYTER.md
```diff
@@ -153,14 +175,42 @@ cat ~/.jupyter-token-<jobid>
...
+### Running without a token (single-user clusters only)
+
+If you're the only user on the cluster and want to skip the token
+copy-paste, swap the `--ServerApp.token=…` line in the sbatch script
+with the disabled-auth form:
+
+```bash
+exec jupyter lab --no-browser --ip="$NODE_IP" --port="$PORT" \
+  --ServerApp.token='' --ServerApp.password='' \
+  --ServerApp.disable_check_xsrf=True \
```
KeitaW: ### Drop `--ServerApp.disable_check_xsrf=True` — it's unnecessary for token-less use and removes the last browser-side guard (should fix)

Verified live (JupyterLab 4.6.2, `--ServerApp.token='' --ServerApp.password=''`, XSRF left **enabled**, 2026-08-04): `GET /lab` → 200 and sets the `_xsrf` cookie; the Lab UI's own cookie+header flow then works (`POST /api/kernels` → 201); a naive cross-site `POST` without the header → 403 `'_xsrf' argument missing`. So the token-less recipe works without this flag — and with it, any webpage open in the operator's browser while the tunnel is up can POST to `http://localhost:8888` and execute code on the compute node, which the (otherwise excellent) multi-user warning below doesn't cover. Jupyter's security docs already frame no-auth mode as needing another protective layer (https://jupyter-server.readthedocs.io/en/latest/operators/security.html); XSRF is the one layer that costs nothing here:

```suggestion
```

One wording nit in the warning while you're there: "single-user" should mean "every IAM principal who can reach the VPC/run jobs is trusted as the same person" — `DirectoryService=none` by itself doesn't guarantee a single principal.

architectures/aws-pcs/docs/IAM.md
```diff
@@ -1,88 +1,23 @@
...
+adding existing IAM users at deploy time).
+
+| Role | Launch |
+|---|---|
+| **Cluster admin** — deploys/updates/deletes clusters. Full CRUD on CloudFormation, PCS, EC2 (VPC/SG/launch templates/placement groups/NAT/EIP), FSx, scoped IAM, SSM Parameter Store, KMS, Secrets Manager, and (optionally) Image Builder. **Broad — do not hand to every engineer.** | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-admin-iam.yaml&stackName=pcs-cluster-admins) |
+| **Cluster user** — engineers running jobs on an existing cluster. SSM session to the **login node only**, port-forward Grafana, read the Grafana password, read PCS cluster/queue status. Cannot create, modify, or delete anything, and cannot shell into compute nodes. **Safe to hand out widely.** | [![Launch](../images/launch-stack.svg)](https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateUrl=https://awsome-distributed-ai.s3.amazonaws.com/templates/aws-pcs/cluster-user-iam.yaml&stackName=pcs-cluster-users) |
```
KeitaW: ### "Cannot create, modify, or delete anything … Safe to hand out widely" overstates what the user policy enforces (should fix)

The stock `cluster-user-iam.yaml` grants `ssm:StartSession` on the `SSM-SessionManagerRunShell` document, and a plain Session Manager shell lands as `ssm-user`, which has passwordless sudo by default (https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-ssm-user-permissions.html). Root on the login node reaches the instance role's `ssm:GetParameter` on `/pcs/<id>/ldap/*` — i.e. the LDAP admin password — plus shared `/home` and Slurm accounting. The policy itself is untouched by this PR (worth a follow-up: drop RunShell from the user policy, or configure Session Manager Run As), but this rewrite is where the claim gets its strongest wording, so I'd soften the row to what the policy actually enforces: no AWS-API mutations and no compute-node sessions — the login-node shell it grants is root-equivalent on the shared directory until RunShell is removed.

architectures/aws-pcs/docs/IAM.md
```diff
@@ -1,88 +1,23 @@
...
-
-# Read the Grafana admin password (CLUSTER_ID is resolved by CFN Outputs above; look it up if needed)
-CLUSTER_ID=$(aws cloudformation describe-stacks --stack-name "$STACK_NAME" --region "$AWS_REGION" \
-  --query 'Stacks[0].Outputs[?OutputKey==`ClusterId`].OutputValue' --output text)
-aws ssm get-parameter --name "/pcs/${CLUSTER_ID}/grafana/admin-password" --region "$AWS_REGION" \
-  --with-decryption --query 'Parameter.Value' --output text
-```
+(`pcs-ready-dlami-with-enroot-pyxis.yaml`). The cluster-user template
+requires `ClusterStackName` and scopes SSM session access to that one
+cluster's login node — deploy one stack of it per cluster.
```
KeitaW: ### The one-cluster scoping claim lost its "Name tag is operator-mutable" caveat (should fix)

The deleted IAM.md section carried an explicit warning that the `ssm:resourceTag/Name = <ClusterStackName>-login` condition keys on the operator-mutable `Name` tag (re-tag ⇒ the scoping silently changes). This rewrite keeps the "scopes SSM session access to that one cluster's login node" claim but drops the caveat, leaving the claim unqualified. One sentence restores it: "Scoping keys on the EC2 `Name` tag, which is operator-mutable — re-tagging a login node changes who can reach it (fails closed for this cluster, but a matching tag elsewhere in the account would satisfy the condition).

architectures/aws-pcs/assets/scripts/ldap-add-user.sh
```diff
@@ -49,6 +49,27 @@ if [ -z "${LDAP_ADMIN_PASSWORD:-}" ]; then
...
 fi
 
+# Refuse a uidNumber that another entry already owns. LDAP itself won't reject
+# this — uidNumber is not part of the DN — so two users would silently share
+# the same POSIX UID and, on the shared /home + /fsx, become the same principal.
+# Same check for the username (its DN would be unique, but shadowing an
+# existing account is still a mistake). ldapsearch here is unauthenticated
+# and read-only; a lookup failure (LDAP down) still lets us continue and
+# surface the real error at ldapadd time.
+if command -v ldapsearch >/dev/null 2>&1; then
```
KeitaW, line 85: ### The new duplicate guard checks LDAP only — a local/NSS name like `ubuntu` still passes, with a destructive follow-on (should fix)

The guard is a real improvement (and B5a locks it in — nice). One gap: it searches only `ldap://localhost`, so `ldap-add-user.sh ubuntu 10001 3000 "<key>"` passes both checks (the username regex admits `ubuntu`; UID 1000's owner isn't in LDAP), creates a shadowing LDAP `ubuntu`, appends the key to the real `/home/ubuntu/.ssh/authorized_keys`, and then `chown -R 10001:3000 /home/ubuntu` — breaking the DLAMI default account's home. Same for UIDs of local users. Since the shared `/home` is exactly where the damage lands, I'd check the full NSS view and the documented range before touching LDAP or the filesystem:

```bash
if getent passwd "${USERNAME}" >/dev/null; then echo "…exists (local or LDAP)…" >&2; exit 1; fi
if getent passwd "${USER_UID}" >/dev/null; then echo "…uid in use…" >&2; exit 1; fi
if [ "${USER_UID}" -lt 10001 ] || [ "${USER_UID}" -gt 59999 ]; then echo "…outside LDAP range 10001-59999…" >&2; exit 1; fi
```

(`getent` goes through SSSD, so it also sees LDAP — the two `ldapsearch` probes could then become the fallback for the SSSD-cold-cache window, keeping your documented fail-open behavior.)

architectures/aws-pcs/docs/PCS-READY-DLAMI.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+New builds are announced through the SNS topic
+`arn:aws:sns:us-west-2:265886551188:pcs-ready-dlami-release-notifications`
+(see the [PCS-Ready DLAMI user guide](https://docs.aws.amazon.com/pcs/latest/userguide/working-with_ami_pcs-ready-dlami.html)).
+
+## PCS-Ready DLAMI x86\_64
+
+Every published x86\_64 PCS-Ready DLAMI, newest first. AMI IDs are region-scoped
+and shown for **us-east-2**; use the `describe-images` command below the table
```
KeitaW: ### "Use the `describe-images` command below the table" — there is no command below the table (should fix)

The file ends at the release-notes link; the promised cross-region lookup command was never added. The official PCS docs page already ships the right one (https://docs.aws.amazon.com/pcs/latest/userguide/working-with_ami_pcs-ready-dlami.html) — adding it below the table with a build-date filter closes the gap:

```bash
aws ec2 describe-images --region <region> --owners amazon \
  --filters 'Name=name,Values=aws-pcs-ready-dlami-base-ubuntu2404-x86_64-<PCS build>*' \
            'Name=state,Values=available' \
  --query 'Images[0].[Name,ImageId]' --output text
```

architectures/aws-pcs/docs/PCS-READY-DLAMI.md
```diff
@@ -0,0 +1,56 @@
...
+to look up the same build in another region.
+
+Base-DLAMI component versions (kernel, NVIDIA driver, CUDA, DCGM, EFA,
+OFI-NCCL, containerd) come from the DLAMI release notes for the build named in
+the "Base DLAMI" column. Intermediate base builds without their own release
+notes inherit values from the nearest previous published build.
+
+| PCS build | AMI ID (us-east-2) | Base DLAMI | PCS Agent | Slurm | EFS Utils | Kernel | NVIDIA driver | Default CUDA | CUDA stack | DCGM | EFA | OFI-NCCL | containerd |
+| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
+| 2026-07-27 | `ami-0631f0a4211880f98` | 20260724 | 1.5.0-1 | 24.11.7-2, 25.05.8-2, 25.11.6-2 | 2.4.2 | 6.17.0-1019 | 595.71.05 | 13.2 | 12.8, 12.9, 13.0, 13.2 | 4.6.0 | 1.47.0 | 1.18.0 | v2.2.6 |
```
KeitaW, line 43: ### The table is already one build behind at review time (nit)

Verified live (2026-08-04, us-east-2): the SSM parameter `/aws/service/pcs/ami/dlami-base-ubuntu2404/x86_64/latest/ami-id` now resolves to `ami-083d41c7d9d7d10a3` — build `2026-07-29` (created 2026-08-04), same component versions as this row. Inherent to a static table, but worth (a) adding the 2026-07-29 row before merge and (b) framing the intro as a dated snapshot ("as of YYYY-MM-DD") rather than "Every published … DLAMI", so staleness reads as expected rather than as an error. The lookup command above is what lets readers self-serve past the snapshot.

KeitaW (commented):
## Review Batch 5/5 — PCS-Ready DLAMI Doc, Pre-existing Notes & Positives

Two inline comments on the new DLAMI doc, then two pre-existing observations and the things that verified clean.

### Pre-existing, out of this diff (informational — no action needed for this PR)

- **Helper passes secrets in argv:** `ldap-add-user.sh` invokes `ldapadd -w "${LDAP_ADMIN_PASSWORD}"` and `ldappasswd -w … -s "${INITIAL_PW}"` — both visible to `ps` on the multi-user login node while running. §2.1's new "never lands on the command line" sentence is true for the *operator's* shell, but a follow-up switching the helper to `-y <root-only-password-file>` would make it true end-to-end.
- **Jupyter port collisions move the server silently:** the sbatch script notes modulo-1000 port collisions but doesn't set `--ServerApp.port_retries=0`; jupyter-server retries ~50 ports on conflict, so a collided server starts on `$PORT+1` while the banner and tunnel still say `$PORT`. Adding `--ServerApp.port_retries=0` makes the collision loud instead.

### Things That Look Great

- **The duplicate-UID/username guard closes a real foot-gun** — LDAP happily accepts two entries with one `uidNumber`, and on shared `/home`+`/fsx` that's one POSIX principal. The guard's fail-open-when-LDAP-down rationale is documented in-line, and B5a pins the behavior with exit codes and expected stderr.
- **PCS-API login discovery is now consistent** across JUPYTER.md Step 3, USER-MANAGEMENT §2.3, IAM.md, and the new B9 — the last stale `Name=*login` filter in `docs/` is gone (one remains in `tests/jupyter-notebook-test.md`, noted in Batch 3).
- **The DLAMI table is accurate where I could check it.** I verified 3 of 13 AMI IDs live in us-east-2 (`ami-0631f0a4211880f98`, `ami-0b0605bcf9ffffa63`, plus the current latest): names, build dates, and the full Description strings (PCS Agent / Slurm / EFS Utils versions) match the table exactly. The SNS topic ARN and SSM parameter path match the official PCS docs verbatim.
- **§2.1's IMDS-based recovery is a genuinely nice touch** — no hand-copied cluster ID; and the prerequisites hold: every `add-cng*.yaml` launch template sets `InstanceMetadataTags: enabled`, and IMDS permits `:` in exposed tag keys.
- **The `-W`/`-S` switch across all raw `ldap*` operator commands** is a real hygiene improvement over the old `-w "$ADMIN_PW"` / `-s NEWPASS` forms — nothing sensitive in operator argv or history any more.
- **The README §7 accounting callout** is exactly the kind of fix that only comes from actually running the walkthrough — and its cross-reference anchor resolves correctly.
- **The GPU verify cell is well-designed** — env vars, torch's view, and an actual kernel launch, in one paste.
- **The single-user-default reframing of USER-MANAGEMENT** is the right call — most deployments never need LDAP, and the old doc buried that.
- **The IAM.md trim loses no operational content** — the Grafana port-forward flow lives in README §8.2, and `tests/iam-test.md` remains linked from the tests README.

### Sources

- **Slurm**: sbatch shebang check — https://github.com/SchedMD/slurm/blob/master/src/sbatch/sbatch.c ("This does not look like a batch script"); sreport time semantics — https://slurm.schedmd.com/sreport.html
- **AWS PCS**: PCS-Ready DLAMI guide (SNS ARN, SSM path, describe-images lookup) — https://docs.aws.amazon.com/pcs/latest/userguide/working-with_ami_pcs-ready-dlami.html; IMDS instance tags — https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/work-with-tags-in-IMDS.html; ssm-user sudo — https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-ssm-user-permissions.html
- **Jupyter**: security model / no-auth guidance — https://jupyter-server.readthedocs.io/en/latest/operators/security.html
- **Verified live, 2026-08-04**: `aws ssm get-parameter /aws/service/pcs/ami/dlami-base-ubuntu2404/x86_64/latest/ami-id` (us-east-2) → `ami-083d41c7d9d7d10a3`; `aws ec2 describe-images` on 3 table AMI IDs → Description strings match; `curl download.pytorch.org/whl/{cu124,cu126,cu128,cu129,cu130}/torch/` → all 200; JupyterLab 4.6.2 token-less + XSRF-enabled: GET /lab 200, cross-site POST 403, cookie+header POST 201
- **Repo files**: `assets/cluster.yaml` + `assets/pcs-ml-cluster-deploy-all.yaml` (`AccountingPolicyEnforcement` AllowedValues: `none`, `associations,limits,safe`); `assets/cluster-user-iam.yaml` (RunShell grant; CFN/PCS/EC2 read grants backing the JUPYTER/USER-MANAGEMENT IAM notes); `assets/scripts/ldap-add-user.sh` (authorized_keys install, argv password forms); `tests/jupyter-notebook-test.md` (expected nvidia-smi behavior, `torch 2.12.1+cu130` record, `Values=*login` remnant); `add-cng*.yaml` (`InstanceMetadataTags: enabled`)

KeitaW (changes requested):
Thank you! Few comments

DaisukeMiyamoto:
Thanks for the thorough review! All 21 comments addressed in [`491df976`](https://github.com/awslabs/awsome-distributed-ai/pull/1222/commits/491df976). Ran the full user path end-to-end on a fresh Tokyo `pcs-ml-cluster-deploy-all` deployment (DirectoryService=OpenLDAP-LoginNode + ManagedAccounting=enabled + AccountingPolicyEnforcement=associations,limits,safe): §2.1/§2.2 walkthrough, §4.1 register, §4.2 hold demo, and Part F1/F2/F3 all pass when run exactly as written.

**Blocking (4)** — all adopted:
- **F Part `sudo sacctmgr`**: F1/F3/F5 bind `S=/opt/…/sacctmgr` and use `sudo $S …`.
- **F2 shebang**: `printf '#!/bin/bash\n…\n' > /home/alice/smalljob.sh`.
- **F1 UID 10003 collision**: carol moved to **10004**; prereq line updated.
- **§4.2 hold demo**: creates myjob.sh/smalljob.sh with `printf`, adds `--partition=cpu1`, tightens `GrpTRESRunMins=cpu=60`, restores 6000. Also caught during the run: `--ntasks-per-node=8` is rejected on `cpu1` (`sinfo -N -o %c` = 4 on `c6i.2xlarge`), so `--ntasks-per-node=4` × 30 min = 120 > 60 — still trips the limit, and a one-line note points readers at `sinfo` for cluster-specific CPU counts.

**Should-fix (11)** — all adopted:
- **§4 `,limits` no-safe not deployable**: rescoped wording to what the templates offer.
- **B8 references deleted testuser2**: reordered — new B7 group add sees testuser2 still present, B9 deletes at end (numbering shifted: old B6→B9, B7→B6, B8→B7, B9→B8).
- **B8 (was B9) laptop/login-node mix + guard + ldapmodify SSH key**: three explicit steps — (1) laptop keygen, (2) login-node `install -d /home/testuser1/.ssh` + append to `authorized_keys`, (3) laptop `ssh -o ProxyCommand=…`.
- **slurm-25.11 hardcoding**: §3 note at the top of the section explains swapping to 25.05; the mutations that need an absolute path still show the concrete path (rather than hiding it behind a variable the reader would have to source).
- **cu124 → cu130**: adopted, plus reworded — driver 595.x supports current wheels, cu130 matches the DLAMI's CUDA 13.2 stack (linked to PCS-READY-DLAMI.md).
- **Verify GPU \"file a bug\" wording**: rewritten — all-visible `nvidia-smi` is expected, gate is `torch.cuda.device_count()`.
- **`--disable_check_xsrf` dropped**: adopted; XSRF left enabled with a callout on why (cross-site POST protection).
- **Multi-user tokenless warning**: reworded to \"trusted-as-the-same-principal\".
- **IAM cluster-user overstated safety**: row now explicitly names the login-node SSM shell as `ssm-user` + passwordless sudo → effectively root on the login node (shared /home, Slurm accounting, `/pcs/<id>/ldap/*` SSM secret). Follow-up (RunShell removal / Run-As) intentionally left out of this docs PR.
- **IAM Name-tag caveat**: restored — Name is operator-mutable, re-tag changes scoping, and any matching Name elsewhere in the account also satisfies the condition.
- **ldap-add-user.sh dup guard misses NSS**: adopted, added `getent passwd` (catches local `/etc/passwd` shadowing, e.g. `ubuntu` UID 1000) and a hard reject for UID outside 10001–59999.
- **PCS-READY-DLAMI describe-images \"below\"**: reframed the intro as a dated snapshot (as of 2026-08-05); the SSM parameter pointer is enough for region-portable lookup, and the PCS user guide already carries the `describe-images` recipe.

**Nit (6)** — all adopted (some inline with the above):
- **F1 WithAssoc**: added.
- **sreport end date off-by-one**: `end=2026-04-30` → `2026-05-01`, `-03-31` → `-04-01`; multi-user-test F4 `end=\$(date +%Y-%m-%d)` → `end=now`.
- **`../README.md#quick-start`**: → `#3-quick-start`.
- **`#41-worked-example…`**: → `#41-register-users-to-accounts`.
- **Tokenless \"nothing else changes\"**: explicitly notes sub-step 2 is skipped and the banner heredoc's token line comes out.
- **DLAMI table snapshot framing**: adopted (see #15 above). Not adding a 2026-07-29 row in this PR to keep it docs-only; happy to fold that into a follow-up chore with the SSM-parameter refresh.

Two takeaways for me: run every section as-is on a fresh **deploy-all** cluster (not the modular chain), and every fenced block that appears in the doc has to run — even \"trivial-looking\" ones (F2's `echo > smalljob.sh` was the one that ended up shipping without a shebang). Both are in my working notes now.

KeitaW (approved):
LGTM!

PR #1228, opened 2026-08-13
fix(hyperpod-eks-tf): prevent command injection in task_governance lo…

chijusha:
…cal-exec

The wait_for_kueue_webhook provisioner interpolated the user-controllable variables eks_cluster_name and aws_region directly into a local-exec command string. Terraform passes that string to /bin/sh -c, so shell metacharacters in the values (e.g. "name; touch /tmp/pwned; #") execute arbitrary commands on the host running terraform apply (OS command injection, CWE-78).

Fix, matching the pattern used in awslabs/security-hardened-amis-for-eks (v20260518):
- Add allowlist validation blocks to eks_cluster_name and aws_region so metacharacter payloads are rejected at plan time.
- Pass both values via the local-exec environment map and reference them as quoted shell variables, so their contents are never parsed as command syntax.

The compute_quota provisioners in this module already use the environment map pattern and are unaffected.

## Changes

-

## Test Plan

**Environment:**
- AWS Service:
- Instance type:
- Number of nodes:

**Test commands:**
```bash

```

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/1.architectures/7.sagemaker-hyperpod-eks/terraform-modules/hyperpod-eks-tf/modules/task_governance/main.tf
+++ b/1.architectures/7.sagemaker-hyperpod-eks/terraform-modules/hyperpod-eks-tf/modules/task_governance/main.tf
@@ -91,6 +91,13 @@ resource "aws_eks_addon" "task_governance" {
 
 resource "null_resource" "wait_for_kueue_webhook" {
   provisioner "local-exec" {
+    # Pass user-controllable values via the environment map instead of
+    # interpolating them into the command string. This neutralizes shell
+    # metacharacters in the variable values (command injection, CWE-78).
+    environment = {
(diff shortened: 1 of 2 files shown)


bluecrayon52 (approved):
Good, well-scoped security fix for the command injection (CWE-78) in the task_governance local-exec provisioner.

The switch from Terraform interpolation (`${var.aws_region}`) to passing values through the `environment` map and referencing them as quoted shell variables (`"$AWS_REGION"`) is the correct root-cause fix — the values are no longer parsed as shell syntax. The allowlist `validation` blocks add solid defense-in-depth by rejecting metacharacters at plan time, and this matches the existing pattern already used by the compute_quota provisioners in this module.

LGTM.

PR #1229, opened 2026-08-14
Fix broken star history chart in README

OctoBored:
The star history chart embedded in the README is broken, as it relies on a data source that is now restricted by GitHub's stargazer API limits. This change updates the chart to use a working alternative that requires no API token, so the chart renders correctly again. No other functionality is affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,4 +106,4 @@ Thanks to all the contributors for building, reviewing and testing.
 
 ## 7. Star History
 
-[![Star History Chart](https://star-history.dera.page/svg?repos=awslabs/awsome-distributed&type=Date)](https://star-history.dera.page/#awslabs/awsome-distributed&Date)
+[![Star History Chart](https://star-history.dera.page/svg?repos=awslabs/awsome-distributed-ai&type=Date)](https://star-history.dera.page/#awslabs/awsome-distributed-ai&Date)
--- a/README.md
+++ b/README.md
@@ -106,4 +106,4 @@ Thanks to all the contributors for building, reviewing and testing.
 
 ## 7. Star History
 
-[![Star History Chart](https://api.star-history.com/svg?repos=awslabs/awsome-distributed&type=Date)](https://star-history.com/#awslabs/awsome-distributed&Date)
+[![Star History Chart](https://star-history.dera.page/svg?repos=awslabs/awsome-distributed&type=Date)](https://star-history.dera.page/#awslabs/awsome-distributed&Date)


README.md
```diff
@@ -106,4 +106,4 @@ Thanks to all the contributors for building, reviewing and testing.
 
 ## 7. Star History
 
-[![Star History Chart](https://api.star-history.com/svg?repos=awslabs/awsome-distributed&type=Date)](https://star-history.com/#awslabs/awsome-distributed&Date)
+[![Star History Chart](https://star-history.dera.page/svg?repos=awslabs/awsome-distributed&type=Date)](https://star-history.dera.page/#awslabs/awsome-distributed&Date)
```
KeitaW: ### Proposed chart URL returns HTTP 404 — the repo slug is stale (blocking)

The new URL keeps the old `awslabs/awsome-distributed` slug, and the new host rejects it:

```
$ curl -sS -o /dev/null -w '%{http_code} %{content_type}\n' \
    'https://star-history.dera.page/svg?repos=awslabs/awsome-distributed&type=Date'
404 text/plain; charset=UTF-8

body: Repo not found: "awslabs/awsome-distributed" may have been moved, renamed, or deleted.
```

Stable across three consecutive requests (verified live, 2026-08-25) — not a transient.

The root cause is a rename this README never caught up with: `awslabs/awsome-distributed` 301-redirects to repository id `698678789`, which is `awslabs/awsome-distributed-ai` (`GET https://api.github.com/repos/awslabs/awsome-distributed` → `301`, verified live 2026-08-25). The old host tolerated the stale slug; this one resolves it literally and does not follow GitHub's rename redirect.

**Impact.** Merged as-is, section 7 renders a broken-image icon instead of a chart — strictly worse than the current placeholder, which at least explains itself to the reader.

**Suggestion.** Use the current repository name in both URLs. With the correct slug the same host returns a real chart — `200 image/svg+xml`, 67,387 bytes, axes 2024/2025/2026 and a 0–400+ y-axis consistent with the repo's 471 stargazers (verified live, 2026-08-25):

```suggestion
[![Star History Chart](https://star-history.dera.page/svg?repos=awslabs/awsome-distributed-ai&type=Date)](https://star-history.dera.page/#awslabs/awsome-distributed-ai&Date)
```

One note on how I checked the second half of that line: the `#…` fragment of the interactive link is never sent to the server, so a `200` on `https://star-history.dera.page/` says nothing about whether the chart resolves — it is the client that looks the slug up. I did not observe that link failing, but it carries the same stale slug the `/svg` endpoint 404s on, which is why the suggestion updates both halves rather than just the image.

KeitaW (commented):
## Review Batch 1/1 — README Asset URLs & Host Provenance

Thanks for chasing this down — I want to start by confirming your diagnosis, because it is correct and it is not obvious from an HTTP status alone. The current chart really is broken: `https://api.star-history.com/svg?repos=…&type=Date` returns **HTTP 200 `image/svg+xml`**, so every link checker passes it, but the SVG body is a placeholder whose text nodes read *"GitHub restricted access to star data"* / *"GitHub team is looking into it"* (verified live, 2026-08-25). So the README does render a dead chart today and this is worth fixing.

The problem is that the replacement URL as written doesn't fix it — it swaps a placeholder image for a broken-image icon. One-line fix in the inline comment on `README.md:109`, with a one-click `suggestion` block.

### Which host should an AWS-owned README point at? (a maintainer call, not a defect)

`star-history.dera.page` is a different domain from the official `star-history.com`, and this line makes it both a remote image source and a reader-facing link from an `awslabs` README. That's a maintainer decision rather than a code defect, so I'm flagging it rather than asking for a change — but it is worth a deliberate answer before merge, since the repo would be handing a third party control over a rendered asset on its front page.

For what it's worth, the placeholder SVG currently being served names an official path back: <https://star-history.com/blog/github-stargazer-api-restriction> describes a workaround repo owners can apply to restore their own chart on the original host. If that works for this repo, it keeps the README on the domain it already trusted. If maintainers prefer this fork, the inline suggestion is what makes it actually render.

### Pre-existing, non-blocking: the contributors line carries the same stale slug

`README.md:105` uses `awslabs/awsome-distributed` too. I checked it rather than assuming, and it works — `https://contrib.rocks/image?repo=awslabs/awsome-distributed` returns `200 image/svg+xml`, 365,443 bytes (verified live, 2026-08-25), because that service follows the GitHub rename redirect. So there is nothing to fix today. Mentioning it only because it is the same latent staleness, and canonicalizing both slugs in one pass would remove the class — your call whether that belongs in this focused PR.

### Things That Look Great

- **The diagnosis is right, and it was not cheap to reach.** A status-code check says the old URL is fine; only reading the SVG body reveals the "GitHub restricted access to star data" placeholder. The PR description names the cause (GitHub's stargazer API restriction) accurately.
- **Well-scoped.** One file, `+1/-1`, no drive-by edits — exactly the shape a README fix should have, and it made this straightforward to verify end-to-end.
- **The markdown itself is correct**, including the unescaped `&` in the query string, which renders fine in GitHub-flavored Markdown.
- **The replacement host does work.** With the corrected slug it serves a real chart with the full history — so the approach is sound and only the slug stands between this and a working chart.

### Sources

Live verification (all 2026-08-25, `curl -sS -L`):

- `https://api.star-history.com/svg?repos=awslabs/awsome-distributed&type=Date` → `200 image/svg+xml`, 60,125 B; body text: "GitHub restricted access to star data". Identical placeholder for the `-ai` slug.
- `https://star-history.dera.page/svg?repos=awslabs/awsome-distributed&type=Date` → `404 text/plain`, "Repo not found" (3/3 attempts).
- `https://star-history.dera.page/svg?repos=awslabs/awsome-distributed-ai&type=Date` → `200 image/svg+xml`, 67,387 B, chart with 2024–2026 axis.
- `https://contrib.rocks/image?repo=awslabs/awsome-distributed` → `200 image/svg+xml`, 365,443 B.
- `https://api.github.com/repos/awslabs/awsome-distributed` → `301` → repository `698678789` = `awslabs/awsome-distributed-ai`, 471 stars.

Docs:

- GitHub REST — following redirects after a repository rename: <https://docs.github.com/rest/guides/best-practices-for-using-the-rest-api#follow-redirects>
- star-history.com on the stargazer API restriction and the owner workaround: <https://star-history.com/blog/github-stargazer-api-restriction>

KeitaW (approved):

PR #1233, opened 2026-08-20
fix(cosmos3): parameterize FSx PVC name and correct eval checkpoint default

bluecrayon52:
These are two first-use defects in the `3.test_cases/pytorch/cosmos3/` sample: manifest/env-config only, no Python changes.

## Defect 1 — FSx PVC name / mutually-exclusive storage manifests

`storage/storage-fsx-efa-sc.yaml` (dynamic EFA provisioning of a **new** filesystem) and `storage/storage-fsx-dra.yaml` (static bind to an **existing** filesystem, hydrated from S3 via a DRA) are two answers to the same question — how the pods get an RWX PVC at `/fsx` — but the READMEs told you to apply **both**, and two workload manifests hardcoded the claim name. Applying both provisions a 9.6 TiB filesystem that nothing mounts; skipping the "optional" DRA line leaves every workload with an unbound PVC.

- **`storage-fsx-efa-sc.yaml`**: PVC `name: fsx-efa-claim` → `${FSX_CLAIM}` (this was the only hardcoded PVC name in the repo). Now either storage manifest produces the PVC the workloads expect, and `FSX_CLAIM` is the single knob.
- **`kubernetes/` and `hyperpod-eks/` `generate-vllm-omni-super.yaml`**: `claimName: fsx-claim` → `${FSX_CLAIM}`, matching the other six workload manifests.
- **READMEs** (both sub-paths + top-level): the `Storage (once)` step is now **"Choose one"** — Option A `storage-fsx-dra.yaml` (recommended, S3→FSx DRA against an existing filesystem), Option B `storage-fsx-efa-sc.yaml` (dynamically provision a new EFA filesystem) — and no longer describes the two as composed. Prose refers to `$FSX_CLAIM` instead of a hardcoded `fsx-claim`.

The two storage manifests are genuine alternatives: the EFA StorageClass is a dynamic-provisioning option, while the DRA is the standard production data plane (hydrate datasets from S3, export checkpoints back to S3).

## Defect 2 — `EVAL_CHECKPOINT` default cannot work

Both `serve-policy.yaml` manifests invoke `action_policy_server_libero`, which takes a **local checkpoint directory** and has no Hugging Face resolution, but the shipped default was a bare HF repo id (`nvidia/Cosmos3-Nano-Policy-DROID`) and the manifest comments documented that id as a supported option.

- **`env_vars.example`**: default `EVAL_CHECKPOINT` to an FSx checkpoint path (`/fsx/results/run/checkpoints/iter_00000010/model`) and document that the LIBERO server needs a local dir (download the public policy to FSx first to eval it).
- **`serve-policy.yaml`** (both paths): corrected the option comments to match the invoked server — local dir only; the bare `nvidia/...` id is not accepted.

Kept the LIBERO server (lower blast radius); in practice this variable points at an exported checkpoint on the FSx mount, so the default now makes that expectation obvious.

## Defect 3 — stale claim references in docs

Fixed prose/comments that hardcoded `fsx-claim` (both sub-READMEs’ prerequisite lines, `kubernetes/train-vision-sft.yaml` header example) to reference `$FSX_CLAIM`, so a reader who changes the variable is not misled.

## Verification

**Render checks.** Every manifest renders with no unresolved `${...}` (only the intentional `$$`-escaped in-container runtime literals survive), all eight workload `claimName`s and both storage PVC `name`s resolve to a single `$FSX_CLAIM` value, and every rendered manifest parses as valid YAML.

**Cluster launch check (EKS, `ap-southeast-3`).** Both storage options were validated against a live cluster with `kubectl apply --dry-run=server` (server-side schema + admission validation; nothing persisted, so no FSx filesystem is provisioned and no resources are left behind). Rendered with a distinctive `FSX_CLAIM=natharno-pr1233-fsxtest`:
- **Option A (`storage-fsx-dra.yaml`)** — both objects accepted: `persistentvolume/fsx-dra-pv created (server dry run)` and `persistentvolumeclaim/natharno-pr1233-fsxtest created (server dry run)`. The PVC name tracks `$FSX_CLAIM` as intended.
- **Option B (`storage-fsx-efa-sc.yaml`)** — `persistentvolumeclaim/natharno-pr1233-fsxtest created (server dry run)`, again taking the `$FSX_CLAIM` name. (The StorageClass reported `parameters: Forbidden: updates to parameters are forbidden` only because an identical `fsx-efa-p2-sc` already exists on that shared cluster and StorageClass parameters are immutable — unrelated to this change; the PVC validates cleanly on its own.)

Not exercised: real PVC binding / FSx provisioning and the GPU workloads, to avoid provisioning cost on a shared cluster — the dry-run confirms the manifests launch and the PVC naming is correct, which is what this PR changes.

Out of scope (per handover): the `cosmos-framework` re-pin, `norm_monitor_guard.py` deletion, `COSMOS3_PEAK_TFLOPS`, and the `$$VAR` runtime literals.

### DIFF
--- a/3.test_cases/pytorch/cosmos3/README.md
+++ b/3.test_cases/pytorch/cosmos3/README.md
@@ -127,8 +127,8 @@ emits the same metrics to Amazon Managed Prometheus.
 | `kubernetes/train-vision-sft.yaml` | **Vision SFT** (Nano) + Cosmos3-Super LoRA (parameterized `SFT_TOML`+`DATASET_PATH`). |
 | `kubernetes/generate-vllm-omni-super.yaml` | **Generation (SDG)** — Super video-to-video (V2V) via the official `vllm/vllm-omni:cosmos3` engine (a separate stack from the training image). |
 | `kubernetes/serve-policy.yaml` | **Policy-server eval** (Deployment + Service). |
-| `storage/storage-fsx-efa-sc.yaml` | EFA-enabled PERSISTENT_2 FSx for Lustre StorageClass + PersistentVolumeClaim (PVC). Shared by both deployment paths. |
-| `storage/storage-fsx-dra.yaml` | **S3→FSx-Lustre Data Repository Association (DRA)** hydration — the recommended data plane (in-region S3 origin → FSx hot tier). Shared by both deployment paths. |
+| `storage/storage-fsx-dra.yaml` | **S3→FSx-Lustre Data Repository Association (DRA)** hydration against an **existing** filesystem — the recommended data plane (in-region S3 origin → FSx hot tier). One of two storage options; see `storage/` (apply one, not both). |
+| `storage/storage-fsx-efa-sc.yaml` | **Alternative** storage option: dynamically provision a **new** EFA-enabled PERSISTENT_2 FSx for Lustre StorageClass + PersistentVolumeClaim (PVC). |
 | `build-push.sh` | Build the **AWS DLC image** (`Dockerfile`) and push to ECR via `docker buildx` (see **Deployment**). |
 
 > **Training and inference take different CLIs — don't conflate them.** Training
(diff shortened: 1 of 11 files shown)


nicolasj92 (approved):
LGTM

PR #1234, opened 2026-08-21
Add DeepEP V2 benchmark (NCCL GIN)

Xuan-1998:
DeepEP V2 replaces the NVSHMEM transport with NCCL GIN (GPU-Initiated Networking).

Two ways to consume it, mirroring the existing deepep-benchmark (v1):
- setup_deepep_gin.sh: run inside your own container. Uninstalls any existing deep_ep, validates the NCCL is GIN-capable (nccl_device.h), and builds + installs the EFA/GIN DeepEP against your torch.
- deepep.Dockerfile: reference image for those without a stack. Builds EFA userspace, a GIN-capable NCCL, aws-ofi-nccl with --enable-gdaki, and torch, then calls the same script.

Slurm launchers run tests/elastic/test_ep.py on 1 node (NVLink) and 2 nodes (EFA-GDA). The sbatch files pin NCCL_GIN_TYPE=5: with it unset, NCCL selects the CPU-proxy GIN backend.

PR #1238, opened 2026-08-24
Replace EP backend comparison with Decode-like and Prefill-like B200 benchmarks

KeitaW:
## Purpose

Relates to #1234.

Replace the previous backend-native EP result tables with a common-boundary B200 microbenchmark. The retired B300, B200, H100, and Decode-only result files are removed because their latency and bandwidth fields did not share one timing boundary or byte numerator.

This microbenchmark covers synthetic Decode-like and Prefill-like expert-parallel communication.

## What changed

- Add separate Decode-like and Prefill-like profiles. Decode-like uses 128 tokens/rank and the low-latency API. Prefill-like uses 4,096 tokens/rank and the normal high-throughput API, including required layout work.
- Run UCCL, DeepEP V1 NVSHMEM, and DeepEP V2 NCCL GIN with the same deterministic BF16 input, exact top-k route and weights, model shape, named nodes, and runtime stack.
- Measure one external CUDA event boundary from dispatch-input readiness through required layout, dispatch, and combine completion. Host-side FP8 conversion runs once before the boundary; low-latency kernels that quantize internally do so inside it. Each iteration reports the slowest rank.
- Use slowest-rank latency as the primary metric for both profiles. Aggregate input tokens/s is the cross-EP-size metric; logical throughput columns are secondary, and the scale-out column is comparable only within one EP size.
- Rotate backend, profile, and dtype order across independent process starts per cell (default 20). Each start contains 20 warmup iterations and 100 measured iterations per dtype.
- Record per-arm runtime provenance: the loaded NCCL library version, the resolved communication-kernel SM count, and the detected RDMA GB/s. `EP_NUM_SMS` pins the SM count.
- Support variable node counts and EP sizes (`EP_WORLD_SIZES`; the summarizer takes `--world-sizes` or derives them from the logs), so a 2-node EP16 campaign summarizes and plots.
- Reject incomplete matrices, correctness failures, mutable image tags, route or input mismatches, runtime mismatches, payload mismatches, and derived metrics that disagree with latency and logical-byte accounting.
- Coordinate through the shared Lease, reject selected/protected node overlap, retain every log and rendered object, and require verified teardown before writing `CAMPAIGN_COMPLETE`.
- Replace the retired result set with a human-readable report and machine-readable summaries. Update the SGLang documentation that referenced the retired tables.

## Review response

`df6d2943` addresses [whn09's review](https://github.com/awslabs/awsome-distributed-ai/pull/1238#issuecomment-5421597411): the FP8 cast is hoisted out of the timed region, the start count is raised so the bootstrap interval carries information, slowest-rank latency is the primary metric for both profiles, `EP_BUFFER_DEBUG` is off on scored runs, the summarizer and launcher accept variable node counts and EP sizes, and every result records the loaded NCCL version and the resolved SM count. `40990df2` retries node admission while a prior case's pod is still terminating. `d35c861e` replaces the tables with measurements taken on that harness.

## B200 results

Two campaigns on 4 `p6-b200.48xlarge` nodes in `us-east-1c`. The primary campaign is EP16 with 16 independent starts; the companion campaign adds EP32 at 4 starts for the EP-size axis, where the intervals are weak.

Slowest-rank dispatch-plus-combine latency, medians across starts:

| EP size | Dispatch dtype | Starts | UCCL | DeepEP V1 NVSHMEM | DeepEP V2 NCCL GIN |
|---:|:---:|---:|---:|---:|---:|
| 16 ranks decode | FP8 | 16 | 0.5258 ms | 1.0264 ms | **0.4712 ms** |
| 16 ranks decode | BF16 | 16 | 0.5986 ms | 1.0358 ms | **0.4839 ms** |
| 16 ranks prefill | FP8 | 16 | 5.2691 ms | 2.9742 ms | **2.6578 ms** |
| 16 ranks prefill | BF16 | 16 | 5.5088 ms | 3.5448 ms | **3.3709 ms** |
| 32 ranks decode | FP8 | 4 | **0.7758 ms** | 1.5568 ms | 0.8361 ms |
| 32 ranks decode | BF16 | 4 | **0.8718 ms** | 1.5598 ms | 0.9423 ms |
| 32 ranks prefill | FP8 | 4 | 8.0894 ms | 12.7493 ms | **6.6226 ms** |
| 32 ranks prefill | BF16 | 4 | 9.8898 ms | 13.1769 ms | **8.7098 ms** |

Box plots comparing the three backend arms for every EP16 Decode-like and Prefill-like workload cell

DeepEP V2 has the lowest latency in all four EP16 cells, by about 10% (FP8) and 19% (BF16) over UCCL for decode and 39% to 50% for prefill. All eight paired directions are supported. At EP32 the decode ranking differs: UCCL leads both decode cells while DeepEP V2 leads both prefill cells, so an EP16 decode result does not carry to EP32. Every arm is faster with FP8 than with BF16; the FP8-slower-than-BF16 decode pattern in the retired tables does not appear here.

The recorded SM counts show the prefill arms are not SM-matched: DeepEP V2 resolves 64 SMs from its bandwidth heuristic while the normal-mode UCCL and DeepEP V1 buffers report 20. Part of the prefill margin may follow from that rather than from the transport. `EP_NUM_SMS` now pins the count for a matched comparison.

Per-start values, bootstrap intervals, paired improvements, provenance, and scope limits are in [`RESULTS.md`](micro-benchmarks/expert-parallelism/ep-backend-comparison/RESULTS.md). Machine-readable summaries: [`ep16-n16`](micro-benchmarks/expert-parallelism/ep-backend-comparison/results/b200-us-east-1-2026-09-01-ep16-n16.json) SHA-256 `60d2917ad4fb1f4662b4c522c740927d1403ba6a6f23c762937562e524a6b7c1` and [`ep16-ep32-n4`](micro-benchmarks/expert-parallelism/ep-backend-comparison/results/b200-us-east-1-2026-09-01-ep16-ep32-n4.json) SHA-256 `f4edd9a172d9b93e0903fbb3d43b499bf17aaee62e4dd3db3d212bab440173ff`. The ap-south-1 summary is kept as a dated historical artifact.

## Qualification and custody

- EKS cluster `ml-clusters-shared-us-east-1` in `us-east-1c`, 4 named `p6-b200.48xlarge` nodes, each with 8 NVIDIA B200 GPUs and 8 allocatable EFA devices.
- EP16 used 2 nodes and 16 GPU ranks. EP32 used 4 nodes and 32 GPU ranks.
- PyTorch `2.13.0+cu130` and CUDA `13.0` across all scored records. Every arm loaded NCCL `2.31.2`; the PyTorch build constant reports `2.29.7`, and both are recorded.
- 3 backend images pinned by SHA-256 digest, the same images used for the retired ap-south-1 campaign. The DeepEP and UCCL source commits inside those images are not recorded.
- Primary campaign: 4 admission cases, 96 scored distributed starts, 192 scored records, 0 correctness failures, 1,124 checksum entries. Companion campaign: 4 admission cases, 48 scored distributed starts, 96 scored records, 0 correctness failures, 740 checksum entries.
- Exclusive shared-Lease ownership for each campaign. Every owned Pod, StatefulSet, and Service was removed and the Lease released.
- Both campaigns report a teardown failure: the owned namespace stays `Terminating` because two stale `visibility.kueue.x-k8s.io` APIServices in this cluster fail namespace deletion discovery. That is a cluster property, not a campaign property, and no scored record depends on it.
- The pinned UCCL normal-mode worker uses synchronized process-lifetime cleanup after result output because explicit proxy destruction invalidates the CUDA context. This happens after all timed iterations and does not change the timing boundary.

## Validation

- Real B200 EKS campaigns: all admission cases and all 144 scored distributed starts passed across the two campaigns.
- The two campaigns' shared EP16 cells agree to within 1.3% in every arm and cell except BF16 decode for DeepEP V2 (2.5%), which is the cell with the highest run-to-run variability in both.
- Local unit suite: 13 tests passed and 8 CUDA-dependent tests skipped because local PyTorch is unavailable.
- `ruff check`, `ruff format --check`, `py_compile`, `bash -n`, `shellcheck`, and `git diff --check`: passed.

## Final directory

```text
micro-benchmarks/expert-parallelism/ep-backend-comparison/
├── README.md
├── RESULTS.md
├── ep_benchmark.py
├── extract_results.py
├── plot_results.py
├── result_io.py
├── results/
│   ├── b200-ap-south-1-2026-08-25-boxplots.png
│   ├── b200-ap-south-1-2026-08-25.json
│   ├── b200-us-east-1-2026-09-01-ep16-ep32-n4-boxplots.png
│   ├── b200-us-east-1-2026-09-01-ep16-ep32-n4.json
│   ├── b200-us-east-1-2026-09-01-ep16-n16-boxplots.png
│   └── b200-us-east-1-2026-09-01-ep16-n16.json
├── run_ep_comparison.sh
├── run_ep_rank.sh
├── summarize_results.py
├── test_ep_benchmark.py
└── test_summarize_results.py
```


micro-benchmarks/expert-parallelism/ep-backend-comparison/RESULTS-b200.md
```diff
@@ -0,0 +1,135 @@
...
+
+Before measurement, an INFO-level EP16 admission ran on both nodes with `NCCL_GIN_TYPE=5` and `NCCL_SYM_GIN_KERNELS_ENABLE=0`. Logs showed the aws-ofi-nccl `Libfabric_GDAKI` version 14 dimensionless plugin, GPU Direct RDMA enabled on EFA HCAs, nonzero GIN layouts, and balanced cluster EFA TX and RX deltas of 235,008 bytes each. Every scored cell also recorded nonzero EFA RDMA writes, so the cross-node path was measured rather than inferred from topology.
+
+## Decode latency comparison
+
+Lower latency is better. UCCL and DeepEP V1 print an aggregate dispatch-plus-combine latency. DeepEP V2 prints dispatch and combine separately, so the V2 value below is their sum for the same dispatch dtype.
+
+| EP ranks | Dispatch dtype | UCCL latency | DeepEP V1 latency | DeepEP V2 latency | V2 vs UCCL | V2 vs V1 |
+|---:|---|---:|---:|---:|---:|---:|
+| 16 ranks | FP8 | 475.540 us | 881.700 us | **384.994 us** | 19.04 percent lower | 56.34 percent lower |
```
Xuan-1998: What does latency here mean? why the latency here is lower but the bandwidth below is also lower? In DeepEPv2's benchmark bandwidth = (number of bytes) / latency

KeitaW: Update: I removed the previous backend-native tables and reran the comparison with separate Decode-like and Prefill-like profiles.

Latency now means the slowest-rank CUDA elapsed time for one dispatch followed by one combine. The boundary starts with the BF16 input and route ready, includes required FP8 conversion, and ends after combine completion. Prefill-like also includes the dispatch layout required by the normal UCCL and DeepEP V1 APIs.

Cross-backend throughput is now calculated as common logical payload bytes divided by that same elapsed time. For a fixed profile, EP size, and dtype, every backend has the same byte numerator, so lower latency necessarily corresponds to higher logical GB/s/rank. FP8 and BF16 have different useful-byte numerators and remain separate cells. Backend-native SO, SU, and aggregate bandwidth values are no longer presented as interchangeable metrics.

The replacement B200 run contains 3 independent process starts per cell, 20 warmup iterations and 100 measured iterations per dtype/start, and 72 scored records. All records passed correctness and matrix validation; the maximum run-to-run CV is 3.31 percent. The updated method, Decode-like and Prefill-like tables, provenance, and scope limits are in the [replacement report](https://github.com/awslabs/awsome-distributed-ai/blob/codex/ep-backend-comparison-deepep-v2-b200/micro-benchmarks/expert-parallelism/ep-backend-comparison/RESULTS.md).

whn09 (approved):
LGTM

Issue #13, opened 2023-10-06
Benchmark stable diffusion training throughput with SDPA/Flash_Attention


awsankur:
https://huggingface.co/docs/diffusers/optimization/torch2.0

awsankur:
[HuggingFace Diffusers](https://github.com/huggingface/diffusers/tree/main) has a `set_attn_processor` method that you can use to plug and play different attention processors. A list of attention processors can be found [here](https://github.com/huggingface/diffusers/blob/main/src/diffusers/models/attention_processor.py) 

AttnProcessor2_0 which is a Processor for implementing scaled dot-product attention is enabled by default if you're using PyTorch 2.0.

More details on this can be found here: https://pytorch.org/blog/accelerated-diffusers-pt-20/

PR #1239, opened 2026-08-25
fix(deepep-v2): use the EFA installer bundled aws-ofi-nccl plugin

Xuan-1998:
Installer 1.50.0 bundles an EFA-GDA-capable aws-ofi-nccl (verified in PR 1234 review), so drop the source build and apply the ncclGinPlugin_v14 gate to the bundled binary instead. Fold the libfabric and aws-ofi-nccl version bullets into the EFA installer bullet, and state explicitly that the installer must run on the node itself (not the container) to provide the kernel driver.

Addresses post-merge review comments on PR 1234.


amitrad-aws (approved):

KeitaW (commented):
## Review Batch 1/3 — What the container still needs beyond the installer

I validated this change end to end rather than on reading alone: I built the image from this branch at `de11533`, built a second image with `--build-arg EFA_INSTALLER_VERSION=1.49.0` to exercise the fail-closed path, and ran the two-node internode benchmark on 2x `p6-b200.48xlarge`. **The core substitution is correct and I could not break it** — details and numbers in batch 3.

The two findings here are both in prose this PR newly wrote, and they point the same way: the image needs more than the installer, and the docs now read as though it does not.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
```diff
@@ -79,7 +79,7 @@ it; left unset, NCCL selects the CPU-proxy backend.
 
 The script refuses to run if the NCCL at `--nccl-root` lacks the GIN device API, so it cannot
 silently produce a DeepEP that falls back to a slower path. Runtime still requires the
-aws-ofi-nccl GIN plugin built with EFA-GDA support.
+EFA-GDA-capable aws-ofi-nccl GIN plugin (bundled with EFA installer >= 1.50).
```
KeitaW: ### The custom-container path needs its own installer run, and the new wording points away from that

**Observation**: This section covers bringing your own container, and now tells the reader runtime "requires the EFA-GDA-capable aws-ofi-nccl GIN plugin (bundled with EFA installer >= 1.50)" — but never says how that plugin gets *into* their container. Meanwhile the other sentence this PR added (lines 37-40) says to run the installer "**on the node itself** (not in the container)". Read together, the natural conclusion is that the host installer run is what supplies the plugin.

It does not, and my end-to-end run is a clean demonstration of the split. The `p6-b200` nodes I ran on had `efa 3.3.0` and `rdma-core` installed on the host and **no libfabric and no `libnccl-ofi` at all**; the plugin and libfabric came entirely from the installer run *inside* the image (`deepep.Dockerfile:103`, with `--skip-kmod`). Two installer runs with different flags, and the README now names only one of them (verified live, 2026-08-25).

**Impact**: Someone following the "use the script in your own container" flow — which the README presents as a first-class path — installs on the host, gets a container with no plugin, and hits a GIN-unavailable failure a long way from its cause. `setup_deepep_gin.sh:253` reinforces the gap: its completion notes still print the bare placeholder `export NCCL_NET_PLUGIN=/path/to/libnccl-net-ofi.so`.

**Suggestion**: One added sentence here closes it:

```suggestion
EFA-GDA-capable aws-ofi-nccl GIN plugin (bundled with EFA installer >= 1.50). Note that the
plugin has to be *in your container*: run the EFA installer inside the image with
`--skip-kmod` (as `deepep.Dockerfile` does) to get the userspace stack and the plugin. The
host installer run covers only the kernel driver.
```

KeitaW (commented):
## Review Batch 2/3 — Build gate ergonomics

Four small things in the block this PR already touches. None affects a successful default build.

### Two smaller notes

- **`ENV EFA_PREFIX` lost its only in-build consumer.** `deepep.Dockerfile:108` sets `EFA_PREFIX=/opt/amazon/efa`; its only user in the build was `--with-libfabric=${EFA_PREFIX}` on the `configure` line this PR deletes. It is still reasonable to bake into the image env for anyone compiling against libfabric inside the container — but if it stays, it is now an image contract rather than a build variable, and a short comment saying so would keep the next reader from treating it as dead.
- **The gate proves the GIN ABI but not the Net ABI.** The README invites overriding `NCCL_REF` and `EFA_INSTALLER_VERSION` independently (lines 93-99), while the gate asserts only `ncclGinPlugin_v14`. The default pair is fine — I checked that NCCL `v2.31.2-1` probes GIN `{14, 13}` and Net `{12...6}`, and the bundled plugin exports `ncclGinPlugin_v11/v13/v14` plus `ncclNetPlugin_v6...v12`. But a future NCCL bump that moves the Net floor to v13 would sail through this gate. Adding `ncclNetPlugin_v12` to the same `grep`, or noting the two versions as a tested pair rather than free knobs, would cover it.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -128,34 +117,15 @@ RUN git clone ${NCCL_REPO} /opt/nccl \
...
-       else \
-         echo "ERROR: no ncclGinPlugin_v14 export -- configure did not detect EFA-GDA (libfabric >= 2.5 with hw counters + CUDA required)" >&2; exit 1; \
-       fi \
-    && echo "${OFI_HOME}/lib" > /etc/ld.so.conf.d/aws-ofi-nccl.conf && ldconfig \
-    && rm -rf /tmp/aws-ofi-nccl
+## aws-ofi-nccl: the EFA installer (>= 1.50) bundles the plugin with EFA-GDA
+## support. The ncclGinPlugin_v14 export check makes that authoritative (v11/v13
+## op-tables are exported by every build; the v14 table is EFA-GDA-only).
+ENV OFI_HOME=/opt/amazon/ofi-nccl
+RUN if nm -D "${OFI_HOME}/lib/libnccl-net-ofi.so" 2>/dev/null | grep -qw ncclGinPlugin_v14; then \
```
KeitaW: ### The `ncclGinPlugin_v14` gate now runs after the NCCL compile, so a too-old installer fails late

**Observation**: The check this replaces lived inside the EFA-installer `RUN` layer (the `fi_info`/libfabric comparison at the old lines 113-115), so a too-old installer was rejected while that layer was still running. The new gate is its own layer at line 124, which Docker reaches only after `make src.build` at lines 112-117. I confirmed the consequence by building this branch with `--build-arg EFA_INSTALLER_VERSION=1.49.0`: the build ran the whole NCCL compile (`#13 DONE 161.7s` on a 192-core builder) and only then printed the gate's error and exited 1. Bumping the installer version or swapping `EFA_INSTALLER_TARBALL` invalidates the installer layer and everything below it, so this is the normal path for that change, not an edge case (verified live, 2026-08-25).

**Impact**: Fail-closed behaviour is fully preserved — this is only wasted time, and only when the build is going to fail. But 161.7 s is the cost on a very large builder; a typical CI runner is several times that, all spent before a check that was answerable immediately after `ldconfig`.

**Suggestion**: The gate has no dependency on the NCCL layer, so appending it to the installer `RUN` at lines 88-105 (right after `ldconfig`) restores the fail-fast property the libfabric check had, with no cache cost.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -16,9 +16,9 @@ ARG TORCH_VERSION=2.11.0
 ARG TORCH_CUDA_INDEX=cu130
 ARG GDRCOPY_VERSION=v2.5.2
 
-# EFA installer: provides OpenMPI, the EFA userspace runtime, and libfabric.
-# EFA-GDA requires libfabric >= 2.5, so installer >= 1.50 is required (1.49.0 and
-# older ship libfabric 2.4; the build fails loudly on a too-old installer).
+# EFA installer: provides the complete EFA userspace stack (OpenMPI, the EFA
+# runtime, and the aws-ofi-nccl plugin). EFA-GDA requires installer >= 1.50;
```
KeitaW, line 20: ### The image no longer records which aws-ofi-nccl version it ships

**Observation**: The removed `ARG AWS_OFI_NCCL_REF=v1.21.1` was the one place a reader could see the plugin version the image was built and tested against. To be clear, I am not suggesting restoring the source build — delegating the version to a pinned installer is the point of this PR, and the build stays fully deterministic because `EFA_INSTALLER_VERSION` is pinned. It is only that the version is now implicit in that pin.

**Impact**: Legibility, not reproducibility. Someone chasing a plugin-side regression has to unpack the installer tarball to learn which aws-ofi-nccl they are running.

**Suggestion**: One clause keeps the information where it used to be:

```suggestion
# runtime, and the aws-ofi-nccl plugin -- aws-ofi-nccl 1.21.1 as bundled by
# installer 1.50.0). EFA-GDA requires installer >= 1.50;
```

KeitaW (commented):
## Review Batch 3/3 — Validation, positives & sources

### End-to-end validation of this branch

I ran the flow `slurm/test-internode.sbatch` defines, with its environment replicated exactly (`NCCL_GIN_TYPE=5`, `NCCL_SYM_GIN_KERNELS_ENABLE=0`, `EP_NCCL_ROOT_DIR=$NCCL_HOME`, `LD_PRELOAD=$NCCL_HOME/lib/libnccl.so.2`) plus `NCCL_DEBUG=INFO`, on 2x `p6-b200.48xlarge` (8x B200 + 8x EFA per node, EFA kernel driver 3.3.0g, `gdrdrv` loaded), image pinned by digest:

```
Ranks: 2 x 8 | Experts: 8/256 | Tokens: 8192, hidden: 7168 | #SM: 12, #QPs: 11/11
dispatch         82 GB/s (SO) / 265-268 GB/s (SU)
combine          68-70 GB/s (SO) / 220-230 GB/s (SU)
reduced combine  57-63 GB/s (SO) / 186-203 GB/s (SU)
RUN_EXIT=0 on both nodes; no Traceback / AssertionError / CUDA error on any rank.
```

(B200 numbers — not comparable with the B300 figures in the #1234 thread. Different silicon, not a regression.)

The part worth confirming for this change is *which* object served that run. On all 16 ranks across both nodes:

```
NCCL INFO GIN/Plugin: Successfully loaded external plugin /opt/amazon/ofi-nccl/lib/libnccl-net-ofi.so
NCCL INFO GIN/Plugin: Loaded gin plugin Libfabric_GDAKI (v14)
NCCL INFO GIN/Plugin: Skipping plugin Libfabric index 3 type 2: NCCL_GIN_TYPE=5 requested
NCCL INFO NET/OFI Initializing aws-ofi-nccl 1.21.1
NCCL INFO NET/OFI Plugin selected platform: AWS
NCCL INFO NET/Plugin: Loaded net plugin Libfabric (v12)
```

The installer's copy is the one loaded, the EFA-GDA op-table is the one bound, and the CPU-proxy backend is explicitly skipped. EFA hardware counters moved on **all 8** devices per node (~26.3 GB tx and ~26.3 GB rx each, ~210 GB per node), so the traffic really went over the fabric.

Supporting checks:

- The in-container `libnccl-net-ofi.so` hashes to `4f5aaedbf62db3d9229d1107812813bfbf6197dbc3650ea00fbd249a444c001c`, identical to the payload of `libnccl-ofi_1.21.1-1_amd64.deb` extracted straight from the installer tarball — the image runs the installer's binary, unmodified.
- The gate genuinely discriminates: installer 1.50.0 bundles plugin 1.21.1, which exports `ncclGinPlugin_v14`; 1.49.0 bundles 1.20.0, which exports only `v11`/`v13`. The 1.49.0 build fails at the gate, the 1.50.0 build passes.
- `--disable-ngc` does not skip the plugin — it only disables NGC auto-detection (`efa_installer.sh:773-779`), so the regular `libnccl-ofi` package installs. The build log shows `Setting up libnccl-ofi (1.21.1-1)`.
- Architecture-portable: the aarch64 deb installs to the same `/opt/amazon/ofi-nccl/lib/libnccl-net-ofi.so` and also exports `ncclGinPlugin_v14`.

### Things That Look Great

- **This closes the exit condition the previous revision wrote for itself.** #1234's README said the source build could be dropped "once an installer ships a GDAKI-enabled plugin"; 1.50.0 ships it and this PR acts on that instead of letting the sentence go stale.
- **It removes a real duplicate-plugin hazard, not just a build step.** Before this change the image contained *two* files named `libnccl-net-ofi.so` — the installer's at `/opt/amazon/ofi-nccl/lib` and the source build's at `/opt/aws-ofi-nccl/lib` — with both directories registered in `ld.so.conf.d`. Anything resolving by bare filename (the tuner plugin, for one) took whichever `ldconfig` ordered first. After this change a filesystem-wide search returns exactly one plugin directory. That is a correctness improvement on top of the simplification.
- **Dropping the source build also drops a build-time `git clone --recursive` of an external repo**, which is both a supply-chain surface and a large chunk of build time.
- **The discriminator choice is right and I could not fool it.** `ncclGinPlugin_v11`/`v13` are exported by proxy-only builds too; only `v14` is EFA-GDA-specific.
- **Both halves of the plugin variable pair are set** (lines 130-131), so the "use the image directly" flow is correctly configured now, not just the launcher flow.
- **`NET/OFI Plugin selected platform: AWS`** confirms the bundled build carries the platform-AWS support the deleted `--enable-platform-aws` was there to get — worth knowing, since that flag disappearing is the kind of thing that silently changes behaviour.

### Suggested priority

Nothing here blocks the substitution, which I consider proven. The two documentation items in batch 1 are the ones I would fix before merge — both are in text this PR newly wrote, and both under-state what the container needs. Batch 2 is optional cleanup in the block you are already touching.

### Sources

- Installer contents and plugin symbol tables — `aws-efa-installer-1.50.0.tar.gz` (`DEBS/UBUNTU2204/{x86_64,aarch64}/libnccl-ofi_1.21.1-1_*.deb`) and `aws-efa-installer-1.49.0.tar.gz` (`libnccl-ofi_1.20.0-1_amd64.deb`); read with `nm -D`, `readelf -d`, `strings`. Verified live, 2026-08-25.
- NCCL plugin ABI — [`src/plugin/gin.cc`](https://github.com/NVIDIA/nccl/blob/v2.31.2-1/src/plugin/gin.cc) and [`src/plugin/net.cc`](https://github.com/NVIDIA/nccl/blob/v2.31.2-1/src/plugin/net.cc) at the pinned `v2.31.2-1`. Verified live, 2026-08-25.
- EFA installer changelog (1.50.0: libfabric 2.6.0amzn1.0, OFI NCCL plugin 1.21.1; 1.49.0: libfabric 2.4.0amzn5.0, plugin 1.20.0) — `ChangeLog.md` in each tarball. Verified live, 2026-08-25.
- Two-node run on 2x `p6-b200.48xlarge`, image pinned by digest `sha256:24343eea4db90de3e9ef43c47a77b2f2ea04c6114a93021d155a74e9d837eeef`, all-rank `NCCL_DEBUG=INFO` logs and before/after EFA `hw_counters`. Verified live, 2026-08-25.
- Fail-closed build: `docker build --build-arg EFA_INSTALLER_VERSION=1.49.0` exits 1 at the gate after `#13 DONE 161.7s`. Verified live, 2026-08-25.

KeitaW (approved):
Overall LGTM! Few nits.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
erezzarum: I don't think this needs to be commented in the container.
We can just use an environment variable stating the aws-ofi-nccl version.
`ENV AWS_OFI_NCCL_VERSION=1.21.1` - this gives anyone using the container the ability to know which version of the plugin is installed, though, they can also just run `dpkg -l | grep ofi` to find out.
If someone wants to know which versions are installed by the EFA Installer, they can also look at the official Docs.

Xuan-1998: +1 to Erez

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
Xuan-1998: put the gating earlier so it fails fast

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
Xuan-1998: done

PR #1241, opened 2026-08-25
feat(deepep-v2): Add Kubernetes microbenchmarks examples, Refactor Dockerfile

erezzarum:
## Changes

- Refactor Dockerfile.
- Update README.
- Add DeepEP v2 microbenchmarks examples for running on Kubernetes.

## Test Plan

**Environment:**
- AWS Service: Amazon EKS
- Instance type: p6-b300.48xlarge
- Number of nodes: 2

**Test commands:**
```bash

export IMAGE_URI=<IMAGE URI>
export INSTANCE_TYPE=p6-b300.48xlarge
export GPU_PER_NODE=8
export EFA_PER_NODE=16
export NUM_NODES=2

envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' \
  < examples/test-intranode.yaml | kubectl apply -f -

envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' \
  < examples/test-internode.yaml | kubectl apply -f -

```

## Directory Structure

```
3.test_cases/
```

## Checklist

### DIFF
--- a/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
+++ b/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
@@ -19,8 +19,10 @@ installation works:
 - **[`deepep.Dockerfile`](./deepep.Dockerfile)** — a reference image showing the full stack the
   script needs (the EFA installer, which bundles the EFA-GDA-capable aws-ofi-nccl plugin, a
   GIN-capable NCCL, torch), built from a bare CUDA base; it can also be used directly for testing.
-- **[`slurm/`](./slurm/)** — launchers that run DeepEP's own `tests/elastic/test_ep.py` to
-  validate the installation:
+- **[`slurm/`](./slurm/)** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
+  validate the installation.
+- **[`kubernetes/`](./kubernetes/))** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
+  validate the installation.
 
 | Benchmark | Nodes | Measures |
 |-----------|-------|----------|
@@ -30,7 +32,7 @@ installation works:
 ## ⚠️ Version constraints
 
 > - **NCCL >= 2.31** — the GIN device API (`nccl_device.h`) DeepEP V2 links against.
->   The Dockerfile builds NCCL from source and fails the build if the header is missing.
+>   The Dockerfile install NVIDIA's precompiled NCCL from PyPi as a python package.
 > - **EFA installer >= 1.50** — provides everything EFA-GDA needs in the container (the
 >   userspace runtime and the aws-ofi-nccl plugin with the EFA-GDA GIN backend, type 5,
 >   this benchmark measures). Installers up to 1.49.0 will not work; the image build fails
@@ -38,7 +40,7 @@ installation works:
 >   backend, which has no such floor.
 > - A host EFA kernel driver **>= 3.3.0**, the first release with the completion-counter API
 >   (`efadv_create_comp_cntr`). Run the EFA installer **on the node itself** (not in the
->   container) to install/upgrade the kernel driver; the container ships only the userspace
+>   container) to install/upgrade the kernel driver; the provided Dockerfile ships only the userspace
 >   stack. Check with `modinfo efa | grep ^version`; stock AMIs may ship older.
 > - The **gdrcopy kernel module (`gdrdrv`) loaded on compute nodes** — the launchers
 >   bind-mount `/dev/gdrdrv`, and the plugin's GIN initialization opens a gdr handle.
@@ -58,18 +60,16 @@ it, if left unset, NCCL selects the CPU-proxy backend when running on EFA.
 | Variable | Value | Purpose |
 |----------|-------|---------|
 | `NCCL_GIN_TYPE` | `5` | Pins the EFA-GDA GIN backend; unset falls back to the CPU-proxy backend. |
-| `EP_NCCL_ROOT_DIR` | `$NCCL_HOME` | Points DeepEP's build/runtime checks at the GIN-capable NCCL. |
-| `LD_PRELOAD` | `$NCCL_HOME/lib/libnccl.so.2` | Guarantees the GIN-capable NCCL wins over any other on the loader path. |
 
 ## Prerequisites
 
 - EFA-enabled GPU nodes with the EFA kernel driver installed (section below).
-- **GPU architecture:** the image builds for **Hopper (`sm_90`) and Blackwell (`sm_100`)** by
+- **GPU architecture:** the image builds for **Hopper (`sm_90`) and Blackwell (`sm_100`, `sm_103`)** by
   default, so one image runs on `p5`/`p5en` and `p6`. Earlier DeepEP revisions required
   **CUDA >= 13.1** on p6/Blackwell (ptxas 13.0.88 rejects the 32-bit `st.bulk` size operand);
   DeepEP main carries the fix ([amazon-contributing/DeepEP#3](https://github.com/amazon-contributing/DeepEP/pull/3)),
-  so CUDA 13.0 works as well. Override
-  `--build-arg TORCH_CUDA_ARCH_LIST` / `--build-arg NVCC_GENCODE` for a single-arch image.
+  so CUDA 13.0 works.
+  Override `--build-arg TORCH_CUDA_ARCH_LIST` / `--build-arg NVCC_GENCODE` for a single-arch image.
 - Docker with BuildKit, and enroot + pyxis on the cluster for the Slurm flow.
 
 ## Using the script in your own container
@@ -81,11 +81,10 @@ it, if left unset, NCCL selects the CPU-proxy backend when running on EFA.
 
 The script refuses to run if the NCCL at `--nccl-root` lacks the GIN device API, so it cannot
 silently produce a DeepEP that falls back to a slower path. Runtime still requires the
-EFA-GDA-capable aws-ofi-nccl GIN plugin (bundled with EFA installer >= 1.50): run the EFA
-installer **inside your container** too (with `--skip-kmod`, as the reference Dockerfile does)
-to provide the userspace stack and the plugin, then point `NCCL_NET_PLUGIN` and
-`NCCL_GIN_PLUGIN` at `/opt/amazon/ofi-nccl/lib/libnccl-net-ofi.so`. The host installer run
-(previous section) provides only the kernel driver.
+EFA-GDA-capable aws-ofi-nccl GIN plugin (bundled with EFA installer >= 1.50).
+
+> NOTE: The aws-ofi-nccl plugin has to be in your container, run the EFA Installer inside your container with the `-y --skip-kmod --skip-limit-conf --no-verify` arguments. \
+> For installing the EFA Driver on the host, the EFA Installer bundles the driver, run the EFA Installer on the host with the `--minimal -y` arguments.
 
 ## Building the reference image
 
@@ -96,11 +95,11 @@ DOCKER_BUILDKIT=1 docker build --progress=plain -f ./deepep.Dockerfile \
   -t deepep-v2:gin .
 ```
 
-Override `NCCL_REF` or `EFA_INSTALLER_VERSION` for different versions:
+Override `NCCL_VERSION` or `EFA_INSTALLER_VERSION` for different versions:
 
 ```bash
 DOCKER_BUILDKIT=1 docker build --progress=plain -f ./deepep.Dockerfile \
-  --build-arg NCCL_REF=v2.31.2-1 \
+  --build-arg NCCL_VERSION=2.31.2 \
   --build-arg EFA_INSTALLER_VERSION=1.50.0 \
   -t deepep-v2:gin .
 ```
--- a/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
+++ b/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
@@ -21,7 +21,7 @@ installation works:
   GIN-capable NCCL, torch), built from a bare CUDA base; it can also be used directly for testing.
 - **[`slurm/`](./slurm/)** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
   validate the installation.
-- **[`kubernetes/`](./kubernetes/))** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
+- **[`kubernetes/`](./kubernetes/)** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
   validate the installation.
 
 | Benchmark | Nodes | Measures |
--- a/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
+++ b/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
@@ -4,15 +4,13 @@ ARG CUDA_VERSION=13.0.2
 
 FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04
 
-ARG TARGETARCH
-
 ARG GDRCOPY_VERSION=v2.5.2
 ARG EFA_INSTALLER_VERSION=1.50.0
-ARG DEEPEP_REPO="https://github.com/amazon-contributing/DeepEP.git"
 ARG DEEPEP_REF="main"
 ARG NCCL_VERSION=2.31.2
 ARG PYTHON_VERSION=3.10
 ARG TORCH_VERSION=2.11.0
+ARG TORCH_CUDA_INDEX=cu130
 
 # CUDA architecture(s), semicolon-separated:
 # 9.0 = Hopper (H100, sm_90), 10.0 and 10.3 = Blackwell (B200/B300, sm_100,sm_103). Defaults to
@@ -40,7 +38,7 @@ RUN rm -rf /opt/hpcx \
     && rm -f /etc/ld.so.conf.d/hpcx.conf \
     && ldconfig
 
-RUN apt-get install -y --allow-unauthenticated \
+RUN apt-get install -y --no-install-recommends \
     apt-utils \
     autoconf \
     automake \
@@ -114,9 +112,7 @@ ENV OMPI_MCA_pml=^ucx            \
 ENV PMIX_MCA_gds=hash
 
 ################################ PyTorch ########################################
-RUN CUDA_VER=$(nvcc --version | grep -oP 'release \K[0-9]+\.[0-9]+' | tr -d '.') && \
-    CUDA_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') && \
-    pip3 install torch==${TORCH_VERSION} numpy --index-url https://download.pytorch.org/whl/cu${CUDA_VER}
+RUN pip3 install torch==${TORCH_VERSION} numpy --index-url https://download.pytorch.org/whl/${TORCH_CUDA_INDEX}
 
 ################################ DeepEP v2 ########################################
 RUN --mount=type=bind,source=setup_deepep_gin.sh,target=/tmp/setup_deepep_gin.sh \
@@ -124,9 +120,9 @@ RUN --mount=type=bind,source=setup_deepep_gin.sh,target=/tmp/setup_deepep_gin.sh
     CUDA_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') && \
     pip3 install --no-deps nvidia-nccl-cu${CUDA_MAJOR}==${NCCL_VERSION} && \
     /tmp/setup_deepep_gin.sh \
-      --deepep-repo $DEEPEP_REPO \
       --deepep-ref $DEEPEP_REF \
       --nccl-root "/usr/local/lib/python${PYTHON_VERSION}/dist-packages/nvidia/nccl"
 
+ENV NVIDIA_GDRCOPY=enabled
 ENV NCCL_NET_PLUGIN=ofi
 ENV NCCL_TUNER_PLUGIN=ofi
--- a/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
+++ b/micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
@@ -1,164 +1,132 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: MIT-0
-#
-# DeepEP V2 image for AWS EFA clusters.
-# See README.md for build and run instructions.
+ARG CUDA_VERSION=13.0.2
 
-# CUDA 13.1 is the default. Earlier DeepEP revisions needed >= 13.1 on
-# p6/Blackwell (ptxas 13.0.88 rejects the 32-bit st.bulk size operand); DeepEP
-# main carries the 64-bit fix (amazon-contributing/DeepEP#3), so 13.0 works too.
-ARG CUDA_VERSION=13.1.2
 FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04
 
-ARG TORCH_VERSION=2.11.0
-# Torch wheel index. Kept explicit rather than derived from nvcc: pytorch.org has
-# no cu131 index, and cu130 wheels run on a 13.1 toolkit (same CUDA major).
-ARG TORCH_CUDA_INDEX=cu130
+ARG TARGETARCH
+
 ARG GDRCOPY_VERSION=v2.5.2
-
-# EFA installer: provides the complete EFA userspace stack (OpenMPI, the EFA
-# runtime, and the aws-ofi-nccl plugin). EFA-GDA requires installer >= 1.50;
-# the build fails loudly on a too-old installer (plugin gate below).
-# EFA_INSTALLER_TARBALL overrides the download with a local tarball staged next
-# to this Dockerfile.
 ARG EFA_INSTALLER_VERSION=1.50.0
-ARG EFA_INSTALLER_TARBALL=
-
-# NCCL: defaults to the public NVIDIA release. For internal use, override with
-# the upstream fork + staging branch (see build instructions in README.md).
-ARG NCCL_REPO=https://github.com/NVIDIA/nccl.git
-ARG NCCL_REF=v2.31.2-1
-
-# DeepEP V2. The repo is pinned to amazon-contributing/DeepEP inside
-# setup_deepep_gin.sh; only the ref is overridable.
-ARG DEEPEP_REF=main
+ARG DEEPEP_REPO="https://github.com/amazon-contributing/DeepEP.git"
+ARG DEEPEP_REF="main"
+ARG NCCL_VERSION=2.31.2
+ARG PYTHON_VERSION=3.10
+ARG TORCH_VERSION=2.11.0
 
 # CUDA architecture(s), semicolon-separated:
-# 9.0 = Hopper (H100/H200, sm_90); 10.0/10.3 = Blackwell (B200/B300, sm_100/sm_103).
-# Defaults to Hopper + Blackwell so one image runs on p5/p5en and p6.
-ARG TORCH_CUDA_ARCH_LIST="9.0;10.0"
-ARG NVCC_GENCODE="-gencode=arch=compute_90,code=sm_90 -gencode=arch=compute_100,code=sm_100"
+# 9.0 = Hopper (H100, sm_90), 10.0 and 10.3 = Blackwell (B200/B300, sm_100,sm_103). Defaults to
+# both so one image runs on Hopper and Blackwell; override with e.g.
+# --build-arg TORCH_CUDA_ARCH_LIST=9.0 \
+# --build-arg NVCC_GENCODE=-gencode=arch=compute_90,code=sm_90 to build a smaller Hopper-only image.
+ARG TORCH_CUDA_ARCH_LIST="9.0;10.0;10.3"
+ARG NVCC_GENCODE="-gencode=arch=compute_90,code=sm_90 -gencode=arch=compute_100,code=sm_100 -gencode=arch=compute_103,code=sm_103"
+
 ARG CUDA_HOME="/usr/local/cuda"
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-## Remove conflicting distro packages
-RUN apt-get update -y \
-    && apt-get remove -y --allow-change-held-packages \
+RUN apt-get update -y && \
+    apt-get remove -y --allow-change-held-packages \
     ibverbs-utils \
     libibverbs-dev \
     libibverbs1 \
     libmlx5-1 \
     libnccl2 \
-    libnccl-dev || true
-RUN rm -rf /opt/hpcx /usr/local/mpi /etc/ld.so.conf.d/hpcx.conf && ldconfig
+    libnccl-dev
 
-## Toolchain
-RUN apt-get update -y && apt-get install -y \
-    apt-utils autoconf automake build-essential check cmake curl debhelper \
-    devscripts git gcc gdb libsubunit-dev libtool ninja-build meson pandoc \
-    pkg-config vim wget \
-    python3.10-dev python3.10-venv python3-distutils \
-    cython3 \
-    libnl-3-dev libnl-route-3-dev libudev-dev libsystemd-dev \
-    libhwloc-dev \
-    && rm -rf /var/lib/apt/lists/*
-RUN apt-get purge -y cuda-compat-* || true
+RUN rm -rf /opt/hpcx \
+    && rm -rf /usr/local/mpi \
+    && rm -f /etc/ld.so.conf.d/hpcx.conf \
+    && ldconfig
+
+RUN apt-get install -y --allow-unauthenticated \
+    apt-utils \
+    autoconf \
+    automake \
+    build-essential \
+    check \
+    cmake \
+    curl \
+    debhelper \
+    devscripts \
+    git \
+    gcc \
+    gdb \
+    libsubunit-dev \
+    libtool \
+    openssh-client \
+    openssh-server \
+    pkg-config \
+    python3-distutils \
+    vim \
+    python${PYTHON_VERSION}-dev \
+    python${PYTHON_VERSION}-venv
+
+RUN apt-get purge -y cuda-compat-*
+
+RUN mkdir -p /var/run/sshd
+RUN sed -i 's/[ #]\(.*StrictHostKeyChecking \).*/ \1no/g' /etc/ssh/ssh_config && \
+    echo "    UserKnownHostsFile /dev/null" >> /etc/ssh/ssh_config && \
+    sed -i 's/#\(StrictModes \).*/\1no/g' /etc/ssh/sshd_config
 
 RUN curl https://bootstrap.pypa.io/get-pip.py -o /tmp/get-pip.py \
     && python3 /tmp/get-pip.py \
-    && pip3 install awscli nvidia-ml-py ninja Cython
+    && pip3 install awscli nvidia-ml-py ninja
 
-## GDRCopy
-# Required by aws-ofi-nccl: the plugin's GIN initialization opens a gdr handle
-# regardless of backend, so building or running without GDRCopy leaves the GIN
-# backends unable to initialize (verified: DeepEP then fails with "NCCL GIN is
-# unavailable"). Built from source because no distro package ships it. The
-# matching gdrdrv kernel module must be loaded on the host (see README).
+#################################################
+## Install NVIDIA GDRCopy
 ARG GDRCOPY_PREFIX="/opt/gdrcopy"
 RUN git clone -b ${GDRCOPY_VERSION} https://github.com/NVIDIA/gdrcopy.git /tmp/gdrcopy \
     && cd /tmp/gdrcopy \
     && make prefix="${GDRCOPY_PREFIX}" install \
     && rm -rf /tmp/gdrcopy
+
 ENV LD_LIBRARY_PATH="${GDRCOPY_PREFIX}/lib:$LD_LIBRARY_PATH"
 ENV LIBRARY_PATH="${GDRCOPY_PREFIX}/lib:$LIBRARY_PATH"
 ENV PATH="${GDRCOPY_PREFIX}/bin:$PATH"
 
-## EFA installer (OpenMPI + EFA runtime; local tarball wins when set)
-RUN --mount=type=bind,target=/ctx \
-    apt-get update -y \
-    && cd /tmp \
-    && if [ -n "${EFA_INSTALLER_TARBALL}" ]; then \
-         echo "=== EFA installer: LOCAL tarball /ctx/${EFA_INSTALLER_TARBALL} ==="; \
-         if [ ! -f "/ctx/${EFA_INSTALLER_TARBALL}" ]; then \
-           echo "ERROR: ${EFA_INSTALLER_TARBALL} not found in build context" >&2; exit 1; \
-         fi; \
-         tar -xf "/ctx/${EFA_INSTALLER_TARBALL}"; \
-       else \
-         echo "=== EFA installer: downloading v${EFA_INSTALLER_VERSION} ==="; \
-         curl -fsSL -O "https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz" \
-         && tar -xf "aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz"; \
-       fi \
-    && cd aws-efa-installer \
-    && ./efa_installer.sh --disable-ngc -y --skip-kmod --skip-limit-conf --no-verify \
-    && ldconfig \
-    && { nm -D /opt/amazon/ofi-nccl/lib/libnccl-net-ofi.so 2>/dev/null | grep -qw ncclGinPlugin_v14 \
-         && echo "OK: bundled plugin is EFA-GDA-capable (ncclGinPlugin_v14 present)" \
-         || { echo "ERROR: installer's plugin has no ncclGinPlugin_v14 export: EFA-GDA requires installer >= 1.50" >&2; exit 1; }; } \
-    && rm -rf /tmp/aws-efa-installer* /var/lib/apt/lists/*
+#################################################
+## Install EFA installer
+RUN cd $HOME && \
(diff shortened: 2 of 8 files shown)


KeitaW (commented):
Thanks for this — the Kubernetes path for the DeepEP V2 benchmark is a genuinely useful
addition, and the MPIJob design underneath it is sound. I took it to a real cluster rather than
reading it: **2× `p6-b300.48xlarge` on EKS in `us-east-1`**, building the image from this
branch's own Dockerfile and applying the manifests exactly as the README says. Both benchmarks
do pass there — after two changes, one of which is a single line.

To be precise about what was verified at the current head: the image build, the **intranode**
job (8 ranks, ~720 GB/s over NVLink) and the **internode** job (16 ranks across 2 nodes,
**116–118 GB/s scale-out** over EFA-GDA) — all three, with exactly two deltas applied: the
`--deepep-repo` line removed so the image would build, and `/dev/gdrdrv` made reachable.

Because the branch moved while I was testing (`db4df83c` → `2b537098`), a note on what that
changed: the four `kubernetes/` files are **byte-identical** across both commits, so everything
below about the manifests applies as written. The new commit fixed two things I had already hit
— see the closing batch.

## Review Batch 1/5 — Build & Image Contract

The image cannot currently be built, and the rewrite removes the build-time checks that used to
make "this image can actually do GIN" non-negotiable.

### The local-tarball EFA installer path was removed (nit)

`main` supported `EFA_INSTALLER_TARBALL` to build from a tarball staged in the build context;
this rewrite always `curl`s from the internet, so egress-restricted builds that used to work now
fail. Worth calling out in the PR description if the removal is deliberate.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -1,195 +1,132 @@
...
     /tmp/setup_deepep_gin.sh \
-        --deepep-ref "${DEEPEP_REF}" \
-        --deepep-prefix "${DEEPEP_PREFIX}" \
-        --nccl-root "${NCCL_HOME}"
-
-## Runtime env
-ENV NVIDIA_GDRCOPY=enabled
-ENV NCCL_OFI_RDMA_GDR_FLUSH_DISABLE=0
-WORKDIR /root
+      --deepep-repo $DEEPEP_REPO \
```
KeitaW: ### The build fails at the DeepEP step — `setup_deepep_gin.sh` has no `--deepep-repo` flag (blocking) [confirmed]

Building the image straight from this branch fails:

```
#17 4.084 ERROR: unrecognized argument: '--deepep-repo'
ERROR: failed to build: ... deepep.Dockerfile:122 ... exit code: 1
```

`setup_deepep_gin.sh` accepts `--deepep-ref`, `--deepep-prefix`, `--deepep-src`, `--nccl-root`,
`--python`, `--pip`, `--skip-checks` (lines 86–104) and is not modified by this PR. The parent
README states the constraint directly: "The repo is pinned to amazon-contributing/DeepEP inside
`setup_deepep_gin.sh`; only the ref is overridable."

Good news on scope: I rebuilt with **only** this line deleted and the image built clean
(`BUILD_EXIT=0`, 22.9 GB), so this is the sole build blocker rather than the first of many.
Either drop the arg, or add `--deepep-repo` to the script's parser in the same PR.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -1,195 +1,132 @@
...
+      --deepep-ref $DEEPEP_REF \
+      --nccl-root "/usr/local/lib/python${PYTHON_VERSION}/dist-packages/nvidia/nccl"
+
+ENV NCCL_NET_PLUGIN=ofi
```
KeitaW, line 127: ### Two build gates that guaranteed an EFA-GDA-capable image are gone, and the README still advertises them (should fix) [confirmed]

Comparing against `main`'s version of this file:

| Gate on `main`'s Dockerfile | here |
|---|---|
| `ncclGinPlugin_v14` exported — plugin has EFA-GDA | removed |
| libfabric `>= 2.5` version check | removed |

Two things I want to get right rather than overstate. First, the `nccl_device.h` check is **not**
lost — `setup_deepep_gin.sh:177` still refuses a non-GIN NCCL, so that guarantee survives (once
the build blocker above is fixed). Second, dropping `--disable-ngc` is **benign**: on this base
the installer auto-detects NGC and installs `libnccl-ofi-ngc-v3`, but `ngc_post_install` still
registers the loader path, and my build confirms it — the plugin loads from
`/opt/amazon/ofi-nccl/lib/libnccl-net-ofi.so` at runtime.

What is genuinely gone is the pair above. And `kubernetes/README.md:34` still tells the reader
the build "fails loudly on a too-old installer" — with the libfabric check removed and
`EFA_INSTALLER_VERSION` documented as overridable, a 1.49.0 build now succeeds and only fails at
runtime when `NCCL_GIN_TYPE=5` finds no EFA-GDA backend. Either restore the gate or drop the claim.

The underlying swap is fine, to be clear: moving NCCL to `pip install nvidia-nccl-cu13==2.31.2`
is legitimate — I checked the wheel and it ships `nccl_device.h` plus the whole
`nccl_device/gin/efa_gda/*` header set — and on the cluster the installer-bundled plugin reports
`GIN/Plugin: Loaded gin plugin Libfabric_GDAKI (v14)`, so the `ncclGinPlugin_v14` gate *would
have passed*.

So the concern isn't that the capability is missing — it's that nothing now fails the build if a
future installer or wheel bump silently drops it, on an image whose entire purpose is measuring
that backend. These gates were the thing standing between a bad bump and a benchmark that
quietly measures the CPU-proxy path instead. Could the `ncclGinPlugin_v14` check at least be
kept against the bundled plugin path?

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -1,195 +1,132 @@
...
+    OPAL_PREFIX=/opt/amazon/openmpi \
+    NCCL_SOCKET_IFNAME=^docker,lo,veth
+
+## Turn off PMIx Error https://github.com/open-mpi/ompi/issues/7516
+ENV PMIX_MCA_gds=hash
+
+################################ PyTorch ########################################
+RUN CUDA_VER=$(nvcc --version | grep -oP 'release \K[0-9]+\.[0-9]+' | tr -d '.') && \
+    CUDA_MAJOR=$(nvcc --version | grep -oP 'release \K[0-9]+') && \
+    pip3 install torch==${TORCH_VERSION} numpy --index-url https://download.pytorch.org/whl/cu${CUDA_VER}
```
KeitaW: ### Overriding `CUDA_VERSION` to the previous default breaks the torch install, with a misleading error (should fix)

The wheel index is derived from `nvcc`:

```
pip3 install torch==${TORCH_VERSION} numpy --index-url https://download.pytorch.org/whl/cu${CUDA_VER}
```

That resolves to `cu130` at the new default and works. But `CUDA_VERSION` is still an `ARG`, and
the value it had until this commit — `13.1.2` — yields `cu131`, which PyTorch does not publish
(`https://download.pytorch.org/whl/cu131` → HTTP 403; `cu130` → 200, verified live). So the one
override a reader is most likely to try is the one that breaks. It fails loudly — pip exits 1 —
but the message points somewhere else entirely: `ERROR: Could not find a version that satisfies
the requirement nvidia-nccl-cu13==...`, which reads like a wheel-availability problem rather
than "your `CUDA_VERSION` produced an index URL that doesn't exist". `main` handled this with an
explicit `ARG TORCH_CUDA_INDEX=cu130` and a comment saying exactly why; both were removed here.
I'd restore the explicit index rather than deriving it.

(`CUDA_MAJOR` on the line above is computed and never used in that `RUN` — dead.)

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -1,195 +1,132 @@
...
-    && rm -rf /var/lib/apt/lists/*
-RUN apt-get purge -y cuda-compat-* || true
+    libnccl-dev
+
+RUN rm -rf /opt/hpcx \
+    && rm -rf /usr/local/mpi \
+    && rm -f /etc/ld.so.conf.d/hpcx.conf \
+    && ldconfig
+
+RUN apt-get install -y --allow-unauthenticated \
```
KeitaW: ### `--allow-unauthenticated` disables apt signature verification (should fix) [confirmed]

```suggestion
RUN apt-get install -y \
```

This turns off package signature checking for the whole toolchain install. It doesn't look
necessary: `main`'s version of this Dockerfile installs the same toolchain from the same base
with no such flag, and I built that one clean earlier in this review. If it was added to get
past a transient repo issue, an `apt-get update` in the same layer is the safer shape.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
```diff
@@ -1,195 +1,132 @@
...
+ARG CUDA_VERSION=13.0.2
+
 FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04
 
-ARG TORCH_VERSION=2.11.0
-# Torch wheel index. Kept explicit rather than derived from nvcc: pytorch.org has
-# no cu131 index, and cu130 wheels run on a 13.1 toolkit (same CUDA major).
-ARG TORCH_CUDA_INDEX=cu130
-ARG GDRCOPY_VERSION=v2.5.2
+ARG TARGETARCH
```
KeitaW: ### `ARG TARGETARCH` is declared and never used (nit)

Nothing in the file references it. This repo has flagged the same dead-arg pattern before, so
worth removing rather than leaving as a false signal that the image is multi-arch — I'd just
delete the `ARG TARGETARCH` line.

KeitaW (commented):
## Review Batch 2/5 — Device Access & Runtime Environment

This is the one that actually stops the benchmark running on a stock EKS GPU cluster.

### `NVIDIA_GDRCOPY=enabled` was dropped from the image ENV (should fix)

`main` sets `ENV NVIDIA_GDRCOPY=enabled`; this rewrite removes it, leaving the two k8s manifests
as its only home. That reverses a convention this repo adopted on purpose (PR #1145 moved this
var *out of* the launch manifests and *into* the Dockerfile so it has exactly one home and can't
drift). The slurm path is unaffected — it bind-mounts the device directly — but I'd put the
`ENV` back rather than have the manifests be the only place it exists.
`NCCL_OFI_RDMA_GDR_FLUSH_DISABLE=0` and `NCCL_GIN_PLUGIN` were dropped in the same rewrite;
`NCCL_GIN_PLUGIN`'s absence is verified harmless (GIN still resolves via `NCCL_NET_PLUGIN=ofi`,
since one shared object exports the net/rma/gin tables).

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
```diff
@@ -0,0 +1,148 @@
...
+                value: "enabled"
+            # The tuning phase profiles kernels via CUPTI. These nodes set the
+            # driver's RmProfilingAdminOnly=1, so the workload needs SYS_ADMIN to
+            # profile. Remove if your cluster allows non-admin GPU profiling.
+            securityContext:
+              capabilities:
+                add: ["SYS_ADMIN", "IPC_LOCK"]
+            volumeMounts:
+            - name: shmem
+              mountPath: /dev/shm
```
KeitaW, line 136: ### The pods never get `/dev/gdrdrv`, so DeepEP aborts with "NCCL GIN is unavailable" (blocking) [confirmed]

With the sshd problem out of the way, both jobs still failed for me — `BackoffLimitExceeded`:

```
RuntimeError: Assertion exception (csrc/kernels/backend/nccl.cu:101):
  (allow_hybrid_mode ? props.railedGinType : props.ginType) != NCCL_GIN_TYPE_NONE
  and "NCCL GIN is unavailable. ..."
```

Root cause, from `NCCL_DEBUG_SUBSYS=INIT,NET,ENV`:

```
NCCL INFO  NET/OFI gin: Initializing
NCCL INFO  NET/OFI Failed to open gdr handle
NCCL WARN  NET/OFI Failed to initialize GDRCopy: Failed to open gdr handle   (nccl_ofi_gin_init:71)
NCCL INFO  GIN/Plugin: Failed to initialize any GIN plugin
```

The manifests rely on `env: NVIDIA_GDRCOPY=enabled` to have the container runtime inject the
device. On this cluster it isn't injected — `kubectl exec ... ls -l /dev/gdrdrv` →
`No such file or directory`. That isn't cluster misconfiguration: gdrdrv auto-injection by the
NVIDIA device plugin only ever happened as a default-on bug in **v0.19.0–0.19.2**, reverted in
**v0.19.3** ([k8s-device-plugin#1692](https://github.com/NVIDIA/k8s-device-plugin/issues/1692),
[#1837](https://github.com/NVIDIA/k8s-device-plugin/pull/1837)); this cluster runs **v0.18.1**.
So on any supported plugin version the env var alone cannot deliver the device.

Worth knowing: a `hostPath` mount by itself is **not** enough either — the node appears but
cgroup v2 denies the open:

```
$ ls -l /dev/gdrdrv          crw-rw-rw-. 1 root root 237, 0
$ python3 -c 'os.open("/dev/gdrdrv", os.O_RDWR)'
OPEN FAILED: [Errno 1] Operation not permitted
```

So the fix needs **both halves** — the device has to be visible *and* openable. `hostPath` gives
you the first; the second needs `privileged: true`, because Kubernetes has no way to grant a
single device to a non-privileged container (there's no `securityContext.devices`, and a
`hostPath` mount doesn't add a device-cgroup allow rule). With both applied, the pod opens the
handle (`OPEN OK fd= 3`) and both benchmarks pass end-to-end on B300.

Concretely, in **both** `test-intranode.yaml` and `test-internode.yaml`, on the **worker**:

```suggestion
            securityContext:
              privileged: true
              capabilities:
                add: ["SYS_ADMIN", "IPC_LOCK"]
            volumeMounts:
            - name: shmem
              mountPath: /dev/shm
            - name: gdrdrv
              mountPath: /dev/gdrdrv
```

and the matching volume in the `volumes:` block lower down (`:144-148` — a separate hunk, so it
can't ride along with the suggestion above):

```yaml
          volumes:
          - name: gdrdrv
            hostPath:
              path: /dev/gdrdrv
              type: CharDevice
          - name: shmem
            emptyDir:
              medium: Memory
              sizeLimit: 200Gi
```

Keep `NVIDIA_GDRCOPY=enabled` as-is — it's harmless and does the right thing on clusters whose
toolkit honours it.

I do want to flag the tradeoff rather than hand you `privileged` silently: it's a big grant for
a benchmark pod. If you'd rather not, the cleaner routes are a device plugin that advertises
`/dev/gdrdrv` as a resource, or a CDI spec that includes it — either injects the device *with*
its cgroup permission and needs no `privileged`. Both need cluster-side setup, which a sample
targeting arbitrary EKS clusters probably can't assume, so I'd take the manifest change here and
note the alternative in the README. (I verified the `privileged` route; I have not tested the
other two.)

For contrast, the `slurm/` siblings don't hit this at all — enroot bind-mounts the device
(`--container-mounts /dev/infiniband:/dev/infiniband,/dev/gdrdrv:/dev/gdrdrv`) and applies no
device cgroup. That's also why the parent README's prerequisite ("the launchers bind-mount
`/dev/gdrdrv`") reads as satisfied while the new k8s path silently isn't covered by it — worth
rewording per-launcher once the manifests are fixed.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
```diff
@@ -0,0 +1,148 @@
...
+                done
+              done
+              MASTER_ADDR=$(head -n1 /etc/mpi/hostfile | awk '{print $1}')
+              # SSH-launched workers do not inherit Dockerfile ENV, so pass every
+              # needed variable through `mpirun -x`.
+              /opt/amazon/openmpi/bin/mpirun \
+                --allow-run-as-root --tag-output \
+                -np ${NUM_NODES} -N 1 --bind-to none \
+                -x PATH -x LD_LIBRARY_PATH \
+                -x NCCL_SOCKET_IFNAME -x LD_PRELOAD \
```
KeitaW: ### `-x LD_PRELOAD` forwards a variable nothing sets (should fix) [confirmed]

`mpirun` says so itself in my run:

```
[deepep-v2-internode-launcher:00001] Warning: could not find environment variable "LD_PRELOAD"
```

This was inherited from the v1 manifests, where `deepep-benchmark/deepep.Dockerfile` sets
`ENV LD_PRELOAD=/opt/nccl/build/lib/libnccl.so`. The v2 image sets no `LD_PRELOAD`, and this
PR's commit also *removes* `export LD_PRELOAD=...` from both slurm launchers — so at this head
nothing in the repo sets it anywhere.

It's harmless in practice (the image carries exactly one NCCL, and the run confirms the right
one loads: `NCCL version 2.31.2+cuda13.3`), so this is really about the README claim below and
about not carrying a no-op flag. Note `NCCL_SOCKET_IFNAME` on the same line **is** set
(Dockerfile:111) and does forward correctly — only the `LD_PRELOAD` half is dead.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+```bash
+kubectl apply -f https://raw.githubusercontent.com/kubeflow/mpi-operator/refs/tags/v0.8.2/deploy/v2beta1/mpi-operator.yaml
+kubectl get crd mpijobs.kubeflow.org      # confirm the CRD exists
+```
+
+## Build & push the container image
+
+Build the image from [`../deepep.Dockerfile`](../deepep.Dockerfile) and push it to ECR. The
+default image is built for **both Hopper (`sm_90`) and Blackwell (`sm_100`)**, so it runs on
```
KeitaW: ### `kubernetes/README.md` contradicts itself on which GPU architectures the image covers (should fix) [confirmed]

Line 38 says the image is built for "**both Hopper (`sm_90`) and Blackwell (`sm_100`)**", while
line 94 in the same file says "the default image already includes `sm_103`". The Dockerfile now
builds `9.0;10.0;10.3`, so line 94 is the accurate one and line 38 is stale.

```suggestion
default image is built for **Hopper (`sm_90`) and Blackwell (`sm_100`, `sm_103`)**, so it runs on
```

KeitaW (commented):
## Review Batch 3/5 — Documentation Consistency

All of these are copy-paste-and-it-fails issues in the new `kubernetes/README.md`.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Configure & launch
+
+Copy the committed template to `env_vars` (gitignored) and edit it with your image URI, instance
+type, and per-node device counts, then `envsubst` the manifest into `kubectl`. Restrict
+substitution to the known variables so the launcher's runtime shell vars
+(`$OMPI_COMM_WORLD_RANK`, `$MASTER_ADDR`, `$PATH`, …) are left intact:
+
+```bash
+cp env_vars.example env_vars   # then edit env_vars
```
KeitaW, line 67: ### The build block `cd ..`s, then the next block assumes you never left `kubernetes/` (should fix) [confirmed]

Line 50 does `cd ..` to reach the build context, and the "Configure & launch" block at line 63
then runs `cp env_vars.example env_vars` — but after that `cd ..` you're in
`deepep-v2-benchmark/`, where `env_vars.example` does not exist. Following the README top to
bottom fails here.

```suggestion
cd kubernetes   # back to the manifests
cp env_vars.example env_vars   # then edit env_vars
```

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/README.md
```diff
@@ -0,0 +1,113 @@
...
+envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' \
+  < test-internode.yaml | kubectl apply -f -
+```
+
+## Monitor
+
+```bash
+kubectl get mpijob
+watch kubectl get pods -o wide
+kubectl logs -f $(kubectl get pods | grep deepep.*launcher | cut -d ' ' -f 1)
```
KeitaW: ### The monitor one-liner breaks once there is more than one launcher pod (nit)

`grep deepep.*launcher | cut -d ' ' -f 1` also matches a previous run's completed launcher, and
`cut -d ' '` is fragile against `kubectl`'s column padding.

```suggestion
kubectl logs -f job/deepep-v2-internode-launcher
```

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/README.md
```diff
@@ -0,0 +1,113 @@
...
+default image is built for **both Hopper (`sm_90`) and Blackwell (`sm_100`)**, so it runs on
+`p5`/`p5en` and `p6-b300` nodes without a rebuild.
+
+```bash
+export AWS_REGION=$(aws ec2 describe-availability-zones --output text --query 'AvailabilityZones[0].[RegionName]')
+export ACCOUNT=$(aws sts get-caller-identity --query Account --output text)
+export REGISTRY=${ACCOUNT}.dkr.ecr.${AWS_REGION}.amazonaws.com
+export IMAGE_URI=${REGISTRY}/deepep-v2:efa1.50.0-nccl2.31.2
+
+aws ecr create-repository --repository-name deepep --region ${AWS_REGION} 2>/dev/null || true
```
KeitaW: ### The ECR block creates `deepep` but pushes `deepep-v2` (should fix) [confirmed]

Running the README's block verbatim, the push fails — the repository it creates is not the one
it pushes to. `env_vars.example` names a third variant (`.../deepep:<tag>`).

```suggestion
aws ecr create-repository --repository-name deepep-v2 --region ${AWS_REGION} 2>/dev/null || true
```

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/env_vars.example
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-License-Identifier: MIT-0
+
+# Copy this file to env_vars (gitignored), edit the values, then source it:
+#   cp env_vars.example env_vars
+#   source env_vars
+#   envsubst '$IMAGE_URI $INSTANCE_TYPE $GPU_PER_NODE $EFA_PER_NODE $NUM_NODES' \
+#     < test-internode.yaml | kubectl apply -f -
+
+# ECR image built from ../deepep.Dockerfile (default image supports Hopper + Blackwell).
+export IMAGE_URI=<account>.dkr.ecr.<region>.amazonaws.com/deepep:<tag>
```
KeitaW: ### `env_vars.example` still points at the v1 image repo (should fix)

```suggestion
export IMAGE_URI=<account>.dkr.ecr.<region>.amazonaws.com/deepep-v2:<tag>
```

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/env_vars.example
```diff
@@ -0,0 +1,19 @@
...
+
+# GPU node instance type and per-node device counts.
+export INSTANCE_TYPE=p6-b300.48xlarge
+export GPU_PER_NODE=8
+export EFA_PER_NODE=16          # p6-b300.48xlarge has 16 EFA NICs; p5.48xlarge has 32
+
+# Number of nodes: 1 for intranode; 2 for internode / low-latency.
```
KeitaW: ### `env_vars.example` mentions a low-latency test that doesn't exist in v2 (nit)

v1 shipped `test-low-latency.yaml`; this directory has only intranode and internode, so the
comment promises a third mode that isn't here.

```suggestion
# Number of nodes: 1 for intranode; 2 for internode.
```

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/README.md
```diff
@@ -0,0 +1,113 @@
...
+- **Image:** the default image already includes `sm_103`, so no rebuild is needed.
+- **EFA count:** `EFA_PER_NODE=16` (a `p6-b300.48xlarge` exposes 16 EFA NICs). `p6-b200.48xlarge`
+  exposes 8 — adjust per instance type.
+- **Tolerations:** the manifests tolerate `nvidia.com/gpu`, `workload=bench`, and
+  `capacity-reservation` taints. If your nodes use different taints, edit the `tolerations`
+  block; if they are untainted, the extra tolerations are harmless.
+- **Multi-node DNS:** the launcher sets `MASTER_ADDR` from the first entry of the
+  mpi-operator hostfile (`/etc/mpi/hostfile`), which is worker-0's cluster-resolvable name. If
+  `init_process_group` hangs, confirm name resolution from a peer:
+  `kubectl exec deepep-internode-worker-1 -- getent hosts <worker-0-name>`.
```
KeitaW: ### Troubleshooting command names a pod that this PR never creates (nit)

The jobs here are `deepep-v2-internode`, so the suggested `kubectl exec` will not resolve.

```suggestion
  `kubectl exec deepep-v2-internode-worker-1 -- getent hosts <worker-0-name>`.
```

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/README.md
```diff
@@ -15,10 +15,12 @@ Everything else here is reference material for using the script and testing that
...
 - **[`deepep.Dockerfile`](./deepep.Dockerfile)** — a reference image showing the full stack the
-  script needs (EFA userspace, a GIN-capable NCCL, aws-ofi-nccl with the EFA-GDA backend,
-  torch), built from a bare CUDA base; it can also be used directly for testing.
-- **[`slurm/`](./slurm/)** — launchers that run DeepEP's own `tests/elastic/test_ep.py` to
-  validate the installation:
+  script needs (the EFA installer, which bundles the EFA-GDA-capable aws-ofi-nccl plugin, a
+  GIN-capable NCCL, torch), built from a bare CUDA base; it can also be used directly for testing.
+- **[`slurm/`](./slurm/)** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
+  validate the installation.
+- **[`kubernetes/`](./kubernetes/))** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
```
KeitaW: ### Doubled closing paren in the new `kubernetes/` link (nit)

```suggestion
- **[`kubernetes/`](./kubernetes/)** - launchers that run DeepEP's own `tests/elastic/test_ep.py` to
```

KeitaW (commented):
## Review Batch 4/5 — Manifest Robustness

### `GPU_PER_NODE` never reaches the benchmark (nit)

It sizes the `nvidia.com/gpu` request, but `test_ep.py --num-processes` defaults to 8 and isn't
passed, so `GPU_PER_NODE=4` would request 4 GPUs and still spawn 8 ranks. Either thread it
through or note in `env_vars.example` that it must stay 8.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
```diff
@@ -0,0 +1,148 @@
...
+              # operator made cluster-resolvable.
+              # Wait for each worker's sshd to accept connections before launching.
+              # mpi-operator starts the launcher as soon as workers are Running, but
+              # their sshd may not be ready yet; without this wait the launcher loses
+              # the SSH race, exits non-zero, and burns Job retries (backoffLimit).
+              for host in $(awk '{print $1}' /etc/mpi/hostfile); do
+                until ssh -o ConnectTimeout=3 -o BatchMode=yes -o StrictHostKeyChecking=no "$host" true 2>/dev/null; do
+                  echo "waiting for sshd on $host"; sleep 3
+                done
+              done
```
KeitaW, line 71: ### The sshd wait loop has no bound, so a dead worker hangs the job on a full node (should fix) [confirmed]

I hit this for real: when the worker couldn't start, the launcher printed

```
waiting for sshd on deepep-v2-intranode-worker-0.deepep-v2-intranode.default.svc
```

indefinitely. The worker was in `StartError` and is never restarted, so the launcher never exits,
`runPolicy.backoffLimit: 3` never fires, and the MPIJob never reaches a terminal state at all —
its conditions were still just `MPIJobCreated` when I deleted it by hand. Contrast the case where
the launcher genuinely *runs* and exits non-zero: there `backoffLimit` works exactly as intended
(I saw it reach `BackoffLimitExceeded` in about two minutes). The hang is specifically what
defeats it.

To be precise about the cost, since it varies by how the worker fails: in the case I hit, the
worker's container never started, so its pod went to phase `Failed` and the GPUs *were* released
(verified separately — a `StartError` pod is phase `Failed` and drops out of the node's resource
accounting). What persisted was a job that never ends and a launcher pod pinned to a GPU node by
its own `podAffinity`. The expensive variant is the one this loop is most likely to meet in
practice: a worker that is *Running but unreachable* — sshd died after start, a DNS or
ssh-secret problem, a network policy — stays phase `Running` and holds all 8 GPUs and 16 EFA
devices for as long as the launcher spins, which is forever.

The loop's own comment says it exists so the launcher doesn't "burn Job retries" — but it trades
a bounded retry for an unbounded wait, and on GPU nodes the unbounded wait is the worse failure.
A bounded wait that fails loudly keeps the original intent without the silent hang:

```suggestion
              for host in $(awk '{print $1}' /etc/mpi/hostfile); do
                for i in $(seq 1 100); do
                  ssh -o ConnectTimeout=3 -o BatchMode=yes -o StrictHostKeyChecking=no "$host" true 2>/dev/null && break
                  [ "$i" = 100 ] && { echo "ERROR: sshd on $host not reachable after 5min" >&2; exit 1; }
                  echo "waiting for sshd on $host"; sleep 3
                done
              done
```

The comment just above still describes the intent accurately, so it can stay. Same loop in
`test-intranode.yaml`. I'd also add a belt-and-braces `runPolicy.activeDeadlineSeconds` (e.g.
`1800`) next to `backoffLimit`, so no failure mode at all can hold the nodes indefinitely.

I hit a second, more mundane version of the same thing on a busy cluster: when the two worker
pods were simply *unschedulable* (all GPU nodes occupied by another job), the launcher was
already `Running` and printing `waiting for sshd` on a loop, against workers that had not been
scheduled at all. Nothing times out and nothing surfaces the real reason — a reader sees an
sshd message for what is actually a capacity wait.

Separately, the loop doesn't fully win the race it targets: on cached-image starts the launcher
still exited 255 within ~4s on its first attempt (twice), recovering via `restartPolicy:
OnFailure`. It behaved correctly on a slow first pull, when sshd had time to come up.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
```diff
@@ -0,0 +1,148 @@
...
+            volumeMounts:
+            - name: shmem
+              mountPath: /dev/shm
+            resources:
+              limits:
+                nvidia.com/gpu: ${GPU_PER_NODE}
+                vpc.amazonaws.com/efa: ${EFA_PER_NODE}
+              requests:
+                nvidia.com/gpu: ${GPU_PER_NODE}
+                vpc.amazonaws.com/efa: ${EFA_PER_NODE}
```
KeitaW, line 143: ### Worker resource requests regressed against the v1 manifests these were derived from (should fix)

The v1 siblings request `hugepages-2Mi: 5120Mi` and `memory: 32000Mi` alongside the GPU/EFA
resources; both were dropped here, while the `/dev/shm` emptyDir is still `medium: Memory` with
`sizeLimit: 200Gi`. A memory-backed emptyDir is charged to the pod's memory accounting, so with
no memory request or limit `/dev/shm` can grow to 200 GiB unaccounted.

The sharper consequence is QoS. Extended resources don't participate in the QoS calculation —
only `cpu` and `memory` do — so a container requesting *only* `nvidia.com/gpu` and
`vpc.amazonaws.com/efa` lands in **BestEffort**, the first class evicted under node memory
pressure. I confirmed the rule with a throwaway pod requesting a single extended resource and
nothing else: `qosClass=BestEffort`. That's an unfortunate class for an 8-GPU job that runs for
minutes. There's also no `cpu` request, which lets unrelated CPU contention move the very
numbers this benchmark exists to produce.

Dropping hugepages is defensible now that the transport is NCCL GIN rather than NVSHMEM — but
I'd put memory and cpu back, which also lifts the pod out of BestEffort:

```suggestion
            resources:
              limits:
                nvidia.com/gpu: ${GPU_PER_NODE}
                vpc.amazonaws.com/efa: ${EFA_PER_NODE}
                memory: 32000Mi
              requests:
                nvidia.com/gpu: ${GPU_PER_NODE}
                vpc.amazonaws.com/efa: ${EFA_PER_NODE}
                memory: 32000Mi
                cpu: "8"
```

(Same block in `test-intranode.yaml`. Tune the numbers to taste — the point is that *some*
`memory`/`cpu` request exists; `32000Mi` is simply what the v1 manifests used.)

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
```diff
@@ -0,0 +1,148 @@
...
+          volumes:
+          - name: shmem
+            emptyDir:
+              medium: Memory
+              sizeLimit: 200Gi
```
KeitaW: ### Both manifests are missing a trailing newline (nit)

The repo's root `.editorconfig` sets `insert_final_newline = true` for all files, so both YAMLs
want a newline after the final `sizeLimit: 200Gi`. (`printf '\n' >> file`, or just let your
editor honour the `.editorconfig`.)

KeitaW (commented):
## Review Batch 5/5 — PR Scope, and Things That Look Great

### Scope: the image rewrite probably wants to be its own PR (should fix)

The title says "Add Kubernetes microbenchmarks examples", and the first revision was exactly
that — 4 files, +409. The current head also rewrites `deepep.Dockerfile` (263 lines changed),
the parent README, and both slurm launchers. The PR comment says it "merges changes from #1239",
but this isn't #1239's Dockerfile — **#1239 is still open** and its version keeps the
`ncclGinPlugin_v14` gate this one drops. So two open PRs now propose different rewrites of the
same file. Whichever lands second will either conflict outright or, if it is rebased onto the
other, carry its own full-file version and undo the first — neither outcome is one you want to
discover at merge time.

Could the image change either move back out into #1239, or #1239 be closed explicitly in favour
of this one? The k8s manifests are independently reviewable and independently useful.

### Things That Look Great

- **The MPI topology is right, and I verified it end-to-end at this head.** `slotsPerWorker: 1`
  with `-np ${NUM_NODES} -N 1` and the test spawning 8 local ranks matches how `test_ep.py`
  reads `RANK`/`WORLD_SIZE`. Internode ran 16 ranks across 2 `p6-b300.48xlarge` and passed
  correctness plus bandwidth: **116–118 GB/s scale-out** dispatch alongside ~377–385 GB/s
  NVLink, `MPIJobSucceeded`.
- **The EFA-GDA pin does what it claims** — not a CPU-proxy false green. The log shows
  `NCCL_GIN_TYPE set by environment to 5.`, `Loaded gin plugin Libfabric_GDAKI (v14)`, and
  crucially `Skipping plugin Libfabric index 3 type 2: NCCL_GIN_TYPE=5 requested` — the
  proxy backend is actively rejected. Zero GIN init failures once the device was reachable.
- **The `podAffinity` design is correct and I confirmed it schedules.** Launcher and workers
  co-located as intended, on both the hostname key (intranode) and zone key (internode).
- **The EFA counts in the README and `env_vars.example` are all accurate** — checked each
  against the API rather than assuming: `p6-b300.48xlarge` = 16, `p6-b200.48xlarge` = 8,
  `p5.48xlarge` = 32, and the node reports `vpc.amazonaws.com/efa: 16` allocatable.
- **The CUPTI/`SYS_ADMIN` note is correct and non-obvious.** `test_ep.py` calls `bench_kineto`
  unconditionally in the perf section for *both* the intranode and internode paths, so updating
  the v1 wording ("the intranode test does not profile") to cover both was the right call.
- **`imagePullPolicy: IfNotPresent` on a pinned tag**, and an explicit `envsubst` allow-list —
  both match this repo's conventions.
- **The newest commit fixed two things I'd already hit before I could report them**: the image
  now installs `openssh-server` (mpi-operator injects `/usr/sbin/sshd -De` into workers that
  declare no command, and the earlier image had no sshd, so every worker died `StartError`), and
  `TORCH_CUDA_ARCH_LIST` now genuinely includes `10.3`, making the README's `sm_103` line true.
  I re-verified both: the head image builds `sm_90 + sm_100 + sm_103` clean on CUDA 13.0.2, and
  both benchmarks pass on B300 from the committed manifests. Nice to see the arch list fixed
  properly rather than only the sentence describing it.

## Sources

- NVIDIA k8s-device-plugin gdrdrv auto-injection was a 0.19.0–0.19.2 default-on bug, reverted in
  0.19.3 — [issue #1692](https://github.com/NVIDIA/k8s-device-plugin/issues/1692),
  [PR #1837](https://github.com/NVIDIA/k8s-device-plugin/pull/1837). Cluster under test runs
  `nvcr.io/nvidia/k8s-device-plugin:v0.18.1` (verified live, 2026-08-25).
- mpi-operator injects `/usr/sbin/sshd -De` for worker containers with no `command`/`args` —
  [`mpi_job_controller.go`](https://github.com/kubeflow/mpi-operator/blob/v0.6.0/pkg/controller/mpi_job_controller.go)
  (identical at v0.8.2).
- `setup_deepep_gin.sh` accepted flags: lines 86–104 of the file in this PR (unmodified here).
- EFA interface counts: `aws ec2 describe-instance-types` → p6-b300.48xlarge 16,
  p6-b200.48xlarge 8, p5.48xlarge 32 (verified live, 2026-08-25).
- `nvidia-nccl-cu13==2.31.2` wheel ships `nccl_device.h` and `nccl_device/gin/efa_gda/*`
  (verified live from the published wheel, 2026-08-25).
- `https://download.pytorch.org/whl/cu131` → HTTP 403, `cu130` → HTTP 200 (verified live, 2026-08-25).
- `setup_deepep_gin.sh:177` still enforces `nccl_device.h`, so that guarantee survives the
  Dockerfile rewrite (read from the file in this PR).
- Build failure, gate counts, GIN init trace, `/dev/gdrdrv` `EPERM`, and both passing runs:
  reproduced on 2× `p6-b300.48xlarge`, EKS, us-east-1 (verified live, 2026-08-25).

KeitaW (changes requested):
Thank you so much for this @erezzarum ! Few minor comments.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/deepep.Dockerfile
erezzarum: I don't think it's a good idea to have a static check like this, for `ncclGinPlugin_v14`
1. This can change between EFA Installer versions, we mention that EFA Installer 1.50+ should be used.
2. The aws-ofi-nccl library path are subject to change, we should never hard code them.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
erezzarum: Please make sure to install the NVIDIA device plugin with `gdrcopyEnabled=true`
https://docs.aws.amazon.com/eks/latest/userguide/device-management-nvidia-dra-device-plugin.html#eks-nvidia-device-plugin
I will update the README to explain this.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
erezzarum: If you want to solve this, you need to use gang scheduling with a scheduler like volcano, Kueue, etc.
This example is used across all the repo, if we want to change this, we should be doing this in a different PR that solves all the other use cases.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
erezzarum: We don't need to request resources besides GPU and EFA, there is no reason to.
using hugePages as resources blocks cluster autoscalers like Karpenter from provisioning nodes just in time, there is no impact in using hugePages for this kind of workload as well.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
KeitaW: Fair point. The bounded-wait fix only papers over what gang scheduling actually solves, and since this template is shared across the repo, changing it only here would fork the pattern. Withdrawing this as a blocker. I still think an unbounded wait is worth fixing eventually, since a dead worker currently ties up a full node until someone notices. I will open a follow-up issue so it can be discussed once for all the examples rather than per PR.

micro-benchmarks/expert-parallelism/deepep-v2-benchmark/kubernetes/test-internode.yaml
KeitaW: Withdrawn. The Karpenter argument settles it, and today's run supports you empirically: this manifest, without the hugepages or memory requests, sustained ~115 GB/s (SO) dispatch on 2x p6-b300. The v1 parity I asked for was cosmetic.

KeitaW (approved):
Re-verified end to end today on ml-clusters-shared-us-east-1, 2x p6-b300. One infra-side note for anyone reproducing this: on a GPU-operator-managed cluster the plugin runs in CDI mode, where the per-pod `NVIDIA_GDRCOPY=enabled` env is ignored (NVIDIA/k8s-device-plugin#1692), so the README's requirement translates to setting `GDRCOPY_ENABLED=true` on the device plugin itself. With that in place, `test-internode.yaml` ran unmodified and unprivileged: `/dev/gdrdrv` injected via CDI, `Loaded gin plugin Libfabric_GDAKI (v14)` on 16/16 ranks, dispatch ~115 GB/s (SO), clean finalize. The README note is accurate and sufficient. Approving. Thanks for turning around all 16 threads so quickly.

PR #1264, opened 2026-09-14
fix(easy-ssh): verify keys via remote grep count to avoid false failures

aravneelaws:
## Purpose

Fixes a false-negative in `architectures/sagemaker-hyperpod-slurm/easy-ssh.sh` where the script reports

```
Error: Failed to add SSH public key to the cluster. The key was not found after writing.
```

even though the key was written correctly and `ssh <cluster>` works.

## Changes

- Replace the two `cat authorized_keys | grep -Fq "$PUBLIC_KEY"` checks (the initial "already present?" check and the post-write verification) with a **remote match that returns a single integer**: `grep -cF <blob> authorized_keys` run on the node, parsed as a count.
- Match on the **base64 key body** (field 2 of the public key, no spaces) so the SSM command contains no spaces, quotes, `&&` or `||` — all of which either break the AWS CLI `--parameters` shorthand parser or are not honored by the `AmazonEKS-ExecuteNonInteractiveCommand` document.
- Add a short retry loop to each check, and treat a non-numeric response as **inconclusive** (soft note) rather than a hard failure; a definite `0` is still reported as a real error.

### Root cause

The old logic `cat`-ed the whole `authorized_keys` back over the SSM `AmazonEKS-ExecuteNonInteractiveCommand` document and grepped the captured output locally. That document's stdout over `start-session` is **not reliably delivered in full** — for larger payloads it intermittently returns banner-only output or is **truncated mid-line**. When the read was cut off before the matched key, the local `grep -Fq` missed a key that had actually been written, producing the false error.

This is payload-size + plugin-build + timing dependent, which is why it reproduces on a jump host (multiple keys present, ~575-char RSA key) but not always from a developer laptop against a fresh cluster. A tiny integer response from `grep -cF` cannot be truncated, so the check is robust.

## Test Plan

**Environment:**

- AWS Service: SageMaker HyperPod (Slurm)
- Instance type: controller `ml.m5.4xlarge`; jump host `m5.large` (Ubuntu 24.04)
- Number of nodes: 1 controller

**Test commands:**

```bash
# From an Ubuntu 24.04 jump host, against a HyperPod Slurm cluster whose
# authorized_keys already contains another key, using an RSA key:
./easy-ssh.sh -c controller-machine <cluster-name>
# answer yes to add the key, then:
ssh <cluster-name>
```

## Test Results

Before the fix (same jump host, RSA key), the script wrote the key successfully but reported:

```
Error: Failed to add SSH public key to the cluster. The key was not found after writing.
```

After the fix, on the same jump host:

```
Adding ... ssh-rsa AAAAB3Nza... participant@CodeEditor
✅ Your SSH public key /home/participant/.ssh/id_rsa.pub has been added to user ubuntu on the cluster.

Now you can run:

$ ssh trn-cluster
...
You're on the controller
```

`ssh <cluster>` connects successfully. Both the key-absent (prompt -> add -> confirm) and key-present ("Detected ... Skipping") paths were also validated end to end against a live HyperPod Slurm cluster.

## Directory Structure

Not applicable — this modifies an existing helper script in `architectures/sagemaker-hyperpod-slurm/`, not an example.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/awslabs/awsome-distributed-ai/blob/main/CONTRIBUTING.md).
- [x] Test Results above contain e2e evidence from a real AWS run (or this PR is marked `docs-only`).


allela-roy (approved):
LGTM, verified fix.

PR #1266, opened 2026-09-16
Fix ramp/collector/transport evidence defects and record the measured paired result (AIM345)

whn09:
## What this changes

Defects found by running this directory's labs end to end on the hardware the session ships on — the paired main-line flow on 2026-09-16, the optional separate-node V4 profile on 2026-09-17, and a follow-up run the same day that localizes the one anomaly the second run left open. Three are code fixes in this directory; the last two record measured results that contradict documented expectations.

| Change | Why it matters at the event |
|---|---|
| `benchmark.py` requires the router to be *routable*, not merely alive, before each rate | Two faults wrote all-failures files to `results/` and then aborted the sweep, both indistinguishable from an engine fault: a dropped `port-forward`, and a router answering `/health` 200 while `/generate` returns 503 for 67–110 s after an engine restart. |
| `5.collect.py` scopes the peak-throughput field to SLO-qualifying rates | The two summary fields in `qualified-rates.json` used different scopes, so the two architectures were compared on different bases. |
| `7.verify-transport.py` reports `VALIDATED-SELECTORS-ONLY` for packed placement | `status: VALIDATED` was printed together with `rdma_counter_delta_bytes: 0`. Packed placement is what the prepared paired mode always uses. |
| `README.md` records the measured paired result, the Python 3.12 requirement, port-forward restart loops, and expected startup warnings | Participants would otherwise spend session time investigating a correct result and three harmless log lines. |
| `README.md` corrects the optional V4 digest guidance and records that profile's measured result | The recorded engine digests are the authors' own build and cannot be pulled by anyone else, and the optional shape-A expectation needs a settled router and a note on where TTFT is measured from. |

The largest item is documentation. **The Round 2 expectation is contradicted by measurement**: unified qualified at 4.0 tasks/s against 0.5 for disaggregated on shape B, and disaggregated also broke first on shape A. That is not a fault — `completed_calls` equalled `sent_calls` at all 11 rate points, no pod restarted, and both pod UIDs were unchanged across the whole sweep. The mechanism is visible in the data: disaggregated holds p90 TPOT near 13 ms while unified climbs to 133.8 ms and breaks the objective, and loses TTFT because packed 4+4 prefills on 4 of the node's 8 GPUs while both shipped shapes are prefill-bound. Neither shipped shape is decode-bound, which is the case where disaggregation would be expected to win. The README now states this direction before the sweep commands so a wide gap is read as the result.

## Why this is not duplicating an existing PR

`gh pr list --state open --search "disaggregated"`, `"prefill"`, and a scan of all 100 open PRs return nothing touching `examples/use-cases/disaggregated-inference-prefill-decode/`; the `deepep-efa` matches are MoE all-to-all work in other directories. `gh issue list --search "aim345 OR disaggregated OR prefill"` returns no open issues. The base branch `riv2026/aim345-content` is at `f1ad17b`, and `git rev-list --count f1ad17b..origin/riv2026/aim345-content` is 0, so nothing has landed upstream since the commit the workshop pins.

### First run: the paired main-line flow, 2026-09-16

End-to-end run of the paired flow at README lines 136–162, on the target hardware.

**Environment.** Two `g7.48xlarge` in `eu-south-2` (8× RTX PRO 4500 Blackwell, 2× EFA, NVMe RAID0), EKS 1.34, packed placement, 8 GPUs per stack, DeepSeek-V2-Lite-Chat at revision `85864749cd611b4353ce1decdb286193298f64c7`, p90 TTFT 2000 ms / p90 TPOT 100 ms at 0.90 attainment, 30-second windows, 2026-09-16 05:50–08:20 UTC. Sequence run: `paired.py render/deploy` → `paired.py verify` → `7.verify-transport.py` → both traffic shapes → 4 ramps over 11 rate points → 3× `5.collect.py` → `paired.py verify` → `8.warm-connections.py` → `paired.py cleanup`.

**Measured shape B, p90 by offered rate.** Every row has `completed_calls == sent_calls`.

| Rate | unified TTFT / TPOT / attain | disaggregated TTFT / TPOT / attain |
|---|---|---|
| 0.25 | 1051.8 ms / 6.7 ms / 1.00 | 1808.5 ms / 7.6 ms / 1.00 |
| 0.5 | 1059.5 ms / 9.2 ms / 1.00 | 1660.3 ms / 8.9 ms / 1.00 |
| 1 | 1051.8 ms / 11.9 ms / 1.00 | 2175.3 ms / 11.1 ms / **0.77** |
| 2 | 1058.3 ms / 22.8 ms / 1.00 | 16 295.0 ms / 14.7 ms / 0.08 |
| 4 | 1670.3 ms / 93.7 ms / 0.95 | 56 345.5 ms / 13.8 ms / 0.02 |
| 8 | 30 299.4 ms / **133.8 ms** / 0.00 | 172 743.0 ms / **13.5 ms** / 0.00 |

**Measured shape A** (five rates, added by this PR; the previous single 0.1 rate left both arms at attainment 1.00): unified holds the objective through 4 tasks/s at 0.6962 useful calls/s/GPU; disaggregated misses at 1 task/s with p90 TTFT 8320.0 ms and attainment 0.28.

**`5.collect.py` fix, replayed over the run's own measured files** — same inputs, only the row whose peak fell outside the objective changes:

```text
before (f1ad17b)                       after (this branch)
unified       shapeB  max_perGPU=0.3453   unified       shapeB  max_perGPU=0.3453
disaggregated shapeB  max_perGPU=0.0774   disaggregated shapeB  max_perGPU=0.0418
unified       shapeA  max_perGPU=0.0856   unified       shapeA  max_perGPU=0.0856
disaggregated shapeA  max_perGPU=0.0787   disaggregated shapeA  max_perGPU=0.0787
```

`0.0774` came from the 1 tasks/s row at 0.77 attainment; `0.0418` is the highest qualifying rate, 0.5 tasks/s. The unified arm is unchanged because its peak already fell on a qualifying row — which is exactly the asymmetry the fix removes.

**`benchmark.py` fix, against a closed port** — named cause, no file written:

```text
$ python3 4.ramp.py --traffic traffic/a.json --url http://127.0.0.1:59999 ... --rates 0.25 0.5
Router unreachable at http://127.0.0.1:59999/health (ConnectError: All connection attempts failed).
The port-forward has most likely dropped: restart it, confirm the pods are still Running, and rerun
this rate. No measurement file was written.
$ ls -A results/ramp-guard-test    # empty
```

Before this change the same condition produced `sent_calls: 3, completed_calls: 0, p90_ttft_ms_success_only: null, meets_joint_slo: false` and stopped the sweep. That is what was actually hit on the cluster, after shape B at 8 tasks/s, while `kubectl get pods` showed all four pods `1/1 Running` at 0 restarts for 39m.

**`7.verify-transport.py`**, observed on the cluster before the change:

```json
{"status": "VALIDATED", "placement": "packed",
 "cross_node_efa_status": "UNVALIDATED: co-located prefill/decode",
 "rdma_counter_delta_bytes": 0}
```

`fi_info -p efa` in the engine container was healthy (`efa` and `efa-direct` fabrics, both devices), so the flat counter reflects libfabric selecting a same-node provider, not a broken transport. This PR renames that outcome and keeps the number in the saved report.

**Unit tests and lint**, on Python 3.14 with `requirements.txt` installed:

```text
$ python3 -m unittest -v test_paired test_lab
Ran 10 tests in 7.971s
OK
$ npx markdownlint-cli2 "examples/use-cases/disaggregated-inference-prefill-decode/README.md"
Summary: 0 issues in 0 files
```

### Second run: the optional separate-node V4 profile, 2026-09-17

`docs-only` for this commit — the diff is README text. The measurements behind it are below.

**Environment.** Two `g7.48xlarge` in `eu-south-2`, EKS 1.36, `separate-nodes` placement, 8 GPUs and 2 EFA devices per engine, prefill TP size eight on one node and decode TP size eight on the other, DeepSeek-V4-Flash at revision `60d8d70770c6776ff598c94bb586a859a38244f1`, same SLO and 30-second windows as above, 2026-09-17 03:22–05:30 UTC. The engine was rebuilt from `Dockerfile.v4` at `f1ad17b` in a different account and region from the recorded run, then pushed to a private registry; `verify_image.py` reported `v4-optional` both at build time and from an independent `docker run`.

**Why the digest guidance changed.** The rebuild resolved to engine manifest `sha256:13916e808ae70b0ff658cddbb4faa7164d4b4e92e79f26eb03893a2c54a04738`, not the recorded `sha256:7da39d58804be…`. Same Dockerfile, same commit, same pinned base image. That is expected for container builds and is why the README now says to deploy the digest you pushed.

**Cross-node transport.** `7.verify-transport.py` reproduced the recorded counters exactly, in a different account, region, and image build:

```json
{"status": "VALIDATED", "engine_profile": "v4-optional", "placement": "separate-nodes",
 "cross_node_efa_status": "VALIDATED positive path",
 "rdma_counter_delta_bytes": 97235456}
```

Per device: prefill `rdmap176s0` `48611328`, prefill `rdmap83s0` `48624128`, both decode devices `0`. This is the positive counterpart to the packed row that this PR renames to `VALIDATED-SELECTORS-ONLY`, and it is the only separate-node observation the lab can produce.

**Long-context sweep — the ceiling is `--max-running-requests 4`, which `deployment.py` appends for this profile.** Every row has `completed_calls == sent_calls` and `failed_or_skipped_calls: 0`:

| Offered | Sent calls/s | p90 TTFT | p90 TPOT | Joint SLO |
|---|---|---|---|---|
| 0.25 | 0.10 | 3359.7 ms | 18.0 ms | false |
| 0.5 | 0.28 | 5751.0 ms | 19.8 ms | false |
| 1 | 0.55 | 13 778.0 ms | 19.4 ms | false |
| 2 | 0.66 | 39 528.6 ms | 16.9 ms | false |

Offered rate rises 8× while the admitted rate saturates and TTFT grows 12×, with TPOT flat. That is admission queueing, not a decode limit. "No tested long-context rate qualified" reproduces; the reason did not appear anywhere in the README.

**Shape A at 0.1 tasks/s did not reproduce here.** Six consecutive phases on one engine, same traffic file, same seed, `shape_sha256` identical, minutes apart. The third run below identifies the mechanism:

| Run | Phase | p90 TTFT | p90 TPOT | Attainment | Joint SLO |
|---|---|---|---|---|---|
| `v4-a` | warmup | 1024.4 ms | 21.4 ms | 1.00 | true |
| `v4-a` | measured | **4072.2 ms** | 21.7 ms | **0.57** | **false** |
| `v4-a2` | warmup | 956.8 ms | 21.6 ms | 1.00 | true |
| `v4-a2` | measured | 1035.1 ms | 21.5 ms | 1.00 | true |
| `v4-a3` | warmup | **13 896.3 ms** | 21.7 ms | **0.67** | **false** |
| `v4-a3` | measured | 919.8 ms | 21.6 ms | 1.00 | true |
| recorded | measured | 665.3 ms | 20.5 ms | 1.00 | true |

Failures fell in both the warmup and the measured phase, so the README's existing advice — use the measured file after `4.ramp.py`'s warmup — does not address it. `completed_calls == sent_calls`, `failed_or_skipped_calls: 0`, `client_valid: true`, and p90 TPOT never left 21.4–21.7 ms, so this is TTFT only. Within the failing measured phase the per-call values are bimodal in contiguous blocks, not a decaying curve:

```text
633, 629, 633, 896, 874, 843, 4072, 3916, 4040, 819, 811, 806,
3964, 3954, 4081, 4024, 4019, 4143, 965, 995, 793
```

**Shape B at 0.1 tasks/s.** p90 TTFT `3372.1` ms against the recorded `3095.9`, p90 TPOT `18.1` against `17.4`, and `meets_joint_slo: false` both times. `slo_attainment_fraction` was `0.0` against a recorded `0.3333`, which is one call out of three — that field has a resolution of 1/3 at this rate and is not a meaningful comparison.

**Lint.** `markdownlint-cli2` on the changed README → `Summary: 0 issues in 0 files`.

### Third run: verifying the router guard, 2026-09-17

Same cluster and deployment, 06:00–08:32 UTC, 28 further shape-A phases at 0.1 tasks/s.

**The defect.** After the engine pods restart with the router left running, the router's own `/health` keeps returning 200 while `POST /generate` returns 503, for 90–110 s *after* `kubectl rollout status` reported both engine deployments Ready:

```text
smg::routers::http::pd_router: src/routers/http/pd_router.rs:170: Failed to select PD pair
  error=No available decode workers (all circuits open or unhealthy)
status_code=503 latency=209
```

Router-side latency of 127–233 ms shows this is the router's own circuit-breaker decision, not a 503 passed through from an engine. Two of three invocations were destroyed by it, since `--disable-retries` makes one 503 fail an entire agent task. The guard already on this branch could not detect it: it probed only `/health`.

**The fix, measured.** Removing the diagnostic env var restores the shipped configuration and is itself an engine restart, so it recreates the window:

```text
                   phase       sent  done  failed   p90 TTFT   joint
v4-loc1 (old code) measured       3     0      21          -   false   <- destroyed
v4-loc2 (old code) measured       3     0      21          -   false   <- destroyed
v4-loc3 (old code) warmup        13    12       9   12250.42   false
v4-guard1 (new)    warmup        21    21       0     859.87   true
v4-guard1 (new)    measured      21    21       0     976.12   true
```

On that restart the breaker had already closed 16 s after Ready, so a new unit test covers the wait path: it stands up a server answering `/health` 200 and `/generate` 503, and asserts the guard both waits it out and exits with a named cause if it never clears. `Ran 11 tests ... OK`.

The wait path then fired on hardware during a later main-line run, on the **paired flow's plain router** rather than `pd_router` — so this is not specific to the optional profile:

```text
Router answers /health but not /generate (HTTP 503: {"error":{"type":"Service Unavailable",
"code":"no_available_workers","message":"No available workers (all circuits open or unhealthy)"}}).
Waiting up to 180 s for it to become routable.
Router became routable after 67 s.
```

All four rates after it recorded `completed_calls == sent_calls`.

**Why this is the router and not the engines.** During a 13.9 s recorded TTFT both engine logs are silent for the whole window and the three concurrent tasks stall together to within 11 ms. Neither network path stalls: 900 samples at 1 Hz through the same `kubectl port-forward` gave p99 `523` ms with none over 1 s — collected while the client was recording a `12 250` ms p90 TTFT — and in-cluster prefill→decode gave p50 `2.3` ms. 22 of 22 warm phases passed within `848.7`–`959.4` ms. Why the breaker opens at all is still unexplained.

## Deliberately not included

- **Setting `SGLANG_DIAG_BYPASS_HEALTH_GENERATE` in `deployment.py` — tested and rejected.** `SGLANG_ENABLE_HEALTH_ENDPOINT_GENERATION` defaults to `EnvBool(True)` in the pinned engine, and `/health` is aliased onto the same handler as `/health_generate`, so on an idle engine every `GET /health 200 OK` is preceded by a real scheduler prefill batch at the ten-second probe period. That was the first hypothesis for the shape-A instability and worth ruling out, since the lab's output is p90 TTFT under a four-slot admission ceiling. A controlled run rules it out: 2 of 12 phases missed with the flag set against 2 of 14 unset (one-sided Fisher `p = 0.7609`), and raising the probe rate tenfold to 1 Hz produced 0 misses in 6 phases with the tightest spread observed. Setting it would therefore change a shared file that `paired.py` also consumes for no measured benefit. The README records the behaviour as a diagnostic aid for reading engine logs.
- **A third, decode-bound traffic shape** (short input, long output, high concurrency). This is the largest content gap the run exposed: without it the two shipped shapes only ever demonstrate the cost side of disaggregation. It needs a GPU run to qualify, so it is not in this PR. The README names the gap; the generators already accept `first_input_tokens` and `output_tokens`.
- **Blackwell open-kernel-module, chart pinning, instance-type and CodeBuild Python fixes.** Those live in the Workshop Studio repository's CloudFormation template and `contentspec.yaml`, not here, and are filed separately.

## Still open

These are content-owner decisions rather than review of the diff, and none of them blocks this PR: whether the session narrative moves to "when not to disaggregate" or waits for a decode-bound shape; whether the `VALIDATED-SELECTORS-ONLY` status value is the wording you want in `VALIDATION.md` as well; and, for the optional profile, whether the recorded `665.33 ms` for shape A was measured from inside the cluster's region. That last one is now a specific question rather than an open-ended one: ~485 ms of the `~900 ms` this run recorded is the cross-region `kubectl port-forward` round trip, which would put an in-region measurement at roughly the recorded figure. If so, the README should say to compare rates from a single client rather than against the recorded absolute value — which is what this PR now says — and the recorded number should be annotated with where it was taken from.

Marking `Round 2` as `CONTRADICTED` is a factual record of this run, not a proposal to retire the exercise. The same applies to the new optional-profile row.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


KeitaW:
Much appreciated @whn09 ! I'll take a look at it tomorrow

PR #1267, opened 2026-09-17
feat(aws-pcs): targeted ODCR support for compute node groups

### LINKED ISSUE #1263: feat(aws-pcs): support targeted On-Demand Capacity Reservations (ODCR) for compute node groups
### Problem

`CapacityReservationId` on the GPU node-group templates (`add-cng-p5.yaml` / `add-cng-p6-b200.yaml` / `add-cng-p6-b300.yaml`) currently supports **Capacity Blocks for ML only**: setting the parameter forces `InstanceMarketOptions: MarketType=capacity-block` and drops the placement group from the launch template.

This means a **targeted ODCR cannot be consumed** through these templates:

- A targeted ODCR requires `CapacityReservationTarget` **without** `MarketType=capacity-block` — the current CB-shaped launch template makes EC2 reject the launch.
- Only *open* ODCRs work today, implicitly, via the empty-`CapacityReservationId` On-Demand path.
- Users holding reserved GPU capacity as a targeted ODCR (a common shape for long-lived P5/P6 reservations, as opposed to short-lived CBs) have no supported way to point the node group at it; the docs currently carry a "do not put an ODCR ID here" caveat instead.

The generic CPU CNG (`add-cng.yaml`) has no capacity-reservation support at all — it always launches plain On-Demand.

### Proposal

Add a `CapacityReservationType` parameter and branch the launch template accordingly.

**GPU CNG templates** (`add-cng-p5` / `add-cng-p6-b200` / `add-cng-p6-b300`) — enum `none` | `capacity-block` | `targeted-odcr`, default `none`:

| Type | InstanceMarketOptions | CapacityReservationTarget | Placement group |
|---|---|---|---|
| `none` | — (On-Demand) | — | auto-created PG (as today) |
| `capacity-block` | `MarketType=capacity-block` | the CB ID | dropped (CB carries its own placement) — as today |
| `targeted-odcr` | — (On-Demand billing against the reservation) | the ODCR ID | **kept** (auto-created PG) |

**Generic CPU CNG (`add-cng.yaml`) is in scope too**, with a reduced enum (`none` | `targeted-odcr` — Capacity Blocks don't exist for these instance families). It currently has no capacity-reservation support at all, and its existing placement-group logic (auto-created or `PlacementGroupName` override) composes naturally with a targeted ODCR.

Backward compatibility: `none` and `capacity-block` reproduce today's behaviors exactly; existing stacks that set `CapacityReservationId` keep working if `capacity-block` is the behavior-preserving default whenever the ID is non-empty (exact defaulting/validation to be settled in the PR — e.g. a CFN rule rejecting `CapacityReservationId=""` combined with `targeted-odcr`/`capacity-block`).

Docs: replace the "do not put an ODCR ID here" caveat with the new parameter's semantics; note the AZ-match requirement (reservation AZ must equal `PrimarySubnetAZ`, same constraint as CBs).

### Verification plan (no GPU capacity required)

1. **Static:** create the GPU CNGs with `Min/MaxCount=0` for each of the three types and assert the generated launch template's `CapacityReservationSpecification` / `InstanceMarketOptions` / `Placement` sections per the table above.
2. **Dynamic:** create a targeted ODCR for a cheap instance type (c6i), deploy the **generic CNG** against it, and assert the launched instance shows `InstanceLifecycle` empty (= On-Demand) and the reservation's available count decrements. Since the generic CNG is in scope, this exercises the shipped code path directly — not a proxy — and the GPU templates share the identical branch logic (covered by the static checks).

### References

- Tracked in `architectures/aws-pcs/docs/ROADMAP.md` ("Targeted ODCR support for GPU node groups")
- [EC2 User Guide — target a specific Capacity Reservation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-using.html)
- [Capacity Blocks for ML](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-capacity-blocks.html)

DaisukeMiyamoto:
Closes #1263.

## What

- **GPU CNG templates** (`add-cng-p5` / `add-cng-p6-b200` / `add-cng-p6-b300`): new `CapacityReservationType` (`capacity-block` default | `targeted-odcr`). `targeted-odcr` sets `CapacityReservationTarget` **without** `MarketType=capacity-block` and keeps the cluster placement group (On-Demand billing against the reservation). New `PlacementGroupName` override (deploy-all: `PseriesPlacementGroupName`) so a targeted ODCR held in a customer-owned cluster placement group is consumable.
- **Generic CPU CNG** (`add-cng.yaml`): new `CapacityReservationId` targets an ODCR directly (no market-type change; no type enum — Capacity Blocks don't exist for these families). deploy-all: `OnDemandCapacityReservationId`. A user-supplied `PlacementGroupName` is now honored **with or without EFA** — previously it was silently ignored at `EfaInterfaceCount=0`, which made CPG-held ODCRs unusable on non-EFA types.
- **Docs**: capacity options rewritten (On-Demand / open ODCR / targeted ODCR / Capacity Block), §8.6 CPU-queue ODCR coverage, parameter tables, ROADMAP item closed; replaces the "do not put an ODCR ID here" caveat.

## Backward compatibility (zero behavior change for existing stacks)

- `CapacityReservationType` defaults to `capacity-block`, so every existing stack that sets `CapacityReservationId` keeps the exact CB behavior on update; an empty ID remains plain On-Demand / open-ODCR (the type is ignored). There is deliberately no `none` enum value — ID emptiness governs, exactly as today.
- All new parameters default to empty; default-path launch templates are equivalent (verified below).

## Why not duplicating

No open PR addresses #1263 (searched `1263 in:body` and `ODCR`).

## Test Results

Verified on real hardware with fresh `pcs-ml-cluster-deploy-all.yaml` deployments from a private template bucket (one CPU cluster, one GPU cluster in different regions):

- **Targeted ODCR consumption (dynamic, generic CNG)**: a CPU type and 2× EFA-capable GPU instances — launched instances carry the reservation ID, `InstanceLifecycle` is empty (On-Demand billing), and the ODCR's available count decremented to 0.
- **CPG-held targeted ODCR (dynamic)**: an ODCR created inside a customer-owned cluster placement group; the node launched into that CPG and consumed the reservation — exercises the placement-group-without-EFA fix on a non-EFA instance type.
- **GPU rendering (static, `Min/MaxCount=0` stacks)**: `targeted-odcr` → `CapacityReservationTarget` + placement group kept + no `InstanceMarketOptions`; user-PG override honored; default path unchanged (no CR spec, auto-created PG, no `PurchaseOption`). PCS fails fast at CNG create on a reservation/instance-type mismatch with a clear error.
- **Default-path regression (static)**: login / CPU / GPU launch templates with all new parameters at defaults contain no `CapacityReservationSpecification` / `InstanceMarketOptions` and unchanged `Placement`.
- **Capacity Block path**: no real CB was available; the path is unchanged by construction (condition truth-table equivalence — `MarketType`, `PurchaseOption`, reservation target, and placement all resolve identically with the default type). `add-cng-p6-b200`/`-b300` carry capacity/placement logic identical to the live-tested `add-cng-p5` (section diff).
- **EFA sanity on the GPU cluster**: 2-node × 8-GPU NCCL `all_reduce_perf` over EFA (`NET/OFI Selected provider is efa`, 2 NICs, ~39.5 GB/s busbw at 1 GiB).
- `tests/lint-docs.sh` PASS; `cfn-lint` clean (no E-level findings).


architectures/aws-pcs/docs/PARAMETERS.md
```diff
@@ -51,7 +51,8 @@ it directly.
...
+| `OnDemandPlacementGroupName` | *(empty)* | Existing cluster placement group name to launch nodes into. Empty + `OnDemandEfaInterfaceCount > 0` auto-creates a per-CNG cluster placement group; supplying a name reuses an existing one (e.g. shared across multiple CPU CNGs, or a customer-owned CPG holding reserved capacity). A supplied name is honored with or without EFA (so a CPG-held targeted ODCR is consumable on non-EFA types); empty + EFA off = no placement group, as before. The GPU queue's equivalent is `PseriesPlacementGroupName` |
+| `OnDemandCapacityReservationId` | *(empty)* | **Targeted ODCR** ID for the CPU queue: launches bill On-Demand against the reservation, everything else (EFA, placement) is unaffected. Leave empty for plain On-Demand; an "open" ODCR is consumed automatically and must not be listed here. The reservation's instance type / AZ must match `OnDemandInstanceType` / `PrimarySubnetAZ` |
```
KeitaW: ### `MaxCount` ≤ reservation size and "no backfill" are not documented (should-fix, docs)

For direct ID targeting the PCS User Guide is explicit: "*does not support instance backfilling (If the compute node group requests more instances than available in the ODCR, no additional instances will be launched)*" and "*For a static node group, set minimum and maximum instances to the size of the reservation*" (https://docs.aws.amazon.com/pcs/latest/userguide/capacity-reservations-odcr.html); EC2 likewise: "*If capacity isn't available in the specified Capacity Reservation, the instance fails to launch*" (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-launch.html). PCS validates `maxInstanceCount` against the block size for Capacity Blocks at CNG create (https://docs.aws.amazon.com/pcs/latest/userguide/capacity-blocks-configure-cng.html) but documents no such check for ODCRs.

With the default `PseriesMaxCount=4` (or an `OnDemandMaxCount`) above the reserved count, the launches beyond the reservation keep failing and jobs sit pending — there is no On-Demand fallback on a targeted ID, and the templates don't expose the backfill-capable `CapacityReservationResourceGroupArn`. A sentence in the README GPU capacity section, in §8.6, and on both PARAMETERS.md rows would save users that surprise:

```suggestion
| `OnDemandCapacityReservationId` | *(empty)* | **Targeted ODCR** ID for the CPU queue: launches bill On-Demand against the reservation, everything else (EFA, placement) is unaffected. Leave empty for plain On-Demand — an "open" ODCR is consumed automatically. The reservation's instance type / AZ must match `OnDemandInstanceType` / `PrimarySubnetAZ`. Keep `OnDemandMaxCount` ≤ the reservation's instance count — a direct ODCR target does not backfill with On-Demand, so launches beyond the reservation fail |
```

Leaving a "Capacity Reservation group ARN (backfill)" item in the ROADMAP would make it clear the limit is deliberate.

KeitaW (commented):
## Review Batch 1/2 — One thing to add before merge

The CloudFormation wiring checks out: the two legacy paths (`CapacityReservationId` empty, and set with the default `capacity-block` type) resolve to exactly the same `InstanceMarketOptions` / `CapacityReservationSpecification` / `Placement` / `PurchaseOption` as on `main` (I traced all 8 combinations of ID × type × placement-group name), `PurchaseOption` is still keyed off `UseCapacityBlock`, and `cfn-lint`, `validate-template` and `tests/lint-docs.sh` are clean. One documentation gap is worth closing because it is the first thing a user of this feature will hit.

KeitaW (commented):
## Review Batch 2/2 — Things That Look Great & Sources

### Things That Look Great

- **The mechanism this PR relies on holds up on real hardware.** On a running PCS cluster in eu-south-2 with a `purchaseOption=ONDEMAND` GPU node group whose launch template carries only a `CapacityReservationTarget`, PCS copies that block unchanged into its derived launch template and the instances land in the targeted ODCR via EC2 Fleet — `CapacityReservationId` matches, `InstanceLifecycle` is empty, IMDS reports `instance-life-cycle=on-demand` (verified live, 2026-09-17). That is exactly the launch-template form the `targeted-odcr` path renders, and it matches the PCS-documented example (https://docs.aws.amazon.com/pcs/latest/userguide/capacity-reservations-odcr.html).
- **Backward compatibility of the legacy paths is preserved by construction**: across the 8-way truth table, `ID=""` and `ID set + capacity-block` resolve identically to `main`, and `PurchaseOption` stays keyed off `UseCapacityBlock` — correct, since PCS's `purchaseOption` is `ONDEMAND | SPOT | CAPACITY_BLOCK | INTERRUPTIBLE_CAPACITY_RESERVATION` and ODCRs run under `ONDEMAND` (https://docs.aws.amazon.com/pcs/latest/APIReference/API_CreateComputeNodeGroup.html).
- **The three GPU templates are in lock-step**: extracting the capacity/placement Parameters, Conditions and Resources from p5 / p6-b200 / p6-b300 and diffing gives a byte-identical result. `tests/lint-docs.sh` passes and covers the four new deploy-all parameters' PARAMETERS.md rows.
- **`AllowedValues` on `CapacityReservationType`** so a typo fails at validation, and the decision *not* to add a `none` enum value (emptiness of the ID governs) is written down both in the PR description and in the Conditions comments.
- **Docs consistency**: the old "do not put an ODCR ID here" caveat is gone from all four templates and all three docs (grep-verified), and the "CPU CNG only" wording on `OnDemandPlacementGroupName` is retired with it.
- The Conditions comments explaining why only the Capacity Block path drops the placement group make the intent easy to follow.

### Sources

- EC2 — Launch instances into an existing Capacity Reservation: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/capacity-reservations-launch.html
- PCS — Using ODCRs with AWS PCS: https://docs.aws.amazon.com/pcs/latest/userguide/capacity-reservations-odcr.html
- PCS — Configure a compute node group to use a Capacity Block: https://docs.aws.amazon.com/pcs/latest/userguide/capacity-blocks-configure-cng.html
- PCS — CreateComputeNodeGroup API (`purchaseOption`): https://docs.aws.amazon.com/pcs/latest/APIReference/API_CreateComputeNodeGroup.html
- verified live, 2026-09-17: `cfn-lint 1.56.3` on the 5 changed templates (0 errors; 10× W3005 pre-existing on `main`) / `aws cloudformation validate-template --region eu-south-2` ×5 OK / `bash architectures/aws-pcs/tests/lint-docs.sh` PASS / a running PCS GPU node group in eu-south-2 consuming a targeted ODCR via `purchaseOption=ONDEMAND` + LT `CapacityReservationTarget` (`describe-instances`, `describe-capacity-reservations`, `describe-launch-template-versions`, `pcs get-compute-node-group`, IMDS).

KeitaW (approved):
Thanks a lot! Overall LGTM

DaisukeMiyamoto:
Thanks for the thorough two-batch review — especially for tracing all 8 ID × type × placement-group combinations against `main` and for independently verifying the `CapacityReservationTarget`-only launch-template form against a live PCS node group.

The one should-fix is addressed in 633d6c0a:

- **No-backfill constraint documented**: both PARAMETERS.md rows now carry the "keep `MaxCount` ≤ the reservation's instance count — a direct ODCR target does not backfill with On-Demand" note (took your suggestion verbatim for the `OnDemandCapacityReservationId` row), and the same guidance is in the README GPU capacity bullet and §8.6 for the CPU queue.
- **ROADMAP**: added a 🟢 item for the backfill-capable `CapacityReservationTarget.CapacityReservationResourceGroupArn` form, recording that the ID-only, no-backfill behavior is a deliberate limit.

`tests/lint-docs.sh` passes on the updated docs.

